Issue #398, opened 2022-02-01
Create source archive build artifacts


dagood:
To contribute to the supply chain for build-from-source scenarios (Mariner in particular), we need to provide source archives. (`src.tar.gz`)

I think during the build, we should have a job clone source (including the submodule), apply patches, then publish two things to the pipeline:

* The repository as a whole as one tarball.
  * This lets people use our infra to build/test/pack, if that's useful for them.
* Just the patched submodule code as another tarball.
  * This should be interchangeable with the tarballs available from https://go.dev/dl/. Useful for anyone who is already building Go.

These artifacts should have checksum files and signatures generated, so we end up with something like this in our blob storage:
```
go.{buildid}.src.infra.tar.gz
go.{buildid}.src.infra.tar.gz.sha256
go.{buildid}.src.infra.tar.gz.asc
go.{buildid}.src.tar.gz
go.{buildid}.src.tar.gz.sha256
go.{buildid}.src.tar.gz.asc
```

> * Right now we use `.sig`, but we should switch to `.asc`: https://github.com/microsoft/go/issues/181. Upstream does have signature files for their source tarballs, they just aren't advertised on the readable download page.

dagood:
Completed the `go.{buildid}.src.tar.gz` archives. We don't have a need for `go.{buildid}.src.infra.tar.gz` currently.

PR #408, opened 2022-02-07
Create source archive during CI

dagood:
* For #398 
* Updates go-infra version to use https://github.com/microsoft/go-infra/pull/24

Create archive of the patched source files during the internal rolling (official) build. Creates a checksum file and a signature for the archive itself. The files are then uploaded to blob storage alongside the binary build.

This PR splits submodule update+patch into two steps. This lets me conditionally add a "create tarball" step in between.

Test build: https://dev.azure.com/dnceng/internal/_build/results?buildId=1601272&view=results
(This PR shouldn't change anything for PR validation builds or outerloop builds.)

### DIFF
--- a/eng/_core/archive/archive.go
+++ b/eng/_core/archive/archive.go
@@ -35,7 +35,9 @@ func CreateFromSource(source string, output string) error {
 
 	// Ensure the target directory exists.
 	archiveDir := filepath.Dir(output)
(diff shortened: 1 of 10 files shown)


dagood:
@jaredpar @qmuntal @chsienki PTAL

After this goes in, I need to backport and merge it through all the release branches.

jaredpar (approved):

qmuntal (approved):

Issue #173, opened 2021-08-13
Add linux-arm64 builders


dagood:
This starts off simple--`GOARCH=arm64`--but there are a few things to solve.
* Always build the `_core` tool for the host OS, not the target. We need to be able to execute the tool. This is not hard (just clear out the value to make Go detect it), prototype here: https://github.com/microsoft/go/compare/microsoft/main...dagood:dev/dagood/cross-support
* Set up cross-compile C++ toolchain to build Go with CGo enabled. Otherwise this comes up when trying to install/build the race runtime:
  ```
  ...
  Building packages and commands for host, linux/amd64.
  Building packages and commands for target, linux/arm64.
  ---
  Installed Go for linux/arm64 in /home/dagood/git/go
  Installed commands in /home/dagood/git/go/bin
  ---- Building race runtime...
  ---- Running command: [../bin/go install -race -a std]
  go install: -race requires cgo
  panic: exit status 2
  
  goroutine 1 [running]:
  main.main()
          /home/dagood/git/go/eng/_core/cmd/build/build.go:60 +0x2c9
  ```
* How will we run tests? (Cross-compile then test in a real machine? Where are these machines available--Helix?) (Probably want to split this to another issue.)

dagood:
I ran some hacky builds with our pipeline on borrowed, real arm64 hardware and got some configs working on 1.17 (picked for branch stability) on an arm64 Mariner 1.0 image I built:
* https://dev.azure.com/dnceng/internal/_build/results?buildId=1629937&view=results - linux-arm64-buildandpack
* https://dev.azure.com/dnceng/internal/_build/results?buildId=1631486&view=results - linux-arm64-test
  * I had to pick a fix from upstream into 1.17 and apply a theoretical flakiness fix to get the tests to go all the way.

dagood:
.NET Docker made these changes to run on a new arm64 pool that we will need to do in go-images (but ideally take it semi-automatically by updating the entire infra to latest): https://github.com/dotnet/docker-tools/pull/983

dagood:
Completed by https://github.com/microsoft/go-images/pull/95

PR #474, opened 2022-03-14
Add arm64 internal builds and build script support

dagood:
Test build: https://dev.azure.com/dnceng/internal/_build/results?buildId=1662815

```
    {
      "env": {
        "GOARCH": "arm64",
        "GOOS": "linux"
      },
      "sha256": "53716a2690c476a20f0d982aec759af3860b29711bf32e93830d071cfc73cb23",
      "url": "https://dotnetbuildoutput.blob.core.windows.net/golang/microsoft/dev/dagood/arm-main/20220314.5/go.20220314.5.linux-arm64.tar.gz"
    },
```

* For https://github.com/microsoft/go/issues/173
* Uses new Docker image from https://github.com/microsoft/go-infra-images/pull/5
* Next steps:
  * Backport to release branches.
  * Make any necessary changes to go-images auto-update infra and add arm64 there.

### DIFF
--- a/eng/utilities.ps1
+++ b/eng/utilities.ps1
@@ -19,19 +19,60 @@ if ($host.Version.Major -lt 6) {
   throw "Missing prerequisites; see logs above for details."
 }
 
+# Machine architecture query based on: https://github.com/dotnet/install-scripts/blob/df8f863720a462448ad244f03ffeb619f0631bad/src/dotnet-install.ps1#L191-L219
+function Get-Machine-Architecture() {
+  # On PS x86, PROCESSOR_ARCHITECTURE reports x86 even on x64 systems.
+  # To get the correct architecture, we need to use PROCESSOR_ARCHITEW6432.
+  # PS x64 doesn't define this, so we fall back to PROCESSOR_ARCHITECTURE.
+  # Possible values: amd64, x64, x86, arm64, arm
+  if( $ENV:PROCESSOR_ARCHITEW6432 -ne $null ) {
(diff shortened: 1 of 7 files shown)


eng/utilities.ps1
```diff
@@ -19,19 +19,60 @@ if ($host.Version.Major -lt 6) {
...
   # pre-installed. This CI script installs a consistent, official version of Go to a directory in
   # $HOME to handle this. This also makes it easier to locally repro issues in CI that involve a
   # specific version of Go. The downloaded copy of Go is called the "stage 0" version.
-  $stage0_go_version = '1.16.5'
+  $stage0_go_version = '1.17.8'
 
   if ($IsWindows) {
-    $stage0_go_sha256 = '0a3fa279ae5b91bc8c88017198c8f1ba5d9925eb6e5d7571316e567c73add39d'
-    $stage0_go_suffix = 'windows-amd64.zip'
+    $proc_arch = Get-Machine-Architecture
```
dagood: Now that I think about it, why not just use `[System.Runtime.InteropServices.RuntimeInformation]::ProcessArchitecture` on Windows, too? Would make this script simpler. I'm going to make this change.

I only grabbed `Get-Machine-Architecture` from the .NET install scripts because I happened to hit the Windows case first, and the .NET install scripts have a Windows-specific implementation in PowerShell. Once I got to the check on Linux, I didn't think back to Windows.

qmuntal (approved):

jaredpar (approved):

Issue #415, opened 2022-02-09
Add auto-retries to Windows test portion


dagood:
Right now we have auto-retries in the Windows build (#299, https://github.com/microsoft/go/issues/241) , but tests also sometimes fail with very similar errors:

> go install cmd/compile: copying C:\Users\VSSADM~1\AppData\Local\Temp\go-build3890923320\b098\exe\a.out.exe: open D:\a\1\s\go\pkg\tool\windows_amd64\compile.exe: Access is denied.

If it's what it seems to be (Windows Defender file locking behaving poorly with file moves) then it may not be practical to fix and retries will provide relief.

dagood:
We have test retries now, but it exacerbates some AzDO UX oddities and bugs which caused problems for me in the previous release. Filing a follow-up issue to try to address that.

* https://github.com/microsoft/go/issues/1060

PR #475, opened 2022-03-14
Add test retries on Windows CI

dagood:
* https://github.com/microsoft/go/issues/415

This PR adds retries to the test portion of the Windows builds in CI, not just build. Test retries are a separate retry context, so the build won't have to also retry--hopefully this makes it so the AzDO job won't completely time out.

Better solutions are possible (skip specific tests in innerloop CI, break apart the test run into "never fail" and "retry" sections, improve Windows' reliability), but those are more difficult, and aren't sure to work. Adding blanket retries takes the pressure off until we can focus on Windows test reliability.

AzDO only allows retrying a failed job once all jobs in the stage have completed, so these retries will improve PR validation wall clock time. This also means a dev won't have to manually monitor their jobs to retry as soon as they can after a Windows flakiness issue shows up.

### DIFF
--- a/eng/_core/cmd/build/build.go
+++ b/eng/_core/cmd/build/build.go
@@ -141,19 +141,10 @@ func build(o *options) error {
 
 		buildCommandLine := append(shellPrefix, "make"+scriptExtension)
 
-		for i := 0; i < maxAttempts; i++ {
-			if maxAttempts > 1 {
-				fmt.Printf("---- Running 'make' attempt %v of %v...\n", i+1, maxAttempts)
-			}
-			err := runCommandLine(buildCommandLine...)
-			if err != nil {
-				if i+1 < maxAttempts {
-					fmt.Printf("---- Build command failed with error: %v\n", err)
-					continue
-				}
-				return err
-			}
-			break
+		if err := retry(maxAttempts, func() error {
+			return runCommandLine(buildCommandLine...)
+		}); err != nil {
+			return err
 		}
 
 		if os.Getenv("CGO_ENABLED") != "0" {
@@ -189,6 +180,11 @@ func build(o *options) error {
 			}
 		}
 
+		maxAttempts, err := getMaxTestRetryAttempts()
+		if err != nil {
+			return err
+		}
+
 		// "-json": Get test results as lines of JSON.
 		if o.JSON {
 			testCommandLine = append(testCommandLine, "-json")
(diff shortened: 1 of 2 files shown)


eng/_core/cmd/build/build.go
```diff
@@ -241,15 +241,43 @@ func runCmd(cmd *exec.Cmd) error {
...
+
 func getMaxMakeRetryAttempts() (int, error) {
-	const retryEnvVarName = "GO_MAKE_MAX_RETRY_ATTEMPTS"
-	a := os.Getenv(retryEnvVarName)
+	return getEnvIntOrDefault("GO_MAKE_MAX_RETRY_ATTEMPTS", 1)
+}
+
+func getMaxTestRetryAttempts() (int, error) {
+	return getEnvIntOrDefault("GO_TEST_MAX_RETRY_ATTEMPTS", 1)
+}
```
qmuntal: IMO `getMaxMakeRetryAttempts` and `getMaxTestRetryAttempts` don't add much value here, I would rather directly call `getEnvIntOrDefault`. This way readers can directly see from where the number of attempts is taken from instead of jumping around.

eng/_core/cmd/build/build.go
```diff
@@ -241,15 +241,43 @@ func runCmd(cmd *exec.Cmd) error {
...
+
+func getEnvIntOrDefault(varName string, defaultValue int) (int, error) {
+	a := os.Getenv(varName)
 	if a == "" {
```
qmuntal: Better use [os.LookupEnv](https://pkg.go.dev/os#LookupEnv). It covers the case where the user has incorrectly set the environment variable to an empty string.

qmuntal (changes requested):

eng/_core/cmd/build/build.go
dagood: Totally agree. I just didn't give these funcs a second thought after duplicating `getMaxMakeRetryAttempts` and extracting the common bits.

eng/_core/cmd/build/build.go
dagood: This depends on what we want the behavior to be, though, right? I was intending to treat unset and empty string both as default, because in my experience, people aren't particularly familiar with `x=` vs. `unset x`.

But... rejecting empty string *could* help if you're trying to set it like `GO_MAKE_MAX_RETRY_ATTEMPTS=$(fun)`, and for some reason `fun` fails rather than sending anything to stdout.

I'll switch over and write an error message for empty string that should point people in the right direction.

eng/_core/cmd/build/build.go
dagood: I'm also moving this into `main` so the command validates the env vars before anything happens, rather than e.g. waiting until after the build's done to tell someone they specified `GO_TEST_MAX_RETRY_ATTEMPTS` wrong.

qmuntal (approved):

dagood:
@jaredpar @chsienki, can one of you take a look?

chsienki (approved):

Issue #481, opened 2022-03-15
Fix TestScript/gopath_std_vendor test in microsoft/dev.boringcrypto.go1.18: our additional vendored library interferes


dagood:
https://github.com/microsoft/go/pull/479 hit an issue with `TestScript/gopath_std_vendor` when trying to update to 1.18.0. I'm filing an issue so it's easier to refer to in the future rather than being buried in a PR discussion thread.

The test failure repros locally with `../bin/go test ./cmd/go -run TestScript/gopath_std_vendor` in `go/src` after building:

https://gist.github.com/dagood/98ecf0e1656061bdb3d5f87ab22cab50

```
[...]
            > go list -test -deps -f '{{.ImportPath}} {{.Dir}}' .
            [stdout]
            vendor/golang.org/x/net/http2/hpack $WORK/gopath/src/vendor/golang.org/x/net/http2/hpack
            issue16333 $WORK/gopath/src/issue16333
            internal/goarch /home/dagood/git/go/go/src/internal/goarch
[...]
            vendor/github.com/microsoft/go-crypto-openssl/openssl/internal/subtle /home/dagood/git/go/go/src/vendor/github.com/microsoft/go-crypto-openssl/openssl/internal/subtle
            runtime/cgo /home/dagood/git/go/go/src/runtime/cgo
            vendor/github.com/microsoft/go-crypto-openssl/openssl /home/dagood/git/go/go/src/vendor/github.com/microsoft/go-crypto-openssl/openssl
            crypto/internal/backend /home/dagood/git/go/go/src/crypto/internal/backend
[...]
            > stdout $GOPATH[/\\]src[/\\]vendor[/\\]golang.org[/\\]x[/\\]net[/\\]http2[/\\]hpack
            > ! stdout $GOROOT[/\\]src[/\\]vendor
            FAIL: testdata/script/gopath_std_vendor.txt:28: unexpected match for `(?m)/home/dagood/git/go/go[/\\]src[/\\]vendor` found in stdout: /home/dagood/git/go/go/src/vendor
```

So, why is this showing up now in 1.18, not 1.17? The test script hasn't changed for three years.

I tried [the `-work` command to keep the test files around](https://github.com/golang/go/blob/master/src/cmd/go/testdata/script/README), but it doesn't seem to work--under `$WORK` I just see a few hundred numbered folders with some unrelated-seeming files in them.

[...]

PR #485, opened 2022-03-16
[microsoft/dev.boringcrypto.go1.18] Improve patches

dagood:
* Improve fix for https://github.com/microsoft/go/issues/481: don't skip the whole test, address the import issue.
* Rearrange patches.
  * Make vendoring easier to reproduce by having the patch that adds the imports get applied *before* the patch that vendors in the library.
  * Better distinguish the changes in each patch. Moved the test fixes together into the last patch and separated out the addition of `crypto/internal/backend` from the changes throughout the codebase that integrated it by changing `import "crypto/internal/boring"` to `import boring "crypto/internal/backend"`.

My tools for rearranging were `git rebase -i`'s `e` mode, and `git reset --soft HEAD~1`.


patches/0005-Adjust-Go-tests-to-work-with-crypto-module.patch
```diff
@@ -0,0 +1,95 @@
...
+ 
+-go list -test -deps -f '{{.ImportPath}} {{.Dir}}' .
++# Use build tag 'gocrypt' while evaluating test dependencies to avoid importing
++# vendored crypto module dependencies like go-crypto-openssl. This test script
++# is not set up to handle any vendored libraries being imported other than
++# golang.org/x/net/http2/hpack, so we must make sure it is the only one.
++#
++# See https://github.com/microsoft/go/issues/481 for more details, such as the
++# dependency chain that would cause the failure if the gocrypt tag isn't used.
++go list -tags=gocrypt -test -deps -f '{{.ImportPath}} {{.Dir}}' .
```
dagood, line 29: Note: I had to add `-tags=gocrypt` to `go list`, not `go build`, because this list isn't actually based on the binary. On the plus side: this is a more minimal change. 😄

dagood:
To verify this PR's changes don't affect the end result, I ran `git go-patch apply` at the PR's commit and at the 1.18 boring branch, grabbed the two resulting temp commit IDs in the submodule, and diffed them, and it gave what I'd expect--only the test fix improvement:

```diff
$ git diff bb913d0f0d 0e82fbf25a
diff --git a/src/cmd/go/testdata/script/gopath_std_vendor.txt b/src/cmd/go/testdata/script/gopath_std_vendor.txt
index 8c39d9a2bd..2d28bd2cf9 100644
--- a/src/cmd/go/testdata/script/gopath_std_vendor.txt
+++ b/src/cmd/go/testdata/script/gopath_std_vendor.txt
@@ -1,7 +1,6 @@
 env GO111MODULE=off
 
 [!gc] skip
-skip 'This test uses a test program that is intended to import no vendored libraries other than golang.org/x/net/http2/hpack, but this is not true in microsoft/go boring 1.18+. https://github.com/microsoft/go/issues/481'
 
 go list -f '{{.Dir}}' vendor/golang.org/x/net/http2/hpack
 stdout $GOPATH[/\\]src[/\\]vendor
@@ -24,7 +23,14 @@ go list -deps -f '{{.ImportPath}} {{.Dir}}' .
 stdout $GOPATH[/\\]src[/\\]vendor[/\\]golang.org[/\\]x[/\\]net[/\\]http2[/\\]hpack
 ! stdout $GOROOT[/\\]src[/\\]vendor
 
-go list -test -deps -f '{{.ImportPath}} {{.Dir}}' .
+# Use build tag 'gocrypt' while evaluating test dependencies to avoid importing
+# vendored crypto module dependencies like go-crypto-openssl. This test script
+# is not set up to handle any vendored libraries being imported other than
+# golang.org/x/net/http2/hpack, so we must make sure it is the only one.
+#
+# See https://github.com/microsoft/go/issues/481 for more details, such as the
+# dependency chain that would cause the failure if the gocrypt tag isn't used.
+go list -tags=gocrypt -test -deps -f '{{.ImportPath}} {{.Dir}}' .
 stdout $GOPATH[/\\]src[/\\]vendor[/\\]golang.org[/\\]x[/\\]net[/\\]http2[/\\]hpack
 ! stdout $GOROOT[/\\]src[/\\]vendor

```

qmuntal (approved):

dagood:
* Waiting to merge this until after https://github.com/microsoft/go/pull/493 to avoid causing a conflict that slows down work Mariner's asking for.

Also, I think I should actually do the rearranging separately, starting on 1.16 (or 1.17) and merging through to 1.18. That way, all our branches have the same sequence of patches, which will hopefully make future changes easier to merge through the branches (or at least, I won't be making it worse).

patches/0005-Adjust-Go-tests-to-work-with-crypto-module.patch
```diff
@@ -0,0 +1,95 @@
...
+diff --git a/src/cmd/link/internal/ld/lib.go b/src/cmd/link/internal/ld/lib.go
+index 5b82dc287d..0e88b3ac94 100644
+--- a/src/cmd/link/internal/ld/lib.go
++++ b/src/cmd/link/internal/ld/lib.go
+@@ -1015,7 +1015,7 @@ var hostobj []Hostobj
+ // These packages can use internal linking mode.
+ // Others trigger external mode.
+ var internalpkg = []string{
+-	"crypto/internal/boring",
++	"vendor/github.com/microsoft/go-crypto-openssl/openssl",
```
dagood, line 42: I noticed this doesn't actually seem to be a testing change. Will move it to `Integrate OpenSSL module`.

Issue #241, opened 2021-10-28
windows-amd64 build flakiness: "compile.exe: Access is denied."


dagood:
```
---- Running command: [cmd.exe /c make.bat]
Building Go cmd/dist using C:\Users\VssAdministrator\.go-stage-0\1.16.5\go
Building Go toolchain1 using C:\Users\VssAdministrator\.go-stage-0\1.16.5\go.
Building Go bootstrap cmd/go (go_bootstrap) using Go toolchain1.
Building Go toolchain2 using go_bootstrap and Go toolchain1.
Building Go toolchain3 using go_bootstrap and Go toolchain2.
go install cmd/compile: copying C:\Users\VSSADM~1\AppData\Local\Temp\go-build3426272395\b098\exe\a.out.exe: open D:\a\1\s\pkg\tool\windows_amd64\compile.exe: Access is denied.
go tool dist: FAILED: D:\a\1\s\pkg\tool\windows_amd64\go_bootstrap install -gcflags=all= -ldflags=all= -a -i cmd/asm cmd/cgo cmd/compile cmd/link: exit status 1
```

All instances I've found so far are on 1.17 branches:

https://dev.azure.com/dnceng/internal/_build/results?buildId=1433378&view=logs&j=b2ef2c90-30c2-5075-badc-2feb75a64cee&t=b6826774-f899-5802-9e9c-ca3c12e8a4c0&l=38

https://dev.azure.com/dnceng/internal/_build/results?buildId=1442944&view=logs&j=5a2439d3-40cd-5cda-520e-11af88128fb3&t=a1b03d73-1679-5340-7eef-275cc3387d35&l=83

https://dev.azure.com/dnceng/internal/_build/results?buildId=1433379&view=logs&j=5a2439d3-40cd-5cda-520e-11af88128fb3&t=a1b03d73-1679-5340-7eef-275cc3387d35&l=38

The first thing that comes to mind is some virus scan locking it, potentially related to the recent build pool migrations.

PR #487, opened 2022-03-16
[microsoft/dev.boringcrypto.go1.18] Update VERSION to go1.18

dagood:
Upstream doesn't maintain this file, so we have our copy that we update. This is ingested into the build asset JSON and used to generate the Dockerfiles, etc. in microsoft/go-images.

This is something I'm eager to get into automatic release infrastructure.


jaredpar (approved):

chsienki (approved):

dagood:
windows-amd64-test failed with flakiness, I hit retry:

```
##### moved GOROOT
ok  	fmt	0.097s
go tool dist: failed to restore GOROOT: rename D:\a\1\s\go-moved D:\a\1\s\go: Access is denied.

=== Errors
go tool dist: failed to restore GOROOT: rename D:\a\1\s\go-moved D:\a\1\s\go: Access is denied.

DONE 14480 tests, 1127 skipped, 1 error in 1057.878s
```

This is an example of what we can only assume is a failure caused by antivirus locking. We have this thread about the way it showed up when it was happening in the build stage (now the build stage auto-retries, and we're hitting it in the test stage):

* https://github.com/microsoft/go/issues/241

Just pointing it out because it came up recently in a standup meeting about this PR I have open to add auto-retries to this part of the Windows jobs, too:

* https://github.com/microsoft/go/pull/475

Issue #486, opened 2022-03-16
Support CBL-Mariner bootstrap build with 1.4 source tarball


dagood:
Mariner bootstraps from Go 1.4 so it can run a pure build from source, using the Go 1.4 compiler written in C.

I think we could simply mirror the source code so it comes from a Microsoft location.

dagood:
Done: I just published a `v1.4.0-1` release with a mirror of the source archive:
https://github.com/microsoft/go/releases/tag/v1.4.0-1
https://github.com/microsoft/go/releases/download/v1.4.0-1/go1.4-bootstrap-20171003.tar.gz

PR #496, opened 2022-03-24
[microsoft/release-branch.go1.4] Update 1.4 branch submodule, add branch-specific readme note

dagood:
This PR updates the readme ([rendered](https://github.com/dagood/go/blob/dev/dagood/1.4/README.md)) and updates the submodule to https://github.com/golang/go/tree/go1.4.

I initially also deleted all the infrastructure, but I decided to leave it alone and just update the readme for now. If we actually end up wanting to build this branch (to produce our own source archives?), we could fit that into the infra.

The idea is that I'll merge this PR, tag it as `v1.4.0-1`, add a GitHub release on that tag, and reupload the source from https://go.dev/doc/install/source#bootstrapFromSource as a GitHub release attachment. (GitHub releases is how Mariner wants to get Go source.)

If we need to update the bootstrap sources, we have options. We could add a `v1.4.0-2` tag on the same commit and make a new release to store the new version. Or, I believe we can always go back and add new files to the existing release. For context about updates: the archive filename mentions 2017-10-03, vs. 1.4's release date of Dec 2014.

* For https://github.com/microsoft/go/issues/486

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,63 +15,12 @@ of Go](https://go.dev/dl/).
 Unless otherwise noted, the Go source files are distributed under the
 BSD-style license found in the LICENSE file.
 
-## Why does this fork exist?
+## The microsoft/release-branch.go1.4 branch
 
-The `microsoft/dev.boringcrypto*` branches produce a modified version of Go that
-can be used to build FIPS 140-2 compliant applications. Our goal is to share
-this implementation with others in the Go community who have the same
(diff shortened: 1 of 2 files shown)


qmuntal (approved):

dagood:
Added a change to make the PR job succeed without building, so PRs can be merged into this branch without admin permissions.

chsienki (approved):

PR #513, opened 2022-03-31
[microsoft/dev.boringcrypto.go1.16] Rearrange FIPS patches, start at 0100

dagood:
This starts with the same set of FIPS patch layering improvements as:
* https://github.com/microsoft/go/pull/485

Then adds in a renumbering command so the infra (1.16) patches start at 0001 and FIPS patches start at 0100. 

I ran the same check as I did there: diff before and after shows zero changes.


qmuntal (approved):

dagood:
After the renumbering, you need this git-go-patch update to be able to round-trip the patch files:
* https://github.com/microsoft/go-infra/pull/35

chsienki (approved):

PR #520, opened 2022-04-01
[microsoft/dev.boringcrypto.go1.18] Use FIPS-enabled Mariner image when running CI tests

qmuntal:
This PR changes the CBL-Mariner image used for testing to a FIPS-enabled version.

With this change we will be exercising the OpenSSL backend on every CI run.


qmuntal:
`TestSetgid` hangs forever and makes the test to fail after a reasonable timeout.

Strangely, it only fails in go1.18 when OpenSSL backend is enabled and using the external linker with the `-static` flag set:

```bash
go test -ldflags '-linkmode=external -extldflags "-static -pthread"' -timeout 30s -run TestSetgid ./cgo/test
```

It works fine in go1.16 and go1.17 under the same conditions.

My take is that go1.18 has a regression for the issue [#3871](https://github.com/golang/go/issues/3871), but I'm still investigating it.

qmuntal:
> It works fine in go1.16 and go1.17 under the same conditions.

> My take is that go1.18 has a regression for the issue https://github.com/golang/go/issues/3871, but I'm still investigating it.

Hmmm, I no longer know what to think. I can reproduce the failure in go1.17 locally, yet [go1.17 CI](https://dev.azure.com/dnceng/public/_build/results?buildId=1694479&view=results) finished successfully...

Even weirder, I've found a minimal reproducer which just involved calling `dlopen` and `C.setgid(0)`, and it doesn't hang when loading OpenSSL 1.0.2 but hangs for all other versions!

```go
package main

// #include <dlfcn.h>
// #include <unistd.h>
// #cgo LDFLAGS: -ldl
import "C"

func main() {
	if C.dlopen(C.CString("libcrypto.so.1.1.0"), C.RTLD_LAZY|C.RTLD_LOCAL) == nil {
		panic("so not found")
	}
	C.setgid(0)
}
```

Command: `go run -ldflags '-linkmode=external -extldflags "-static"' .`

qmuntal:
The same logic written in C works well, so Go must be doing something that OpenSSL don't like:

```c
#include <dlfcn.h>
#include <unistd.h>
#include <stdlib.h>

int main(int argc, char* argv[])
{
    void* p = dlopen("libcrypto.so.1.1.1", RTLD_LAZY|RTLD_LOCAL);
    if (p == NULL) {
        abort();
    }
    setgid(0);
}
```

Command: `gcc -o main main.c -ldl -static`

dagood:
> minimal reproducer

For me, it repros in `mcr.microsoft.com/oss/go/microsoft/golang:1.18-fips-cbl-mariner1.0` with our copy of `go` it contains, but also with the official build of Go 1.18 and official build of Go 1.17.8 if I download them into the container. I'm using the `/usr/lib/libcrypto.so.1.1` binary included in the image. It also repros if I put the code all on the cgo side:

```go
package main

// #include <dlfcn.h>
// #include <unistd.h>
// #cgo LDFLAGS: -ldl
// void doit()
// {
//   void* p = dlopen("libcrypto.so.1.1", RTLD_LAZY|RTLD_LOCAL);
//   setgid(0);
// }
import "C"

func main() {
        C.doit()
}
```

The pure C `gcc` code also doesn't repro in my container.

I tested it in a pure `cblmariner.azurecr.io/base/core:1.0` and `ubuntu:20.04` container with upstream Go 1.17.8 to make sure I didn't have any contamination--also repro'd.

qmuntal:
I've created an upstream issue, as this is clearly not go-crypto-openssl fault: https://github.com/golang/go/issues/52141.

I've been able to minimize even more the reproducer, to the point that libcrypto is no longer needed. See the issue for more info.

qmuntal:
As @dagood suggested offline, 1.16 and 1.17 pipelines are not failing because they are not importing go-openssl-crypto in when running `TestSetgid`. Go 1.18 does import it, so it triggers the bug.

I'll run the offending test with `-tags=gocrypt` until https://github.com/golang/go/issues/52141 is fixed upstream.

patches/0103-Adjust-Go-tests-to-work-with-crypto-module.patch
```diff
@@ -3,12 +3,50 @@ From: microsoft-golang-bot <microsoft-golang-bot@users.noreply.github.com>
...
+diff --git a/misc/cgo/test/pkg_test.go b/misc/cgo/test/pkg_test.go
+index 14013a4cd9..83bb5f5323 100644
+--- a/misc/cgo/test/pkg_test.go
++++ b/misc/cgo/test/pkg_test.go
+@@ -50,7 +50,9 @@ func TestCrossPackageTests(t *testing.T) {
+ 		t.Fatal(err)
+ 	}
+ 
+-	cmd := exec.Command("go", "test")
++	// Use build tag 'gocrypt' to avoid importing go-crypto-openssl, which makes TestSetgid to hang.
```
dagood:
```suggestion
+	// Use build tag 'gocrypt' to avoid importing go-crypto-openssl, which makes TestSetgid hang.
```

patches/0103-Adjust-Go-tests-to-work-with-crypto-module.patch
```diff
@@ -3,12 +3,50 @@ From: microsoft-golang-bot <microsoft-golang-bot@users.noreply.github.com>
...
+diff --git a/src/cmd/dist/test.go b/src/cmd/dist/test.go
+index f65d9bf980..564d870af3 100644
+--- a/src/cmd/dist/test.go
++++ b/src/cmd/dist/test.go
+@@ -1195,11 +1195,13 @@ func (t *tester) cgoTest(dt *distTest) error {
+ 				t.addCmd(dt, "misc/cgo/nocgo", t.goTest(), "-ldflags", `-linkmode=external`)
+ 				if goos != "android" {
+ 					t.addCmd(dt, "misc/cgo/nocgo", t.goTest(), "-ldflags", `-linkmode=external -extldflags "-static -pthread"`)
+-					t.addCmd(dt, "misc/cgo/test", t.goTest(), "-tags=static", "-ldflags", `-linkmode=external -extldflags "-static -pthread"`)
++					// Use build tag 'gocrypt' to avoid importing go-crypto-openssl, which makes TestSetgid to hang.
```
dagood:
```suggestion
+					// Use build tag 'gocrypt' to avoid importing go-crypto-openssl, which makes TestSetgid hang.
```

patches/0103-Adjust-Go-tests-to-work-with-crypto-module.patch
```diff
@@ -3,12 +3,50 @@ From: microsoft-golang-bot <microsoft-golang-bot@users.noreply.github.com>
 Date: Thu, 31 Mar 2022 13:11:06 -0500
 Subject: [PATCH] Adjust Go tests to work with crypto module
 
+use gocrypt tag for cgo static tests
```
dagood, line 6: Hmm. Not sure whether we should include an overall summary of each change here, since this file contains some loosely related changes. For most changes, we can have code comments that describe the change in-place, which is great. But others, like the change in the middle of a string in `src/go/build/deps_test.go` can't really have comments all that nearby, so maybe those need more description in the summary--but then it's hard to know which description describes which changes.

I think test changes probably actually deserve their own patch files, so the patch summary can be more tightly linked to specific changes. This also makes it easier to remove changes where we're changed various parts of Go due to a single upstream issue: just remove the whole individual patch file once the issue is fixed, rather than searching for references to the issue in a single patch.

dagood (approved):
LGTM, just a small grammar suggestion, and starting to think we should split up the FIPS test patch--but I don't think this all needs to be resolved in this PR.

patches/0103-Adjust-Go-tests-to-work-with-crypto-module.patch
qmuntal: Agree, but as you also proposed, let's fix this in another PR so we can start testing FIPS mode.

chsienki (approved):

Issue #519, opened 2022-03-31
Add linux-armv6l builders (armhf)


dagood:
* Related: https://github.com/microsoft/go/issues/173

The arm64 hardware we have access to should be able to build this. It looks like Google is using arm64 machines to build their armv6l releases:

https://github.com/golang/build/blob/ce4ab8eee5f9e0633d12454f2e5b19824b73f4a7/dashboard/builders.go#L562-L581
https://github.com/golang/build/tree/d652ff4bfce54f226acee801facd43b1dc1174fb/env/linux-arm/aws

The Docker Hub official golang images use the armv6l archive for armhf images:
https://github.com/docker-library/golang/blob/9f431d40c96988dd4302eee93789fa6f071354d2/1.17/bullseye/Dockerfile#L33-L43

```sh
		'armel') \
			export GOARCH='arm' GOARM='5' GOOS='linux'; \
			;; \
		'armhf') \
			url='https://dl.google.com/go/go1.17.8.linux-armv6l.tar.gz'; \
			sha256='3287ca2fe6819fa87af95182d5942bf4fa565aff8f145812c6c70c0466ce25ae'; \
			;; \
		'arm64') \
			url='https://dl.google.com/go/go1.17.8.linux-arm64.tar.gz'; \
			sha256='57a9171682e297df1a5bd287be056ed0280195ad079af90af16dcad4f64710cb'; \
			;; \
```

dagood:
I've been having trouble getting a hold of a known-good armv6 base image to use in CI.

Upstream's [Dockerfile](https://github.com/golang/build/blob/7e46584255555b85d9b9dabff7edc05649260df2/env/linux-arm/aws/Dockerfile#L5) uses `FROM arm32v6/debian:buster`, which doesn't exist. (Anymore?)

I'm able to `docker pull arm32v5/debian:buster` and `docker pull arm32v7/debian:buster`, but not `arm32v6`.

According to https://github.com/docker-library/official-images#architectures-other-than-amd64, v5 is not on the list of supported architectures. But, I don't know if the omission is intentional.

Ubuntu 18.04 (our current base image for Go builds) only has v7+, `arm32v7/ubuntu:18.04`.

The only Linux distro I see in arm32v6 is `alpine`, but it uses musl instead of glibc so we can't use it here.

Options I see to get *something* shipped:

[...]

PR #523, opened 2022-04-12
Add arm/v6 cross-compiled build

dagood:
* For https://github.com/microsoft/go/issues/519

Includes related changes:
* Update bootstrap Go to 1.18 for `strings.Cut`.
* Add missing err checks in pack code.
* Diff reduction: include directory entries in the tar.gz/zip, use `/` rather than `\` for zips even though they're for Windows.

To check the results, I cherry-picked to 1.18 and compared the file lists on the platforms we're currently building:

```diff
#> diff <(tar -tf go1.18.linux-amd64.tar.gz | sort) <(tar -tf eng/artifacts/bin/go.dev.linux-amd64.tar.gz | sort)
29d28
< go/codereview.cfg

#> diff <(tar -tf go1.18.linux-armv6l.tar.gz | sort) <(tar -tf eng/artifacts/bin/go.dev.linux-armv6l.tar.gz | sort)
29d28
< go/codereview.cfg
802d800
< go/pkg/linux_arm/os/signal/
804,805d801
< go/pkg/linux_arm/os/signal/internal/
< go/pkg/linux_arm/os/signal/internal/pty.a
821d816
< go/pkg/linux_arm/runtime/cgo.a

# (Copied from a non-cross build on a Windows machine.)
#> diff <(unzip -Z1 go1.18.windows-amd64.zip | sort) <(unzip -Z1 eng/artifacts/bin/go.dev.windows-amd64.zip | sort)
29d28
< go/codereview.cfg
598d596
< go/pkg/tool/windows_amd64/api.exe
11366,11367c11364,11365
< go/test/fixedbugs/issue27836.dir/Äfoo.go
< go/test/fixedbugs/issue27836.dir/Ämain.go
---
> go/test/fixedbugs/issue27836.dir/├Дfoo.go
> go/test/fixedbugs/issue27836.dir/├Дmain.go
```

* Missing `go/codereview.cfg` is normal.
* `go/pkg/linux_arm/os/signal/internal/pty.a` is interesting. According to [the source, it's only used to run tests](https://github.com/golang/go/blob/d65a41329ee87f46f35719129d1d4b03d4a07cc8/src/os/signal/internal/pty/pty.go#L9).
* `go/pkg/linux_arm/runtime/cgo.a` is probably missing because of the cross-compile: something like not being able to build cgo without running a C compiler that targets arm/v6. This might not be an issue, because when someone needs cgo their copy of Go might automatically build cgo if this `.a` prebuilt file is missing.
  * Building a simple cgo app works fine without this file. When I run `go tool dist test`, this file is built.
* `go/pkg/tool/windows_amd64/api.exe` is an existing diff. I think it's actually an upstream bug--filed https://github.com/golang/go/issues/52335
* `go/test/fixedbugs/issue27836.dir/├Дfoo.go` is an existing diff, not new to this change. However, it seems to only be a problem with `unzip -Z1`. When I extract the file on Windows, it ends up being `Ä`, not `├Д`.

---

As a *very* basic check that the cross-compiled Go works, I installed Debian Bullseye **armhf** (**arm/v7**) onto my RPi 4 (arm64 chip), and on there, the armv6l build from CI works to:

* Build a basic app that uses cgo
* Run `go tool dist test`.

Our partners are going to be using armhf, so this is probably good enough for now.


chsienki (approved):

eng/utilities.ps1
```diff
@@ -126,6 +126,19 @@ function Invoke-WithRetry([ScriptBlock]$ScriptBlock, [int]$MaxAttempts = 3, [int
...
+  $oldGOARCH = $env:GOARCH
+
+  $env:GOOS = $GOOS
+  $env:GOARCH = $GOARCH
+
+  & $block
+
+  $env:GOOS = $oldGOOS
+  $env:GOARCH = $oldGOARCH
+}
```
jaredpar: consider doing this in a finally block

eng/run.ps1
```diff
@@ -99,12 +99,18 @@ try {
...
-  Write-Host "In '$tool_module', building '$module_local_script_path' -> $tool_output"
-  & (Join-Path $stage0_goroot "bin" "go") build -o $tool_output $module_local_script_path
-  if ($LASTEXITCODE) {
-    Write-Host "Failed to build tool."
-    exit 1
+  # The caller may have passed in GOOS/GOARCH to cross-compile Go. We can't use those values here:
+  # we need to be able to run the tool on the host, so we must always target the host OS/ARCH. Clear
+  # out the GOOS/GOARCH values (empty string) to detect host OS/ARCH automatically for the tools.
+  Invoke-CrossGoBlock "" "" {
+    Write-Host "In '$tool_module', building '$module_local_script_path' -> $tool_output"
```
jaredpar, line 106: One thing to be careful about is how block execution and variable scoping works in powershell. I've gotten burned by this in the past because I made bad assumptions about how name lookup works here. IIRC the variables are not actually captured, aka it's not a lambda, but instead the variable lookup occurs in the context where the block is evaluated. 

Been many years since I got burned here so maybe I'm having bad memory but I would look into this a bit.

eng/_core/cmd/build/build.go
```diff
@@ -271,18 +288,26 @@ func getMaxAttemptsOrExit(varName string, defaultValue int) int {
...
 	}
 	return i, nil
 }
+
+func getEnvOrDefault(varName, defaultValue string) (string, error) {
+	v, ok := os.LookupEnv(varName)
+	if !ok {
+		return defaultValue, nil
+	}
+	if v == "" {
```
jaredpar, line 315: Not sure this is a valid code path. Environment variables in Windows can't be empty strings. Using an empty string as an environment variable value is the equivalent of deleting it. Is this for non-Windows code paths?

dagood: Yep, this will only happen on non-Windows. It's just a sanity check: https://github.com/microsoft/go/pull/475#discussion_r826666904

> rejecting empty string _could_ help if you're trying to set it like `GO_MAKE_MAX_RETRY_ATTEMPTS=$(fun)`, and for some reason `fun` fails rather than sending anything to stdout.

eng/run.ps1
dagood: That's right, no capturing.

In this particular case, capture vs. non-capture doesn't have any impact because `Invoke-CrossGoBlock` doesn't set any variables that are being used in the block, the block is defined right here where it's being passed to the function, and the block doesn't escape. But definitely something to be aware of while modifying these scripts.

I could inline `Invoke-CrossGoBlock` to avoid the weirdness, but we rely on blocks for `Invoke-WithRetry` too, so it wouldn't really remove the need to know about this quirk.

I'm glad we're able to get out of this script and into the _core Go module relatively quickly. 😄

dagood:
I ran an internal build on the latest commit: https://dev.azure.com/dnceng/internal/_build/results?buildId=1718680&view=results, and I tried ingesting it into go-images, but there are a couple things to fix:

The build asset json lists the build as `"env": { "GOARCH": "armv6l", "GOOS": "linux" }`, but it should be `"GOARCH": "arm", "GOARM": "6"`. This requires a fix to go-infra and updating the version in `eng/_util`, so I'll fix that in this PR.

The `dockerupdate` command doesn't update the `manifest.json` file quite right. This fix is just on the go-infra side.

jaredpar (approved):

qmuntal (approved):

dagood:
Build asset JSON from new dev internal build looks good: https://dev.azure.com/dnceng/internal/_build/results?buildId=1727301&view=results. Generates expected armhf results in go-images with latest `dockerupdate`.

Issue #453, opened 2022-03-01
Create/update aka.ms links for binary releases of the Microsoft-built Go compiler


dagood:
dotnet/arcade has functionality to create aka.ms links. We should use that to create a way for people to get the latest build, with various constraints: latest stable, latest in a given major version, latest with openssl crypto, latest `main`, etc.

dagood:
We now have a pipeline to use on release day to generate these.
* https://github.com/microsoft/go-infra/pull/29

Links are generated for all the binary and source archive artifacts:

* https://aka.ms/golang/release/latest/go1.18.linux-amd64.tar.gz
* https://aka.ms/golang/release/latest/go1.18.linux-amd64.tar.gz.sha256
* https://aka.ms/golang/release/latest/go1.18.linux-amd64.tar.gz.sig
* https://aka.ms/golang/release/latest/go1.18.windows-amd64.zip
* https://aka.ms/golang/release/latest/go1.18.windows-amd64.zip.sha256
* https://aka.ms/golang/release/latest/go1.18.src.tar.gz
* https://aka.ms/golang/release/latest/go1.18.src.tar.gz.sha256
* https://aka.ms/golang/release/latest/go1.18.src.tar.gz.sig

There are multiple levels of version granularity:

* https://aka.ms/golang/release/latest/go1.18.linux-amd64.tar.gz
* https://aka.ms/golang/release/latest/go1.18.0.linux-amd64.tar.gz
* https://aka.ms/golang/release/latest/go1.18.0-1.linux-amd64.tar.gz

FIPS too:

* https://aka.ms/golang/release/latest/go1.18-fips.linux-amd64.tar.gz
* https://aka.ms/golang/release/latest/go1.18.0-fips.linux-amd64.tar.gz
* https://aka.ms/golang/release/latest/go1.18.0-1-fips.linux-amd64.tar.gz

At the moment the most floaty version we have is 1.18, nothing that will automatically go from 1.18.x -> 1.19.x -> 1.20.x. Filed https://github.com/microsoft/go/issues/499 to track that.

PR #531, opened 2022-04-14
Add aka.ms link doc, link to it from readme

dagood:
Add a doc that has a table of our aka.ms links. Link to it from the main readme.

* Makes the links accessible from https://github.com/microsoft/go/issues/453

I'm using full URLs for these links because I think that we should keep this table up to date in microsoft/main and let it go out of date in release branches. (This is how upstream manages https://github.com/golang/go/blob/dev.boringcrypto/misc/boring/RELEASES, and it's just naturally how things work with our branching structure.) The full URLs will point you at the microsoft/main doc even if you're looking at the readme of an old branch. I also included a note at the top of the download doc itself to make it easy to get back to microsoft/main if you somehow make it to a release branch's version of the doc.

Rendered: <https://github.com/dagood/go/blob/dev/dagood/akams-readme/eng/doc/Downloads.md>


qmuntal (approved):

eng/doc/Downloads.md
```diff
@@ -0,0 +1,29 @@
...
+<!-- END TABLES -->
+
+> <sup>1</sup> - Validate GPG signatures using our public key:
+> * https://download.microsoft.com/download/f/a/2/fa2420dd-3f08-4621-82cf-5a22abcbc8f9/microsoft-managed-lang-compiler.asc
+> * Shortened link for sharing: https://aka.ms/managed-lang-compiler-public-key
+
+
+The links in this table "float": each one downloads the latest Go release built by Microsoft that matches that column's major release number. For example, a 1.17 link will download 1.17.8-1 if that is the latest version of 1.17 available.
+
+To download a specific version, create a static link by copying one of the above URLs and adding a patch and revision number to the end of the existing version number section. More aka.ms links are generated than the ones that appear in the table, because the table isn't updated after every single build. Browse recent GitHub releases in the microsoft/go repository to see the valid patch and revision numbers to use.
```
chsienki: Nit: could add an example here, such as how to change the url for 1.17 to 1.17.8-1

chsienki (approved):

Issue #428, opened 2022-02-11
How should devs find usage of FIPS incompatible crypto in their code?


dagood:
By default, the OpenSSL-backed crypto implementation in https://github.com/microsoft/go-crypto-openssl falls back to Go crypto when an OpenSSL function is not available. This comes from a crypto board recommendation: https://github.com/microsoft/go/issues/277#issuecomment-969102781. (Note: Just because a function is implemented using OpenSSL doesn't mean the function, or the way the function is used, is FIPS-compliant.)

So, with the goal of checking if an existing codebase will be able to pass FIPS 140-2 certification, how do devs find where fallback happens? Or more broadly, where FIPS compatibility is probably broken?

A possibility is to run the application with a flag that makes the crypto libraries panic when a non-OpenSSL-backed function is executed. This could point out usage in the normal execution path.

You could potentially catch more with a code analysis tool: check for used packages and used funcs. In theory it could also check how the funcs are used--deeper checks than you can do at runtime. (Of course, it's impossible to get perfect analysis, but maybe there are some bad patterns that could be pointed out.) An analyzer also helps for relatively unknown code with tests that might not fully exercise the app's crypto.

/cc @microsoft/golang-compiler We've discussed this kind of thing before, but don't remember where we left it. I didn't spot a tracking issue yet.

qmuntal:
I'm somehow skeptical about the code analysis tool, crypto API is full of interfaces and indirection layers which makes it difficult to build a tool that infers which types are being passed to lower level crypto functions.

Additionally, we are really just a thin wrapper around OpenSSL (and others in the future), so the issues we can catch at compile time vs the issues OpenSSL will catch at runtime make me thing that investing time in an automatic tool will be only useful for a handful of cases.

Luckily Go already provides a knob to turn off support for non-FIPS TLS configuration, that is importing `crypto/tls/fipsonly`. Doing that will be good enough for downstream users who do not use any crypto API other than `crypto/tls`, which basically means using `net/http`.

For the other cases, we could build an small tool that checks the dependency tree looking for calls to crypto APIs which are not backed by OpenSSL, and add a environment variable that throws a panic if it detects any intent of fallback from go-crypto-openssl to Go crypto.

dagood:
> we could build an small tool that checks the dependency tree looking for calls to crypto APIs which are not backed by OpenSSL

[...]

Issue #397, opened 2022-01-31
Resolve TODO about `GOLANG_FIPS` and `GOFIPS` environment variables


dagood:
The code checks for `GOFIPS` and then `GOLANG_FIPS` to see if they specify FIPS mode:

https://github.com/microsoft/go/blob/ea9cc8ea52d0a83c36f55d5ffcfb179d55bbdf9d/patches/0003-Integrate-OpenSSL-crypto-module.patch#L374-L381

`GOLANG_FIPS` is used by Red Hat: https://developers.redhat.com/blog/2019/06/24/go-and-fips-140-2-on-red-hat-enterprise-linux#fips_140_2_validated_cryptography

But `GOFIPS` is more similar to the existing `GO*` variables like `GOROOT`, `GOFLAGS`, etc: https://github.com/microsoft/go/pull/382. No built-in env vars have `GOLANG` in the name.

Which of these variables should we keep? Or both?

gdams:
Presumably we could remove support for `GOLANG_FIPS` now that we've got our own fork of the OpenSSL backend?

dagood:
> now that we've got our own fork of the OpenSSL backend?

I don't think that's a factor here. The env var name is already defined in the Go code, not the backend. I read "compat with the Red Hat Go toolchain" as product-usage compat, not an implementation detail.

If I have it right, we would need to find the business requirement and prove it doesn't apply.

dagood:
And just a note: removing it would be a break in itself. We've [decided previously](https://github.com/microsoft/go-lab/blob/main/docs/adr/0017-remove-systemcrypto-goexperiment.md) that removing config variables is worth breaking users, but it's worth pointing out that we would be leaning on this justification as the motivation to do *anything* here vs. leaving it alone.

PR #532, opened 2022-04-19
[microsoft/dev.boringcrypto.go1.18] Panic if GOFIPS=1 set but FIPS not supported

### LINKED ISSUE #494: Panic if cgo disabled and FIPS requested CGO_ENABLED=0 and GOFIPS=1
Our toolchain requires cgo to interop with OpenSSL, there is no way around it.

Once #493 lands, if cgo is disabled (`CGO_ENABLED=0`), the toolchain will always use Go crypto instead of OpenSSL. This is the expected behavior unless FIPS mode is requested (`GOFIPS=1`), as FIPS won't be available.

In this situation the best we can do is panic with a meaningful message.

qmuntal:
This PR adds a runtime check that verifies the crypto backend does not fallback to Go crypto if FIPS mode is requested via GOFIPS environment variable and FIPS mode is not supported.

Fixes #494
Updates #428

### DIFF
--- a/patches/0100-Add-OpenSSL-crypto-module.patch
+++ b/patches/0100-Add-OpenSSL-crypto-module.patch
@@ -6,15 +6,17 @@ Subject: [PATCH] Add OpenSSL crypto module
 github.com/microsoft/go-infra/cmd/git-go-patch command: patch number 0100
 ---
  src/crypto/internal/backend/backend_test.go  |  30 ++++
+ src/crypto/internal/backend/common.go        |  19 +++
  src/crypto/internal/backend/dummy.s          |  10 ++
  src/crypto/internal/backend/iscgo.go         |  10 ++
- src/crypto/internal/backend/nobackend.go     | 124 ++++++++++++++++
+ src/crypto/internal/backend/nobackend.go     | 123 ++++++++++++++++
  src/crypto/internal/backend/nocgo.go         |  10 ++
- src/crypto/internal/backend/openssl_linux.go | 145 +++++++++++++++++++
+ src/crypto/internal/backend/openssl_linux.go | 141 +++++++++++++++++++
  src/go.mod                                   |   1 +
  src/go.sum                                   |   2 +
- 8 files changed, 332 insertions(+)
+ 9 files changed, 346 insertions(+)
  create mode 100644 src/crypto/internal/backend/backend_test.go
+ create mode 100644 src/crypto/internal/backend/common.go
  create mode 100644 src/crypto/internal/backend/dummy.s
  create mode 100644 src/crypto/internal/backend/iscgo.go
  create mode 100644 src/crypto/internal/backend/nobackend.go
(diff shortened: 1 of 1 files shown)


patches/0100-Add-OpenSSL-crypto-module.patch
```diff
@@ -90,10 +110,22 @@ index 0000000000..4de8d404f8
...
 +	"hash"
 +	"math/big"
++	"runtime"
++	"syscall"
 +)
 +
 +const Enabled = false
 +
++func init() {
++	if v, ok := syscall.Getenv("GOFIPS"); ok && v == "1" {
```
dagood: We still handle `GOLANG_FIPS` in the OpenSSL backend (and might actually end up changing to only support it, with https://github.com/microsoft/go/issues/397), should this have an equivalent check for `GOLANG_FIPS=1` too? Maybe this could be unified with the implementation on the OpenSSL backend by moving the env checks to a shared file?

https://github.com/microsoft/go/blob/a34208fdb5c5d337bd245671eeabb791c94a9746/patches/0100-Add-OpenSSL-crypto-module.patch#L286-L292

patches/0100-Add-OpenSSL-crypto-module.patch
```diff
@@ -90,10 +110,22 @@ index 0000000000..4de8d404f8
...
++		if runtime.GOOS != "linux" {
++			panic("FIPS mode requested (GOFIPS=1) but not supported on " + runtime.GOOS)
++		} else if !iscgo {
++			panic("FIPS mode requested (GOFIPS=1) but not supported with cgo disabled")
++		}
```
dagood: Hmm, no `else` means if someone on Linux with cgo sets `GOFIPS=1` and `-tags gocrypt`, there's no panic, right?

This isn't necessarily bad (we could say gocrypt takes precedence), but it seems like the same type of scenario that should probably also panic to avoid false expectations.

(BTW, appreciate the detailed panic reasons in the messages here, I'm sure it'll help out with debugging enable issues later. 🙂)

qmuntal: > Hmm, no else means if someone on Linux with cgo sets GOFIPS=1 and -tags gocrypt, there's no panic, right?

Yep. If a binary is built with `-tags gocrypt` it is because the developer intent is to always use Go crypto, regardless of the runtime configuration. If we panic when `GOFIPS=1` and the binary is built with `gocrypt`, then we would be forcing all those binaries to explicitly run with `GOFIPS=0` if `GOFIPS=1` system-wide, even for the once that do not care about FIPS, which would be really unfortunate.

dagood: Ok, yeah, that makes sense.

Thinking about it a bit more, a big difference between cgo support and `-tags gocrypt` is that cgo is defined by your toolset, and a particular dev may not even know what toolset they're ending up using. But `-tags gocrypt` has to be very intentionally applied to the build--hard to imagine it being accidental.

patches/0100-Add-OpenSSL-crypto-module.patch
qmuntal: Done! Environment variable retrieval is now in a common file.

patches/0100-Add-OpenSSL-crypto-module.patch
```diff
@@ -53,6 +59,47 @@ index 0000000000..c2c06d3bff
...
++	if v, _ := envgofips(); v == "1" {
++		if runtime.GOOS != "linux" {
++			panic("FIPS mode requested (GOFIPS=1) but not supported on " + runtime.GOOS)
++		} else if !iscgo {
++			panic("FIPS mode requested (GOFIPS=1) but not supported with cgo disabled")
++		}
++	}
++}
++
++func envgofips() (string, bool) {
```
dagood: This should probably be `envGoFIPS`? (Matching general naming convention, and `needFIPS` later.)

dagood (approved):
LGTM! Just a naming suggestion.

patches/0100-Add-OpenSSL-crypto-module.patch
qmuntal: Done!

jaredpar (approved):

PR #561, opened 2022-05-11
Add linux-armv6l to table of aka.ms links

dagood:
We're building this now via cross-compilation. It should be in the table in case someone with an oldish RPi wants to try it out without getting Docker involved.

Rendered: https://github.com/dagood/go/blob/dev/dagood/add-armv6l-akams/eng/doc/Downloads.md


jaredpar (approved):

dagood:
(Merged with one approval: simple doc change.)

PR #596, opened 2022-06-08
Update submodule to latest `master` in `microsoft/main`

ghost:
Hi! I'm a bot, and this is an automatically generated upstream sync PR. 🔃

After submitting the PR, I will attempt to enable auto-merge in the "merge commit" configuration.

For more information, visit [sync documentation in microsoft/go-infra](https://github.com/microsoft/go-infra/tree/main/docs/automation/sync.md).

PR #597, opened 2022-06-13
Update `microsoft/main` to latest upstream with translation layer for NewGCMTLS

dagood:
Addresses a refactor that happened in upstream that broke our build:

* https://github.com/microsoft/go/pull/596

~~This PR fixes the problem with a patch to the openssl backend file to add logic there. We should have this logic in go-crypto-openssl instead. (It would also be less awkward because we wouldn't have to work around the API being semi-non-exported with the type assertion.)~~ Reverted.

This PR uses this new API in go-crypto-openssl to implement `NewGCMTLS`:
* https://github.com/microsoft/go-crypto-openssl/pull/30

### DIFF
--- a/patches/0005-Add-OpenSSL-crypto-module.patch
+++ b/patches/0005-Add-OpenSSL-crypto-module.patch
@@ -10,14 +10,14 @@ Subject: [PATCH] Add OpenSSL crypto module
  .../internal/backend/bbig/big_openssl.go      |  12 ++
  src/crypto/internal/backend/common.go         |  34 ++++
  src/crypto/internal/backend/iscgo.go          |  10 ++
- src/crypto/internal/backend/nobackend.go      | 115 ++++++++++++++
+ src/crypto/internal/backend/nobackend.go      | 116 +++++++++++++
  src/crypto/internal/backend/nocgo.go          |  10 ++
- src/crypto/internal/backend/openssl_linux.go  | 145 ++++++++++++++++++
+ src/crypto/internal/backend/openssl_linux.go  | 159 ++++++++++++++++++
  src/crypto/internal/backend/stub.s            |  10 ++
  src/go.mod                                    |   1 +
  src/go.sum                                    |   2 +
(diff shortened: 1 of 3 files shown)


qmuntal (approved):

dagood:
I'll be changing the name of the merge commit because this PR no longer uses a translation layer. `Update microsoft/main to latest upstream; include NewGCMTLS implementation`. (Leaving the PR title alone to avoid breaking Outlook inboxes.)

patches/0005-Add-OpenSSL-crypto-module.patch
```diff
@@ -31,7 +31,7 @@ Subject: [PATCH] Add OpenSSL crypto module
 
 diff --git a/src/crypto/internal/backend/backend_test.go b/src/crypto/internal/backend/backend_test.go
 new file mode 100644
-index 00000000000..c2c06d3bff8
+index 0000000000..c2c06d3bff
```
chsienki, line 34: Why did all the hashes change? It looks like they just dropped the last digit?

chsienki (approved):

patches/0005-Add-OpenSSL-crypto-module.patch
dagood: Interesting... my guess for the count change would be different Git versions or settings between me and @qmuntal. I'll take a look and see if that's likely, and maybe we can make `git-go-patch` set a specific length for the underlying `git` commands.

In general though, the hashes change when the underlying files change upstream, so I tend to ignore the index hash changes altogether. 😄 Later in the PR is an example of one of those:

```diff
- index b7e59d7edb6..994749c10c8 100644
+ index 1482b22d08..5f545148bb 100644
```

In general, if I made an intentional change to a patch file, I'll commit the file's diff and let Git include these index hash changes. I don't `add` patch files that didn't have any intentional changes. Not a great workflow--I want to make the tooling smarter here. Seems a bit tough, though, especially if we want to continue to avoid writing code that actually parses the patch content.

PR #598, opened 2022-06-22
Support boringcrypto experiment

qmuntal:
This PR adds support for building Go binaries using `GOEXPERIMENT=boringcrypto`. It is important to note that the boringcrypto backend behavior is the same as upstream, most notably, it can't be disabled at runtime via `GOFIPS`.

This PR contains many patch changes, put the real additions are minimal:

- Run CI tests using `GOEXPERIMENT=boringcrypto`. 8ae11617d904dbaf7032748e97e2640f923bef26
- Add the shims for the boringcrypto packages in [0007-Add-BoringSSL-crypto-module.patch](https://github.com/microsoft/go/compare/dev/qmuntal/boring?expand=1#diff-f34df67b80c90da2851b5e0945406257232f32082149de8bfdcc0b87599b78a3)

All other changes are just renaming of error strings, moving a couple of functions from `openssl_linux.go` to `common.go` and splitting `0005-Add-OpenSSL-crypto-module.patch` into `0005-Add-crypto-backend-foundation.patch`+`0006-Add-OpenSSL-crypto-module.patch`.


patches/0005-Add-crypto-backend-foundation.patch
```diff
@@ -0,0 +1,322 @@
...
++// when backend is in use.
++func Unreachable() {
++	if Enabled {
++		panic("cryptobackend: invalid code execution")
++	} else {
++		// Code that's unreachable is exactly the code
++		// we want to detect for reporting standard Go crypto.
++		sig.StandardCrypto()
++	}
++}
```
dagood, line 133: I looked into `sig.StandardCrypto()` and I *think* this change is correct, but it's pretty subtle... the point seems to be that if `sig.StandardCrypto()` is linked into the final executable at all, then `go version` (or https://github.com/rsc/goversion specifically?) will detect the sequence of bytes in the `sig.StandardCrypto()` function itself defined in ASM (the call itself being a NOOP): https://github.com/golang/go/blob/ea5d7cbc2644643331bd675b1ebdf0aaac7419f1/src/crypto/internal/boring/sig/sig_amd64.s#L7-L15

When we have the OpenSSL module enabled, even though the `sig.StandardCrypto()` call is behind an `if`, I believe it will get linked in either way because `Enabled` is a `var`. So a binary compiled with the OpenSSL module will always show up as using standard crypto when inspected statically. This seems correct to me: an OpenSSL module might use standard crypto or might use OpenSSL crypto depending on the runtime environment. The OpenSSL `init` also links in `sig.BoringCrypto()`, so an inspection should show that *both* markers are present.

With the boring module enabled, we have `const Enabled = true` rather than `var`, so I think that Go eliminates the `false` branch and we end up not linking `sig.StandardCrypto()`. This seems right to me. `sig.BoringCrypto()` ends up being linked in via `src/crypto/internal/boring/boring.go`.

With nobackend, we have `const Enabled = false`, so the `true` branch is eliminated (not that it makes a difference in this case) and we end up linking `sig.StandardCrypto()` only. Also sounds right.

Do you know if it works out like this in practice? I'm not totally sure if branch elimination and the linker works how I'm imagining.

dagood (approved):
I think we should confirm the `sig` stuff still works as intended, other than that LGTM. Reviewed using `git diff` on the commits that come out of `git go-patch apply -f` before and after this PR's changes to the patch files.

patches/0005-Add-crypto-backend-foundation.patch
qmuntal: Your branch elimination and the linker work like you described. On the other hand, `go version` does not report the existence of `sig.*()` functions, but it does report `GOEXPERIMENTS`.

I've tested your assumptions about `rsc/goversion`, which does understand `sig.*()` functions, and these are the results:
- `GOEXPRIMENT=opensslcrypto`: Only standard crypto is reported. It does not detect `sig.BoringCrypto()` because the function name created by the linker has changed from `crypto/internal/boring/sig.BoringCrypto` to `crypto/internal/boring/sig.BoringCrypto.abi0`, and `goversion` does an strict comparison against the former (see [code](https://github.com/rsc/goversion/blob/597212e462da05a7902d6cea0ec895a0d9b8b218/version/read.go#L60) and https://github.com/rsc/goversion/issues/20).
- `GOEXPERIMENT=boringcrypto`: Reports both standard and boring crypto. It neither finds `sig.BoringCrypto` due to previous reasoning, but it does find functions starting with `_Cfunc__goboringcrypto_`, which has the same effect. Standard crypto should not be reported because boring is a compile time switch and `sig.StandardCrypto` does not appear in the final binary, but it also looks for [these function names](https://github.com/rsc/goversion/blob/597212e462da05a7902d6cea0ec895a0d9b8b218/version/read.go#L111), and the `sha1` can't branch-eliminate `sha1.digest` due to [this change](https://github.com/microsoft/go/blob/86b601f5fffce395dee5c319017d4e7c029e4f02/patches/0006-Integrate-OpenSSL-module.patch#L408) in our patches.

qmuntal: I've submitted https://go-review.googlesource.com/c/go/+/413754/ upstream so we don't have to redefine `sha1.boringEnabled` as a variable instead of a constant.

I would rather keep `borincrypto` linking `sha1.digest` until upstream accepts my changes, as fixing it on our side would mean more diffs to maintain for a backend that we don't really officially support.

dagood: > * `GOEXPRIMENT=opensslcrypto`: Only standard crypto is reported. It does not detect `sig.BoringCrypto()` because the function name created by the linker has changed from `crypto/internal/boring/sig.BoringCrypto` to `crypto/internal/boring/sig.BoringCrypto.abi0`, and `goversion` does an strict comparison against the former (see [code](https://github.com/rsc/goversion/blob/597212e462da05a7902d6cea0ec895a0d9b8b218/version/read.go#L60) and [With golang 1.17 the symbols are different  rsc/goversion#20](https://github.com/rsc/goversion/issues/20)).

Interesting... seems strange at first glance that there's code to check for the sig byte sequences in `findCryptoSigs`, but that func doesn't end up being used if the code's able to find `runtime.buildVersion` (and a debug flag isn't set). And if it does end up using `findCryptoSigs`, it's *in addition to* the symbol name check.

Seems like there's probably some history here--one is more reliable over the other, or one works with different versions of Go than the other. Just my guess.

At the end of the day, `rsc.io/goversion` isn't part of the Go toolset, so I don't think we need to worry about it all that much. 🙂

chsienki (approved):

PR #614, opened 2022-06-30
Organize patch files as logical units

qmuntal:
This PR is a proposal to reorganize patch files so the changes are grouped by functionality instead of a mix between implementation details and scope of the changes.

The goals are:
- Make it easier to add the CNG backend. Currently the OpenSSL backend implementation is fragmented in different patches, so is hard to understand what it should be done to support the new backend and in which patch file place each added/changed line.
- Reduce number of patch files. There are already 7 patch files related to crypto backend, when adding CNG we would have at least 2 more.

This is what I've done:
- Delete `internal-goexperiment-add-GOEXPERIMENT-opensslcrypto.patch` and move the content to `Add-OpenSSL-crypto-module.patch`.
- Move OpenSSL logic from `Add-crypto-backend-foundation.patch` to `Add-OpenSSL-crypto-module.patch`
- Delete `Use-crypto-backend-instead-of-boring.patch` and move its content to `Add-crypto-backend-foundation.patch`
- Appy `Add-BoringSSL-crypto-module.patch` before `Add-OpenSSL-crypto-module.patch`
- Delete `Adjust-Go-tests-to-work-with-crypto-module.patch` and split its content between ` `Add-crypto-backend-foundation.patch` and `Add-OpenSSL-crypto-module.patch`.

The resulting patches are:
- 0003-Add-crypto-backend-foundation.patch: Adds the backend foundation, the no-backend implementation, and renamed `"crypto/internal/boring"` to `boring "crypto/internal/backend"`.
- 0004-Add-BoringSSL-crypto-module.patch: Implements the boring backend
- 0005-Add-OpenSSL-crypto-module.patch: Implements the openssl backend and updates some tests
- 0006-Vendor-crypto-modules.patch: run `go mod vendor`


patches/0005-Add-OpenSSL-crypto-module.patch
```diff
@@ -1,20 +1,183 @@
 From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
-From: Quim Muntal <qmuntaldiaz@microsoft.com>
-Date: Wed, 22 Jun 2022 12:31:24 +0000
+From: qmuntal <qmuntaldiaz@microsoft.com>
+Date: Thu, 30 Jun 2022 10:06:19 +0200
 Subject: [PATCH] Add OpenSSL crypto module
```
dagood: I think you mentioned in Go sync changing the patch names to call these "crypto backends" rather than modules?

dagood (approved):

patches/0005-Add-OpenSSL-crypto-module.patch
qmuntal: Done!

jaredpar (approved):

PR #645, opened 2022-07-19
Add CNG crypto backend

### LINKED ISSUE #476: Implement FIPS-compatible crypto for Windows using platform-native crypto library
Use a native Windows crypto library (CNG) to implement Go's crypto library in the boring branches, so we can rely on the native library's FIPS compatibility.

This is the same approach we took on Linux, where we rely on OpenSSL crypto's FIPS compatibility.

Progress:

- [x] CI tests: https://github.com/microsoft/go-crypto-winnative/pull/2
- [x] rand: https://github.com/microsoft/go-crypto-winnative/pull/2
- [x] sha: 
  - [x] sha1: https://github.com/microsoft/go-crypto-winnative/pull/2
  - [ ] sha224: https://github.com/microsoft/go-crypto-winnative/issues/4
  - [x] sha256: https://github.com/microsoft/go-crypto-winnative/pull/2
  - [x] sha384: https://github.com/microsoft/go-crypto-winnative/pull/2
  - [x] sha512: https://github.com/microsoft/go-crypto-winnative/pull/2
- [x] hmac: https://github.com/microsoft/go-crypto-winnative/pull/3
- [x] rsa
  - [x] OAEP padding: https://github.com/microsoft/go-crypto-winnative/pull/6
  - [x] PKCS1 padding: https://github.com/microsoft/go-crypto-winnative/pull/6
  - [x] PSS padding: https://github.com/microsoft/go-crypto-winnative/pull/6
  - [x] No padding: https://github.com/microsoft/go-crypto-winnative/pull/6
- [x] ecdsa
- [x] aes
  - [x] ECB mode: https://github.com/microsoft/go-crypto-winnative/pull/5
  - [x] CBC mode: https://github.com/microsoft/go-crypto-winnative/pull/5
  - [x] GCM mode: https://github.com/microsoft/go-crypto-winnative/pull/5
  - [ ] CRT mode: https://github.com/microsoft/go-crypto-winnative/issues/4
- [ ] vendor into Go toolchain: https://github.com/microsoft/go/pull/645

qmuntal:
This PR integrates https://github.com/microsoft/go-crypto-winnative as a Go backend.

Tips for the reviewers:
- The first commit is just adding the CNG backend as if it was 100% compatible with Go crypto. You can review this fast because it is the same as we already did for openssl and boring.
- The second commit adds a new CI job that builds and test Go using `goexperiment=cngcrypto`.
- The third commit is the important one. It contains all the hacks and fallbacks I had to do to fit CNG into Go.

As agreed, we will fallback to Go crypto when CNG does not implement an algorithm or a parameter required by Go. The complete list can be found at https://github.com/microsoft/go-crypto-winnative/issues/4.

⚠️ I have not implemented a fallback for unsupported RSA key lengths. This is a special case, as CNG might want to restrict even more the supported key lengths based on the configured security provider or when running in FIPS mode. If we fallback to Go crypto in such case, we would be bypassing an important security policy.

I'll add an exhaustive documentation of what to expect and how to use CNG backend once this PR lands.

@dagood @jaredpar @chsienki how does `goexperiment=cngcrypto` resonates to you? People will understand that `cngcrypto` is the CNG Windows backend? Alternatives?

Closes #476


dagood:
> how does `goexperiment=cngcrypto` resonates to you? People will understand that `cngcrypto` is the CNG Windows backend? Alternatives?

I think it works--it parallels nicely with `boringcrypto` and `opensslcrypto` being named after the native library they use rather than the platform. Anyone who wants to use the mode should be looking in our docs and understand it fine from there. And even if someone sees it in the wild and doesn't understand, I was able to plop `cngcrypto` into a few search engines and see the Microsoft CNG docs at or near the top.

jaredpar:
I like `cngcrypto` 

>  I was able to plop cngcrypto into a few search engines and see the Microsoft CNG docs at or near the top.

This makes me like it more :smile:

patches/0006-Add-CNG-crypto-backend.patch
```diff
@@ -0,0 +1,1197 @@
...
+ 		var bkey *boring.PrivateKeyRSA
+ 		bkey, err = boringPrivateKey(priv)
+ 		if err != nil {
+@@ -285,7 +292,9 @@ func SignPKCS1v15(random io.Reader, priv *PrivateKey, hash crypto.Hash, hashed [
+ 		return nil, ErrMessageTooLong
+ 	}
+ 
+-	if boring.Enabled {
++	if boring.Enabled &&
++		(!goexperiment.CNGCrypto || (len(priv.Primes) == 2 && hash != crypto.MD5SHA1)) {
```
dagood: Here's what I was mentioning in the Go sync--the inversion:

```
 (!goexperiment.CNGCrypto || (len(priv.Primes) == 2 && hash != crypto.MD5SHA1))
!( goexperiment.CNGCrypto && (len(priv.Primes) != 2 || hash == crypto.MD5SHA1))
```

I think of the right side as a list of unsupported/fallback cases and it's quite a bit easier to read as positives rather than negatives:
* `len(priv.Primes) == 2 && hash != crypto.MD5SHA1)` "does every unsupported case not match"
* `len(priv.Primes) != 2 || hash == crypto.MD5SHA1)` "does any unsupported case match"

Maybe it's just me. I can certainly figure it out either way, so I'm happy to go with whatever makes the most sense for anyone else. 😄

qmuntal: Done! 0236a2e8f5ae5e0fbea38941dcdbef78bd4a623c

patches/0006-Add-CNG-crypto-backend.patch
```diff
@@ -0,0 +1,1181 @@
...
+ 		}
+ 	}
+@@ -36,7 +39,11 @@ func envGoFIPS() (string, bool) {
+ // when backend is in use.
+ func Unreachable() {
+ 	if Enabled {
+-		panic("cryptobackend: invalid code execution")
++		// CNGCrypto has too many stdlib fallbacks,
++		// so Unreachable is not always true.
++		if !goexperiment.CNGCrypto {
```
dagood: I think it's worth elaborating a bit here... my understanding from poking around is that this is because CNGCrypto has more fallbacks than boring/openssl, and I think this is saying that there are *enough* extra fallbacks that if we were to remove `boring.Unreachable()` from each of those code paths, `Unreachable` becomes useless. Is that right?

I assume it's also not worthwhile to make backend-specific Unreachable calls, so they can be tailored to the fallbacks necessary for each backend?

patches/0006-Add-CNG-crypto-backend.patch
```diff
@@ -0,0 +1,1181 @@
...
++	"internal/goexperiment"
+ 	"math/big"
+ 	"testing"
+ )
+@@ -115,6 +116,9 @@ func testKeyBasics(t *testing.T, priv *PrivateKey) {
+ 	}
+ 
+ 	if boring.Enabled {
++		if goexperiment.CNGCrypto && priv.N.BitLen() < 512 {
++			t.Skip("CNGCrypto does not support key sizes lower than 512")
```
dagood: Is it necessary to skip the test rather than let the func fall back to Go crypto?

dagood (approved):
LGTM, just a comment/doc suggestion and a question.

patches/0006-Add-CNG-crypto-backend.patch
qmuntal: This is what we discussed about not falling back when RSA key lengths are unsupported by CNG. See this paragraph from the PR description for more info:

> ⚠️ I have not implemented a fallback for unsupported RSA key lengths. This is a special case, as CNG might want to restrict even more the supported key lengths based on the configured security provider or when running in FIPS mode. If we fallback to Go crypto in such case, we would be bypassing an important security policy.

patches/0006-Add-CNG-crypto-backend.patch
qmuntal: I give it another thought, and it was easier than expected to make backend-specific Unreachable calls: 
358a6e6.

patches/0006-Add-CNG-crypto-backend.patch
qmuntal: We would be testing a code path that is not reachable when using the public API, so skipping the test seems the right approach.

dagood: It seems reachable to me--at least, this is all based on exported APIs, so I can run the same sequence of functions. But yeah, I didn't connect this to:

> unsupported RSA key lengths ... If we fallback to Go crypto in such case, we would be bypassing an important security policy.

I don't know if it's necessary to go into more detail in a test func, but maybe this could be added in for anyone in the future who also doesn't quite make the connection while reviewing the patch (perhaps somewhat standalone): 😄 

```suggestion
+			t.Skip("CNGCrypto does not support key sizes lower than 512 and intentionally fails rather than fall back to Go crypto")
```

dagood (approved):

chsienki (approved):

Issue #104, opened 2021-06-15
Build official build using `buildmode=pie`, relro, stack protection


dagood:
Normally, Go builds without PIE (Position Independent Executable) or other C-style security measures because these attacks are dealt with at a language/runtime level: https://groups.google.com/g/golang-nuts/c/Jd9tlNc6jUE. However, Go compiles to native binaries, so our SDL tooling (`binskim`) treats it like any other binary and scans for these security measures.

Cgo and `unsafe` *may* make it worthwhile to apply the C-style security measures to Go, but this is debatable.

* For PIE mode, you can pass `-buildmode=pie`: https://golang.org/cmd/go/#hdr-Build_modes
* `relro` is automatically enabled by `pie` build mode, since https://go-review.googlesource.com/c/go/+/22687/
* Stack protection doesn't exist for Go, but can be enabled in Cgo using the standard compiler flag: `-fstack-protector` https://github.com/golang/go/issues/21871

dagood:
PIE is enabled by default when Go builds for Windows since go1.15: https://github.com/golang/go/issues/35192. `binskim` checks our Windows builds for PIE and is ok with them.

However, when `binskim` checks linux-amd64, linux-arm64, and linux-armv6l, their binaries are flagged as not PIE.

PR #649, opened 2022-07-22
Add Guardian SDL pipeline after internal builds

### LINKED ISSUE #94: Apply SDL validation task on windows-amd64 outputs
Building for windows-amd64 is tracked by https://github.com/microsoft/go/issues/27.
Applying SDL tasks is WIP on an internal branch, as of writing.

I'm preemptively filing this issue to track the combination of the two: make sure the windows-amd64 assets are processed properly by the SDL tasks. There may be different requirements/issues that come up for Windows vs. Linux.

dagood:
Add SDL validation stage to our internal (official) builds, based on Arcade. See https://github.com/dagood/go/blob/dev/dagood/sdl-stage/eng/compliance/Guardian/README.md

* Fixes #94 

My initial implementation was to include SDL validation as a stage in the internal rolling pipeline, but it added 25 minutes (not including queue time) to the completion time--nearly doubling the build's total time. Instead, I changed SDL validation into a new pipeline that triggers automatically after each internal rolling build. This keeps the extra time out of our servicing release critical path, while still running validation on every build so we know as soon as possible when there's some compliance/security issue detected. (I'll configure AzDO to send an email notification when this pipeline fails. I haven't seen any flakiness yet in my test runs, but we'll see how this goes.)

A bonus of moving this build out of the main rolling build pipeline is that if it *is* flaky, or if it returns a false positive, we can continue the release without having to retry the build or otherwise work around the fail status.

Example run: https://dev.azure.com/dnceng/internal/_build/results?buildId=1898747&view=results

* Submitted https://github.com/dotnet/arcade/pull/10124 in upstream Arcade for the change to `eng/common/templates/job/execute-sdl.yml`.

### DIFF
--- a/eng/compliance/Guardian/BinSkimConfig.xml
+++ b/eng/compliance/Guardian/BinSkimConfig.xml
@@ -11,7 +11,7 @@
   to disable "BA3001.EnablePositionIndependentExecutable.Options".
 -->
 <Properties Type="PropertiesDictionary">
-  <!-- Go doesn't build PIE executables by default: https://github.com/microsoft/go/issues/104 -->
+  <!-- Go doesn't build PIE executables by default on Linux, only on Windows: https://github.com/microsoft/go/issues/104 -->
   <Properties Key="BA3001.EnablePositionIndependentExecutable.Options" Type="PropertiesDictionary">
     <Property Key="RuleEnabled" Value="Disabled" Type="Driver.RuleEnabledState" />
   </Properties>
--- a/eng/compliance/Guardian/BinSkimConfig.xml
+++ b/eng/compliance/Guardian/BinSkimConfig.xml
@@ -1,0 +1,26 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (c) Microsoft Corporation. Use of this source code is governed by a BSD-style license that can be found in the LICENSE file. -->
+<!--
+  To figure out what to put in this file, run this command:
+
+    {nuget-cache}\Microsoft.CodeAnalysis.BinSkim.{version}\tools\netcoreapp3.1\win-x64\BinSkim.exe
+      export-config {target-location}\BinSkimConfigFull.xml
+
+  Then, search in the resulting file to find the XML element with the configuration you want and
+  copy it into this file. For example, searching for "BA3001" (as seen in SDL tool logs) shows how
+  to disable "BA3001.EnablePositionIndependentExecutable.Options".
+-->
+<Properties Type="PropertiesDictionary">
+  <!-- Go doesn't build PIE executables by default: https://github.com/microsoft/go/issues/104 -->
+  <Properties Key="BA3001.EnablePositionIndependentExecutable.Options" Type="PropertiesDictionary">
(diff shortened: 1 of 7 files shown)


eng/compliance/Guardian/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+Internal rolling builds run Guardian and report results.
+
+The microsoft/go implementation of Guardian execution is based on [dotnet/arcade](https://github.com/dotnet/arcade). See [HowToAddSDLRunToPipeline.md](https://github.com/dotnet/arcade/blob/main/Documentation/HowToAddSDLRunToPipeline.md).
+
+# Running Guardian locally on Windows
+
+1. Create a temporary folder, e.g. `C:\temp\sdl`.
+1. Go to
+    https://dev.azure.com/SecurityTools/SecurityIntegration/_packaging?_a=package&feed=Guardian&package=Microsoft.Guardian.Cli&protocolType=NuGet
```
chsienki, line 15: Is this publicly available?

chsienki (approved):

eng/compliance/Guardian/README.md
dagood: No, I'll add a few notes.

eng/compliance/Guardian/BinSkimConfig.xml
```diff
@@ -0,0 +1,26 @@
...
+
+    {nuget-cache}\Microsoft.CodeAnalysis.BinSkim.{version}\tools\netcoreapp3.1\win-x64\BinSkim.exe
+      export-config {target-location}\BinSkimConfigFull.xml
+
+  Then, search in the resulting file to find the XML element with the configuration you want and
+  copy it into this file. For example, searching for "BA3001" (as seen in SDL tool logs) shows how
+  to disable "BA3001.EnablePositionIndependentExecutable.Options".
+-->
+<Properties Type="PropertiesDictionary">
+  <!-- Go doesn't build PIE executables by default: https://github.com/microsoft/go/issues/104 -->
```
qmuntal: In case it makes a difference, windows defaults to PIE since go1.19: https://github.com/golang/go/issues/35192

dagood: Hmm, the commit that closed the issue actually says it's in go1.15, is there something else in go1.19 that ends up enabling it?

But either way, I last touched this file for go1.16 or go1.17, so I'll try a fresh run without this exception and see what happens.

qmuntal: Wops, I meant 1.15.

qmuntal (approved):

eng/compliance/Guardian/BinSkimConfig.xml
dagood: BinSkim sees that the Windows binaries are PIE, but it complains about the Linux binaries. (Every arch.) I added a note to https://github.com/microsoft/go/issues/104 and I'll add some more detail to this comment. I think the exception has to stay for now--I haven't been able to find a way to filter a rule for some files and not others.

https://dev.azure.com/dnceng/internal/_build/results?buildId=1905501&view=results

PR #653, opened 2022-07-25
Add FIPS mode builders

dagood:
* Add new `fips` optional bool in builder list.
* On any platform, handle `fips: true` by passing run-builder a `-fipsmode` flag that sets `GOFIPS=true` before running tests.
* On Windows, also enable system-wide FIPS using the `registry` package.
* Change the test run names from the job ID mess:  
  `linux_default_default_amd64_test_default_False` to the human-readable name, instead:
  `linux-amd64 test`
  * `linux-amd64 test [opensslcrypto] FIPS`
  * `windows-amd64 test [cngcrypto] FIPS`

---

This actually fills a gap in our Linux/OpenSSL tests since the boring branch merge. The boring branch used an image that sets GOFIPS=true:

https://github.com/microsoft/go/blob/ae8300fc957c60070df9846e4571085df9db128b/eng/pipeline/jobs/run-job.yml#L29-L30

And this isn't the case in `microsoft/main`:

https://github.com/microsoft/go/blob/39476b4a06b19a3b241c3c3e7a036632254aec66/eng/pipeline/jobs/run-job.yml#L38-L39

We need to run both in FIPS and not in FIPS in `microsoft/main`, so we need flexibility, and I think putting it in `run-builder` is a reasonable place, rather than picking different base images in the yml.

### DIFF
--- a/eng/_util/cmd/run-builder/run-builder.go
+++ b/eng/_util/cmd/run-builder/run-builder.go
@@ -38,6 +38,7 @@ var dryRun = flag.Bool("n", false, "Enable dry run: print the commands that woul
 func main() {
 	var builder = flag.String("builder", "", "[Required] Specify a builder to run. Note, this may be destructive!")
 	var experiment = flag.String("experiment", "", "Include this string in GOEXPERIMENT.")
+	var fipsMode = flag.Bool("fipsmode", false, "Run the Go tests in FIPS mode.")
 	var jUnitFile = flag.String("junitfile", "", "Write a JUnit XML file to this path if this builder runs tests.")
 	var help = flag.Bool("h", false, "Print this help message.")
 
@@ -114,6 +115,18 @@ func main() {
 		// Set GOEXPERIMENT in the environment now that we're using the just-built version of Go.
 		if *experiment != "" {
 			buildutil.AppendExperimentEnv(*experiment)
+		}
+
+		if *fipsMode {
+			env("GOFIPS", "true")
+			// Enable system-wide FIPS if supported by the host platform.
+			restore, err := enableSystemWideFIPS()
+			if err != nil {
+				log.Fatalf("Unable to enable system-wide FIPS: %v\n", err)
+			}
+			if restore != nil {
+				defer restore()
+			}
 		}
 
 		// The tests read GO_BUILDER_NAME and make decisions based on it. For some configurations,
--- a/eng/_util/cmd/run-builder/systemfips_fallback.go
+++ b/eng/_util/cmd/run-builder/systemfips_fallback.go
@@ -6,8 +6,11 @@
 
 package main
 
-// enableSystemWideFIPS is a no-op because the current platform either doesn't support or doesn't
-// require system-wide FIPS to be enabled to run tests.
+import "log"
+
+// enableSystemWideFIPS fallback is a no-op because the current platform either doesn't support or
+// doesn't require system-wide FIPS to be enabled to run tests.
 func enableSystemWideFIPS() (restore func(), err error) {
+	log.Println("Using fallback (no-op) for enableSystemWideFIPS. It either isn't supported on this platform or isn't necessary.")
 	return nil, nil
 }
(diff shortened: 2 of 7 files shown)


eng/_util/cmd/run-builder/run-builder_windows.go
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//go:build windows
```
qmuntal: Inline build tags are normally omitted if the constraint is already covered in the file name.

eng/_util/cmd/run-builder/run-builder_windows.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"fmt"
+	"log"
+
+	"golang.org/x/sys/windows/registry"
+)
+
+func init() {
+	enableSystemWideFIPS = func() (func(), error) {
```
qmuntal: On windows it is not possible to enable FIPS mode in a per-process basis, so this function affects all processes running in the same VM. I'm wondering if this could cause issues in case there are parallel non-fips jobs running in the same VM. I'm probably overthinking, and Azure spins a new VM for every job, but asking just in case.

eng/_util/cmd/run-builder/run-builder_windows.go
dagood: Ah, right, slipped my mind that the filename is actually a constraint (rather than just a naming convention).

eng/_util/cmd/run-builder/run-builder_windows.go
dagood: Yeah, each AzDO build always has its own VM, even when we are also running in a container. We eventually got to this point for security and cross-repo-interference reasons. There is a higher throughput system the .NET Engineering team has that I believe shares machines more readily, and there I think we'd need to be more careful, but we aren't using it. ([Helix](https://github.com/dotnet/arcade/blob/main/Documentation/Helix.md))

eng/_util/cmd/run-builder/run-builder.go
```diff
@@ -167,6 +184,13 @@ func main() {
 	}
 }
 
+// enableSystemWideFIPS enables system-wide FIPS if possible and returns a state-restoring func if
+// necessary. Enabling system-wide FIPS is platform-specific and may require build tags. For
+// example, the registry APIs only exist on Windows. This var is replaced by a platform-specific
+// func in a build-tagged file. If the var is nil, run-builder has no implementation for enabling
+// system-wide FIPS.
+var enableSystemWideFIPS func() (restore func(), err error)
```
dagood: Note from Go sync: this `var` func pointer and the `init` defined in `run-builder_windows.go` was just me being lazy and not wanting to write a `run-builder_fallback.go` file. 😄 Removing in favor of conventional "real" `func`.

qmuntal (approved):

qmuntal:
CNG PR has been merged. Would be good to rebase and run test again.

eng/_util/cmd/run-builder/systemfips_fallback.go
```diff
@@ -0,0 +1,13 @@
...
+
+//go:build !windows
+// +build !windows
+
+package main
+
+// enableSystemWideFIPS is a no-op because the current platform either doesn't support or doesn't
+// require system-wide FIPS to be enabled to run tests.
+func enableSystemWideFIPS() (restore func(), err error) {
+	return nil, nil
```
chsienki, line 15: Do we want to log here? Just so we know that the fallback is being used and in case for some reason it gets called on windows?

chsienki (approved):

eng/_util/cmd/run-builder/systemfips_fallback.go
dagood: Wouldn't hurt! I think lack of logs is technically enough as far as diagnosis goes (it will always log *something* in the other case), but I do think it'd be worth it to call it out in case it catches the eye.

PR #654, opened 2022-07-26
Update main FIPS doc for 1.19

dagood:
I tried adjusting the main FIPS doc to match the changes that are coming in 1.19. @qmuntal feel free to push any changes or more info you think this should have into the branch. 🙂

Rendered: https://github.com/microsoft/go/blob/dev/dagood/fips-doc-1.19/eng/doc/fips/README.md

PR #655, opened 2022-07-26
Always use OpenSSL backend when opensslcrypto is set

### LINKED ISSUE #641: Always use OpenSSL backend when goexperiment=opensslcrypto
We are currently only using the OpenSSL backend if `goexperiment=opensslcrypto` and FIPS mode is requested, either via `GOFIPS=1` or system-wide setting. In all other situations we fallback to Go crypto.

We inherited this approach from RedHat fork and I initially though it would be good to keep it for the following reasons:
- We don't have to support using OpenSSL in non-FIPS mode.
- An application would still work in case it could not load the OpenSSL libraries, either because the system doesn't have them or because it is using an unsupported version, unless `GOFIPS=1`.

On the other hand, there are some downsides:
- `boring.Enabled` is no longer constant, because the OpenSSL backend can be disabled at runtime in some situations. This hurts the Go reachability, inlining and escape analyses, because it can't eliminate non-OpenSSL code paths even when `goexperiment=opensslcrypto`.
- We diverge from the Boring approach, where `goexperiment=boringcrypto` always uses the boring backend.

The OpenSSL backend is now more robust, better tested and has a smarter OpenSSL library loading strategy, so I'm leaning towards always using the OpenSSL backend so that `boring.Enabled` can be made constant and we diverge less from the Boring approach.

If we do this change, applications running with `GOFIPS=1` would still panic if OpenSSL can't be load or FIPS can't be configured.

@dagood @jaredpar @chsienki thoughts?

qmuntal:
This PR changes how OpenSSL backend and FIPS more are enabled:

- OpenSSL backend is always used when `goexperiment=opensslcrypto` is set. We were previously falling back to Go crypto if OpenSSL shared libraries were not found or OpenSSL could not be initialized.
- The process will be aborted if FIPS mode can't be enabled and either `GOFIPS=1` or FIPS mode is enabled system system-wide.
- The process will be aborted if FIPS mode can't be disabled and `GOFIPS=0`.
- The FIPS mode will be whatever is configured in OpenSSL if neither GOFIPS nor system-wide FIPS are set.

This PR also updates the openssl and boring shims. They are defined as functions instead of variables. The compiler missed some optimizations because function variables can be redefined, therefore they are not constant.

Updates #654 (@dagood)
Updates #641

### DIFF
--- a/patches/0004-Add-BoringSSL-crypto-backend.patch
+++ b/patches/0004-Add-BoringSSL-crypto-backend.patch
@@ -1,12 +1,12 @@
 From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
 From: Quim Muntal <qmuntaldiaz@microsoft.com>
 Date: Wed, 22 Jun 2022 12:16:05 +0000
-Subject: [PATCH] Add BoringSSL crypto module
+Subject: [PATCH] Add BoringSSL crypto backend
 
 ---
- .../internal/backend/bbig/big_boring.go       | 12 ++++
- src/crypto/internal/backend/boring_linux.go   | 59 +++++++++++++++++++
- 2 files changed, 71 insertions(+)
+ .../internal/backend/bbig/big_boring.go       |  12 ++
+ src/crypto/internal/backend/boring_linux.go   | 116 ++++++++++++++++++
+ 2 files changed, 128 insertions(+)
  create mode 100644 src/crypto/internal/backend/bbig/big_boring.go
  create mode 100644 src/crypto/internal/backend/boring_linux.go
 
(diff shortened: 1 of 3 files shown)


dagood:
> * The process will be aborted if FIPS mode can't be enabled and either `GOFIPS=1` or FIPS mode is enabled system system-wide.

Thinking about this in terms of how we would doc it in https://github.com/microsoft/go/pull/654: I prefer `panics` rather than `will be aborted` because it's more direct--what the Go program does, rather than something that will be done to it. (Aborted by what, when?) It's also somewhat more specific, in case anyone cares. Do you think `panics` would work?

qmuntal:
> Thinking about this in terms of how we would doc it in https://github.com/microsoft/go/pull/654: I prefer panics rather than will be aborted because it's more direct--what the Go program does, rather than something that will be done to it. (Aborted by what, when?) It's also somewhat more specific, in case anyone cares. Do you think panics would work?

Panic is clearly more understandable in this context.

dagood (approved):

chsienki (approved):

PR #684, opened 2022-08-03
Update download table for 1.19, no distinct FIPS

dagood:
Update the download table. Now that 1.19 includes compile-time backends, instead of `1.18 | 1.18-fips` I added `1.19 (+FIPS)`.

I think that once 1.20 comes out, we can simplify to `1.20 | 1.19` without `(+FIPS)`. For now with `1.19 | 1.18 | 1.18-fips`, it would appear that 1.19 is "missing" FIPS, and I want to make sure it's clear that isn't what's going on.

https://github.com/microsoft/go/blob/dev/dagood/dl-table-1.19/eng/doc/Downloads.md


dagood:
Enabling auto-merge for 1 approval: just a doc change plus simple tool change to generate that doc change.

chsienki (approved):

PR #687, opened 2022-08-05
Add cmdscan to elevate logs to warnings

dagood:
`cmdscan` runs a command and listens to its output to pull out warnings for runfo.

These are only warnings because if the retry works, the original error might not actually cause the build to fail.

Example build:

https://dev.azure.com/dnceng/public/_build/results?buildId=1924212&view=results

> image

```
Building Go toolchain3 using go_bootstrap and Go toolchain2.
go install cmd/compile: copying C:\Users\VSSADM~1\AppData\Local\Temp\go-build2959388607\b103\exe\a.out.exe: open D:\a\1\s\go\pkg\tool\windows_amd64\compile.exe: Access is denied.
##[warning]"access denied" (https://github.com/microsoft/go/issues/241): go install cmd/compile: copying C:\Users\VSSADM~1\AppData\Local\Temp\go-build2959388607\b103\exe\a.out.exe: open D:\a\1\s\go\pkg\tool\windows_amd64\compile.exe: Access is denied.
##[debug]Processed: ##vso[task.logissue type=warning]"access denied" (https://github.com/microsoft/go/issues/241): go install cmd/compile: copying C:\Users\VSSADM~1\AppData\Local\Temp\go-build2959388607\b103\exe\a.out.exe: open D:\a\1\s\go\pkg\tool\windows_amd64\compile.exe: Access is denied.
Found pattern '(?i)Access is denied'
go tool dist: FAILED: D:\a\1\s\go\pkg\tool\windows_amd64\go_bootstrap install -a -i cmd/asm cmd/cgo cmd/compile cmd/link: exit status 1
---- Attempt failed with error: exit status 1
```

This tracking issue automatically gets updates from runfo:

* https://github.com/dagood/go/issues/3

(To avoid adding runfo to the Microsoft GitHub org, I've just set it up with my own user/repo for now.)

We could also add other patterns to the tool if it makes it easier to debug common test failures, even if they aren't related to flakiness.

### DIFF
--- a/eng/_core/cmd/cmdscan/cmdscan.go
+++ b/eng/_core/cmd/cmdscan/cmdscan.go
@@ -1,0 +1,101 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// Cmdscan runs the command given to it in os.Args and monitors the output for text patterns that
+// should be elevated to AzDO Pipeline warnings, and uses AzDO logging commands to do so. Timeline
+// events can be discovered more easily in the UI and by automated tools like runfo.
+//
+// If the pattern is associated with a known issue, the warning includes a link.
+//
+// Uses the "log issue" pipeline logging command: https://docs.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands?view=azure-devops&tabs=bash#logissue-log-an-error-or-warning
+package main
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"log"
+	"os"
+	"os/exec"
+	"regexp"
+)
+
+var filters = []filter{
+	{
+		"access denied",
+		"https://github.com/microsoft/go/issues/241",
+		regexp.MustCompile("(?i)Access is denied"),
+	},
+}
+
+type filter struct {
+	// name is a simple description of the detected error that should be uniquely searchable so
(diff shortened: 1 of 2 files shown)


qmuntal (approved):

eng/_core/cmd/cmdscan/cmdscan.go
```diff
@@ -0,0 +1,101 @@
...
+	"io"
+	"log"
+	"os"
+	"os/exec"
+	"regexp"
+)
+
+var filters = []filter{
+	{
+		"access denied",
```
chsienki, line 25: I think this is fine for now, but feels a little odd to have them hard coded into source. As a future step, could we pass these in on the command line, then store them as a variable group in yaml that gets passed in by the CI? Feels like the actual set of issues belongs more with the infra than with code?

chsienki (approved):

eng/_core/cmd/cmdscan/cmdscan.go
dagood: Hmm, hadn't considered a variable group. Yeah, it would be nice to specify patterns in one place for all branches rather than per-branch. Having args (other than passing all args through to the inner command) makes the code a little more complex, but with a `--` separator, could avoid ambiguities, at least. I'll do that in a follow-up PR.

PR #691, opened 2022-08-10
Use variable group for cmdscan rules

dagood:
* Followup from https://github.com/microsoft/go/pull/687

We now have a variable group to hold our scan rules, so we can change it at any time across all branches without touching our code. (Unfortunately, the variable groups in the internal and public projects are isolated. However, since runfo is only looking at the public data, it's not terrible if the internal side gets out of date.)

* https://dev.azure.com/dnceng/internal/_library?itemType=VariableGroups&view=VariableGroupView&variableGroupId=162&path=go-cmdscan-rules
* https://dnceng.visualstudio.com/public/_library?itemType=VariableGroups&view=VariableGroupView&variableGroupId=163&path=go-cmdscan-rules

These are injected into the job as env vars like this:

```
GO_CMDSCAN_RULE_ACCESS_DENIED='{"pattern": "(?i)Access is denied", "url": "https://github.com/microsoft/go/issues/241"}'
```

Cmdscan takes `-envprefix GO_CMDSCAN_RULE_` and looks for all env vars with that prefix to parse as rules. This makes it easy to add new rules: just add a new variable group entry. (Don't need to also modify a list that points to each env var.)

I tried to make sure the command won't crash given bad input so we don't break CI by tweaking the variable groups incorrectly. Also, if one rule is bad, the others still get loaded.

### DIFF
--- a/eng/_core/cmd/cmdscan/cmdscan.go
+++ b/eng/_core/cmd/cmdscan/cmdscan.go
@@ -1,52 +1,131 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-// Cmdscan runs the command given to it in os.Args and monitors the output for text patterns that
-// should be elevated to AzDO Pipeline warnings using AzDO logging commands. Timeline events can be
-// discovered more easily in the UI and by automated tools like runfo.
-//
-// If the pattern is associated with a known issue, the warning includes a link.
-//
-// Uses the "log issue" pipeline logging command: https://docs.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands?view=azure-devops&tabs=bash#logissue-log-an-error-or-warning
 package main
 
 import (
 	"bufio"
+	"encoding/json"
+	"errors"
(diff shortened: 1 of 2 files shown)


qmuntal (approved):

chsienki (approved):

dagood:
By the way, if you ever want to make a variable group in the Public dnceng AzDO project, permissions are locked down and it's not something the team takes lightly, because if it affects the build (result/output) in any way it introduces a reproducibility problem. (And, of course, secrets are a no-go altogether.) My convo in FR Teams: https://teams.microsoft.com/l/message/19:afba3d1545dd45d7b79f34c1821f6055@thread.skype/1660167101453?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=4d73664c-9f2f-450d-82a5-c2f02756606d&parentMessageId=1660167101453&teamName=.NET%20Core%20Eng%20Services%20Partners&channelName=First%20Responders&createdTime=1660167101453&allowXTenantAccess=false

PR #692, opened 2022-08-11
Update OpenSSL FIPS user guide

qmuntal:
While working on the CNG user guide I noticed that the OpenSSL user guide was missing the `crypto/rsa` package and that it was not up-to-date with latest go-crypto-openssl changes.

This PR fixes the previous issues.

Rendered markdown: https://github.com/microsoft/go/blob/dev/qmuntal/openssl-docs/eng/doc/fips/UserGuide.md

### DIFF
--- a/eng/doc/fips/UserGuide.md
+++ b/eng/doc/fips/UserGuide.md
@@ -60,6 +60,18 @@ The Go crypto documentation is available online at https://pkg.go.dev/crypto.
       - [func Sum512](#func-sum512)
       - [func Sum512_224](#func-sum512_224)
       - [func Sum512_256](#func-sum512_256)
+    - [crypto/rsa](#cryptorsa)
+      - [func DecryptOAEP](#func-decryptoaep)
+      - [func DecryptPKCS1v15](#func-decryptpkcs1v15)
+      - [func DecryptPKCS1v15SessionKey](#func-decryptpkcs1v15sessionkey)
+      - [func EncryptPKCS1v15](#func-encryptpkcs1v15)
+      - [func SignPSS](#func-signpss)
+      - [func VerifyPKCS1v15](#func-verifypkcs1v15)
+      - [func VerifyPSS](#func-verifypss)
+      - [func GenerateKey](#func-generatekey-1)
+      - [func GenerateMultiPrimeKey](#func-generatemultiprimekey)
+      - [func PrivateKey.Decrypt](#func-privatekeydecrypt)
+      - [func PrivateKey.Sign](#func-privatekeysign-1)
     - [crypto/subtle](#cryptosubtle)
     - [crypto/tls](#cryptotls)
 
@@ -87,9 +99,9 @@ NewCipher creates and returns a new [cipher.Block](https://pkg.go.dev/crypto/cip
 
 `cipher` implements the cipher.Block interface using an OpenSSL cipher function that depends on the `key` length:
 
-- If `len(key) == 16` then the cipher used is [EVP_aes_128_ecb](https://www.openssl.org/docs/man3.0/man3/EVP_aes_128_ecb.html).
-- If `len(key) == 24` then the cipher used is [EVP_aes_192_ecb](https://www.openssl.org/docs/man3.0/man3/EVP_aes_192_ecb.html).
-- If `len(key) == 32` then the cipher used is [EVP_aes_256_ecb](https://www.openssl.org/docs/man3.0/man3/EVP_aes_256_ecb.html).
+- If `len(key) == 16` then the cipher used is [EVP_aes_128_ecb].
+- If `len(key) == 24` then the cipher used is [EVP_aes_192_ecb].
+- If `len(key) == 32` then the cipher used is [EVP_aes_256_ecb].
 
 The cipher.Block methods are implemented as follows:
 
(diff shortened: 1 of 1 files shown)


keithmattix:
Thanks for this! I was looking for the RSA stuff; glad to see this PR. Quick question: is there any value in adding an x509 section as well?

qmuntal:
> Thanks for this! I was looking for the RSA stuff; glad to see this PR. Quick question: is there any value in adding an x509 section as well?

Thanks for reaching and glad that this guide is being helpful to you! Feel free to request any clarification via PR comment or a separate issue.

About `crypto/x509`, I'll give it a thought and add some docs for it in another PR, although it would not cover all the API, just the ones related to encryption and signing. Note that `crypto/x509` has not been modified as part of the FIPS support, but it does use crypto primitives that are under the FIPS scope, so I would not enter in implementation details but just API restrictions.

eng/doc/fips/UserGuide.md
```diff
@@ -505,11 +518,11 @@ New returns a new hash.Hash computing the SHA1 checksum.
 
 **Return values**
 
-The hash.Hash methods are implemented as follows:
+The hash.Hash methods are implemented usingas follows:
```
dagood: Perhaps a find-replace gone wrong? 😄 A few instances of this:

```suggestion
The hash.Hash methods are implemented as follows:
```

eng/doc/fips/UserGuide.md
```diff
@@ -645,6 +658,204 @@ sha512.Sum512_224 is not FIPS compliant.
...
+```
+
+DecryptOAEP decrypts ciphertext using RSA-OAEP.
+
+**Parameters**
+
+`h` must be the result of one of the following functions in order to be FIPS compliant: sha1.New(), sha224.New(), sha256.New(), sha384.New(), or sha512.New().
+If this invariant is not met, DecryptOAEP won't be FIPS compliant but still will decrypt the message.
+
+`rand` is not used.
```
dagood, line 678: This made me wonder if it's not used because it's not necessary (if OpenSSL uses its own randomness to protect against side-channel attacks) or if it's a limitation that reduces security. I guess the important part as far as the FIPS guide goes is that it doesn't affect FIPS compliance, so the answer doesn't necessarily have to be in the guide. 😄

(Out of curiosity, it looks like OpenSSL's RSA blinding is handled by setting some state, instead, although it's deprecated in 3.0: https://www.openssl.org/docs/man3.0/man3/RSA_blinding_on.html. Not sure what the deal is in 3.0.)

keithmattix:
> Thanks for reaching and glad that this guide is being helpful to you! Feel free to request any clarification via PR comment or a separate issue.
> 
> About `crypto/x509`, I'll give it a thought and add some docs for it in another PR, although it would not cover all the API, just the ones related to encryption and signing. Note that `crypto/x509` has not been modified as part of the FIPS support, but it does use crypto primitives that are under the FIPS scope, so I would not enter in implementation details but just API restrictions.

Thanks so much! Yep, API restrictions are all I'm really looking for; like you said, x509 doesn't directly fall under the purview of FIPS 140-2, but it utilizes primitives that do, so some high level guidance would be super beneficial :+1

eng/doc/fips/UserGuide.md
```diff
@@ -645,6 +658,204 @@ sha512.Sum512_224 is not FIPS compliant.
...
+func rsa.SignPKCS1v15(rand io.Reader, priv *rsa.PrivateKey, hash crypto.Hash, hashed []byte) ([]byte, error)
+```
+
+SignPKCS1v15 calculates the signature of hashed using RSASSA-PKCS1-V1_5-SIGN from RSA PKCS #1 v1.5.
+
+**Parameters**
+
+`rand` is not used.
+
+`hash` can be one of the following values: crypto.MD5, crypto.MD5SHA1, crypto.SHA1, crypto.SHA224, crypto.SHA256, rypto.SHA384, or crypto.SHA512. Else SignPKCS1v15 will fail.
```
dagood: I would say "must be", to make it clear from the start that they're the only options.

I also think it might be a good idea to use the same wording for "Else" as "If this invariant is not met" elsewhere, to make it easier to spot these clauses when reading through. (I don't have a strong opinion on whether it'd be better to change "If this invariant is not met" to "Else" or the other way around. I'm not used to seeing "Else" at the beginning of a sentence, but it certainly makes sense.)

```suggestion
`hash` must be one of the following values: crypto.MD5, crypto.MD5SHA1, crypto.SHA1, crypto.SHA224, crypto.SHA256, rypto.SHA384, or crypto.SHA512. Else SignPKCS1v15 will fail.
```

eng/doc/fips/UserGuide.md
```diff
@@ -645,6 +658,204 @@ sha512.Sum512_224 is not FIPS compliant.
...
+```
+
+Decrypt decrypts `ciphertext` with `priv`.
+
+If `opts` is nil, `priv.Decrypt` is calls `rsa.DecryptPKCS1v15(rand, priv, ciphertext)`.
+If `opts` is of type `*rsa.OAEPOptions`, `priv.Decrypt` calls `rsa.DecryptOAEP(opts.Hash.New(), rand, priv, ciphertext, opts.Label)`.
+If `opts` is of type `*rsa.PKCS1v15DecryptOptions` and `opts.SessionKeyLen > 0`, `priv.Decrypt` calls `rsa.DecryptPKCS1v15SessionKey(rand, priv, ciphertext, plaintext)` with a random `plaintext`.
+If `opts` is of type `*rsa.PKCS1v15DecryptOptions` and `opts.SessionKeyLen == 0`, `priv.Decrypt` calls `rsa.DecryptPKCS1v15(rand, priv, ciphertext)`.
+Else it returns an error.
+Check those function for the parameters restrictions.
```
dagood: This looks kind of messy in rendered form without the linebreaks. Word wrap would still potentially make it hard to read even with forced line ends. Maybe use a bulleted list to direct the eye, and use the opportunity to nest the PKCS1v15DecryptOptions options? (Also, suggestion on rewording last sentence.)

```
- If `opts` is nil, `priv.Decrypt` calls `rsa.DecryptPKCS1v15(rand, priv, ciphertext)`.
- If `opts` is of type `*rsa.OAEPOptions`, `priv.Decrypt` calls `rsa.DecryptOAEP(opts.Hash.New(), rand, priv, ciphertext, opts.Label)`.
- If `opts` is of type `*rsa.PKCS1v15DecryptOptions`:
  - If `opts.SessionKeyLen > 0`, `priv.Decrypt` calls `rsa.DecryptPKCS1v15SessionKey(rand, priv, ciphertext, plaintext)` with a random `plaintext`.
  - If `opts.SessionKeyLen == 0`, `priv.Decrypt` calls `rsa.DecryptPKCS1v15(rand, priv, ciphertext)`.
- Else it returns an error.

Check the function named in the applicable case for the parameters' restrictions.
```

Could also add section/anchor links to each func for easier navigation.

(Wanted to mention that I did see https://rhodesmill.org/brandon/2012/one-sentence-per-line/ linked from https://tip.golang.org/doc/comment recently and I do like the idea of semantic linefeeds in general, which might have been the idea here--I just think this bit probably deserves a proper list.)

eng/doc/fips/UserGuide.md
```diff
@@ -362,10 +373,12 @@ GenerateKey generates a public and private key pair.
 
 **Parameters**
 
-`rand` must be boring.RandReader, else Sign will panic. `crypto/rand.Reader` normally meet this invariant as it is assigned to boring.RandReader in the crypto/rand init function.
+`rand` must be boring.RandReader, else GenerateKey will panic. `crypto/rand.Reader` normally meet this invariant as it is assigned to boring.RandReader in the crypto/rand init function.
```
```suggestion
`rand` must be boring.RandReader, else GenerateKey will panic. `crypto/rand.Reader` normally meets this invariant as it is assigned to boring.RandReader in the crypto/rand init function.
```

qmuntal:
Closing in favor of #697, which incorporates the feedback from this PR.

eng/doc/fips/UserGuide.md
qmuntal: OpenSSL RSA blinging is opt-out since 0.9.6. This fact is not documented in the official docs, I've only seen a comment here: https://github.com/openssl/openssl/blob/12ad22dd16ffe47f8cde3cddb84a160e8cdb3e30/crypto/rsa/rsa.h#L204-L210, which matches what's implemented.

I'll extend the `rand` comment noting that blinding does not depend on this parameter.

Issue #700, opened 2022-08-18
Windows tests failing in CI


dagood:
I'm investigating the difference between this PR CI:
https://dev.azure.com/dnceng/public/_build/results?buildId=1946703&view=logs&j=14ce9276-1b71-5587-760e-edce85b6645f
And PR CI that failed on a doc PR:
https://dev.azure.com/dnceng/public/_build/results?buildId=1948832&view=logs&j=14c594f8-3676-59b1-1953-e09c98fbfc28

Example test failures are:
```
##### GOMAXPROCS=2 runtime -cpu=1,2,4 -quick
FAIL	runtime	300.069s
2022/08/18 10:49:46 Failed: exit status 1
go tool dist: FAILED

=== Failed
=== FAIL: runtime TestCgoCallbackGC (unknown)
=== FAIL: runtime TestCgoTracebackContextPreemption (unknown)
=== FAIL: runtime TestRaceSignal (unknown)
=== FAIL: runtime TestCgoCheckBytes (unknown)
...
```

(In some cases the build fails due to timeout rather than test failure because our CI contains auto-retry logic.)

So far I've found that the provisioner has changed and agent name has a different format, but AzDO agent and virtual environment (usual suspects) are not changed:

```diff
-fail
+pass
 Starting: Initialize job
-Agent name: 'Azure Pipelines 26'
-Agent machine name: 'fv-az24-384'
+Agent name: 'Azure Pipelines 80'
+Agent machine name: 'WIN-KRLIO8K7R52'
 Current agent version: '2.206.1'
 Operating System
 Microsoft Windows Server 2019
 10.0.17763
 Datacenter
 Virtual Environment
 Environment: windows-2019
 Version: 20220814.1
 Included Software: https://github.com/actions/runner-images/blob/win19/20220814.1/images/win/Windows2019-Readme.md
 Image Release: https://github.com/actions/runner-images/releases/tag/win19%2F20220814.1
-Runner Image Provisioner
-1.0.0.0-main-20220812-3
+Virtual Environment Provisioner
+1.0.0.0-main-20220805-2
 Current image version: '20220814.1'
```

The CPU is different--the 8171M (fail/new configuration) is an older CPU, but with 104 threads, double that of the successful 8272CL (52 threads):

```diff
-fail
+pass
 ##### Test execution environment.
 # GOARCH: amd64
-# CPU: Intel(R) Xeon(R) Platinum 8171M CPU @ 2.60GHz
[...]

PR #701, opened 2022-08-18
Add longer timeout on Windows

dagood:
* For https://github.com/microsoft/go/issues/700

I think our builders simply run more slowly than upstream's. Start out with increasing the timeout.

https://github.com/golang/go/blob/833367e98af838a2511ee7e4e19dc8f1da7b8ed7/src/cmd/dist/test.go#L553 includes `t.timeout(300)`, which should enable the multiplier to work in this case. (Some dist tests don't have this and would need something like it to be patched in to work.)

### DIFF
--- a/eng/_util/cmd/run-builder/run-builder.go
+++ b/eng/_util/cmd/run-builder/run-builder.go
@@ -10,6 +10,7 @@ import (
 	"log"
 	"os"
 	"os/exec"
+	"strconv"
 	"strings"
 
 	"github.com/microsoft/go/_core/buildutil"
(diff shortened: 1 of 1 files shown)


qmuntal (approved):

qmuntal (approved):

dagood:
Confirmed, the new timeout let this job succeed:

```
##### GOMAXPROCS=2 runtime -cpu=1,2,4 -quick
ok  	runtime	359.114s
```

Example time from upstream, for fun:

https://build.golang.org/log/9d616a828926dec57ef8133ef71a8a56dcfd3ed9
```
##### GOMAXPROCS=2 runtime -cpu=1,2,4 -quick
ok  	runtime	79.846s
```

jaredpar (approved):

Issue #407, opened 2022-02-07
Use 1ES pools for all internal jobs


dagood:
We need to change this:

https://github.com/microsoft/go/blob/9b40a19bd6840ed5c2818e1896d9868603a2a120/eng/pipeline/jobs/run-job.yml#L16-L30

to be more like this:

https://github.com/microsoft/go/blob/9b40a19bd6840ed5c2818e1896d9868603a2a120/eng/pipeline/jobs/sign-job.yml#L19-L22

Note that release branches use a different pool:

https://github.com/microsoft/go/blob/86be7f0711e77709c0629899f75b9b6268e7481b/eng/pipeline/jobs/sign-job.yml#L19-L22

(We should change the hard-coded internal/public pool into a condition, so we don't need to keep track of that when forking new release branches.)

As of writing, we have no specific deadline for this. It depends on https://github.com/dotnet/core-eng/issues/15256.

---

On the Docker side, the .NET Docker infrastructure made these changes to use 1ES pools: https://github.com/dotnet/docker-tools/pull/954/files.

dagood:
I tested out the replacement image in https://github.com/dotnet/core-eng/issues/15256, and it worked, and I've just been told it's now live. This work should be unblocked now. 🎉

dagood:
Done (but different targets) by 1ES PT migration.

PR #709, opened 2022-08-25
Use dnceng/1ES pools; template for common logic

dagood:
Use dnceng / 1ES pools throughout pipelines.

* For https://github.com/microsoft/go/issues/700. The machines in the dnceng pools have more uniform performance than Azure hosted machines (see quotes from the FR thread). I removed the larger Windows timeout (earlier workaround) to make sure this helps the way it should.

* For https://github.com/microsoft/go/issues/407. We never ended up with a deadline for this (like .NET's), but it's something we should do at some point. This seemed like a good opportunity now that it makes sense to put *everything* on dnceng pools for perf/reliability reasons.

* I wrote a doc describing some of the quirks at play here (https://github.com/microsoft/go-infra/pull/61) and linked from here. Should merge that first to avoid a dead link.

Internal build test (for publish + signing jobs particularly): https://dev.azure.com/dnceng/internal/_build/results?buildId=1967499&view=results

### DIFF
--- a/eng/pipeline/stages/pool-core.yml
+++ b/eng/pipeline/stages/pool-core.yml
@@ -1,0 +1,44 @@
+# Copyright (c) Microsoft Corporation.
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file.
+
+# This template defines common pool names for dnceng agents for various platforms.
+
+parameters:
+  - name: public
+    type: boolean
+  - name: servicing
+    type: boolean
+  # The inner template: { template string, parameters object }
+  - name: inner
+    type: object
+
+stages:
+  - template: ${{ parameters.inner.template }}
+    parameters:
+      ${{ insert }}: ${{ parameters.inner.parameters }}
+
+      dncengPool:
+        # Pick pool. https://github.com/dotnet/arcade/blob/0db07252ccb18afdf94820ba6125da6de729ec04/Documentation/AzureDevOps/AzureDevOpsOnboarding.md#agent-queues
+        ${{ if parameters.public }}:
+          ${{ if parameters.servicing }}:
+            name: NetCore1ESPool-Svc-Public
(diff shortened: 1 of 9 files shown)


qmuntal (approved):

eng/pipeline/stages/pool-core.yml
```diff
@@ -0,0 +1,44 @@
...
+  - template: ${{ parameters.inner.template }}
+    parameters:
+      ${{ insert }}: ${{ parameters.inner.parameters }}
+
+      dncengPool:
+        # Pick pool. https://github.com/dotnet/arcade/blob/0db07252ccb18afdf94820ba6125da6de729ec04/Documentation/AzureDevOps/AzureDevOpsOnboarding.md#agent-queues
+        ${{ if parameters.public }}:
+          ${{ if parameters.servicing }}:
+            name: NetCore1ESPool-Svc-Public
+          ${{ if not(parameters.servicing) }}:
```
akoeplinger, line 26: FYI you can do `{{ else }}` now in .yml: https://docs.microsoft.com/en-us/azure/devops/release-notes/2021/sprint-192-update#new-yaml-conditional-expressions

dagood: Holy crap. Thanks!

(Following up on this in another PR.)

dagood:
> I removed the larger Windows timeout (earlier workaround) to make sure this helps the way it should.

On second thought, I'm putting this back in. We're at 200s/300s in this CI run for the test that has been commonly failing:

```
##### GOMAXPROCS=2 runtime -cpu=1,2,4 -quick
ok  	runtime	207.977s
```

I think "more than double" is probably a reasonable rule of thumb when it comes to test timeouts. If time starts increasing (again), it's less disruptive to keep an eye on the pipeline statistics to catch it than hitting hard failures someday.

jaredpar (approved):

PR #722, opened 2022-09-02
[microsoft/dev.boringcrypto.go1.18] Merge 1.18 infra

dagood:
Merge 1.18 into boring 1.18 branch. Two conflicts:

* run-job.yml - This file was deleted and replaced by run-stage.yml. In the boring 1.18 branch, we'd changed the CI image to one that includes GOLANG_FIPS=1, to run tests in FIPS mode. This change is no longer necessary: we have FIPS mode builders that set the env var before building/testing without requiring a different image. So, I left out this one-line change. The 1.18 boring branch is now a little closer to the normal 1.18 branch, and the 1.19+ branches.
* go submodule - As usual, I left it where it was, so it continues to point at the upstream boring branch.


qmuntal (approved):

chsienki (approved):

dagood:
(Failed because this branch doesn't have the pool name change yet. But at least it's reported back to the GitHub UI now. Will need to merge in a bit more from 1.18 before this runs cleanly.)

dagood:
Forcing merge. CI check is green, but branch protection isn't happy about its source. We need the flakiness fix to get the release through.

> image

> Required status check "microsoft-go" was not set by the expected GitHub app.

PR #753, opened 2022-10-03
[microsoft/dev.boringcrypto.go1.18] Update OpenSSL crypto backend

dagood:
Includes one fix:

* https://github.com/microsoft/go-crypto-openssl/compare/v0.1.1...v0.1.2


qmuntal:
I've updated the pinned go-crypto-openssl version to point to the named release `v0.1.2`, which I've just created.

qmuntal (approved):

PR #755, opened 2022-10-04
Update README.md

abhi99555:
Update README.md with grammatical error.


dagood:
Thanks for the contribution! As far as I know, omitting `that` is valid in this case.

dagood:
Let me know if you have a reference on this rule--it's possible I'm wrong. Then I can reopen and take a look.

PR #770, opened 2022-10-11
Upgrade CNG backend and use it for ECDSA P-224

qmuntal:
CNG supports ECDSA P-224 curve after https://github.com/microsoft/go-crypto-winnative/pull/29 landed.

This PR updates the CNG patch to remove the Go crypto fallback when creating ECDSA P-224 private keys.

FIPS docs are also updated to account for the new implementation.

### DIFF
--- a/eng/doc/fips/UserGuide.md
+++ b/eng/doc/fips/UserGuide.md
@@ -487,7 +487,7 @@ GenerateKey generates a public and private key pair.
 
 `priv` is generated using [BCryptGenerateKeyPair].
 
-`priv` [algorithm identifier] is `BCRYPT_ECDSA_ALGORITHM ` and the [named elliptic curve] depends on the value of `c`:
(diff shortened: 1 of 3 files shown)


eng/doc/fips/UserGuide.md
```diff
@@ -487,11 +487,12 @@ GenerateKey generates a public and private key pair.
 
 `priv` is generated using [BCryptGenerateKeyPair].
 
-`priv` curve [algorithm identifier] depends on the value of `c`:
+`priv` [algorithm identifier] is `BCRYPT_ECDSA_ALGORITHM ` and the [named elliptic curve] depends on the value of `c`:
```
```suggestion
`priv` [algorithm identifier] is `BCRYPT_ECDSA_ALGORITHM` and the [named elliptic curve] depends on the value of `c`:
```

dagood (approved):
Do you think this is worth porting to 1.19?

qmuntal:
> Do you think this is worth porting to 1.19?

I would rather not unless someone asks for it. Conceptually backports are just for bug fixes, and this is a neat feature.

jaredpar (approved):

PR #780, opened 2022-10-20
Print OpenSSL version if GOFIPS can't be honored

qmuntal:
This PR adds a trace showing the loaded OpenSSL version before panicking in case `GOFIPS` can't be honored.

It will help debugging OpenSSL configuration issues.


dagood (approved):

qmuntal:
Done in #943

PR #804, opened 2022-11-21
[cng] Support ECDH

qmuntal:
This PR updates the `crypto/ecdh` package so it points to `crypto/internal/backend`. It also implements ECDH support for the CNG and the Boring backends. OpenSSL support will come later, as `go-crypto-openssl` still doesn't support it.

### DIFF
--- a/patches/0003-Add-crypto-backend-foundation.patch
+++ b/patches/0003-Add-crypto-backend-foundation.patch
@@ -7,6 +7,8 @@ Subject: [PATCH] Add crypto backend foundation
  src/crypto/aes/cipher.go                    |   2 +-
  src/crypto/aes/cipher_asm.go                |   2 +-
  src/crypto/boring/boring.go                 |   2 +-
+ src/crypto/ecdh/ecdh.go                     |   2 +-
+ src/crypto/ecdh/nist.go                     |   2 +-
  src/crypto/ecdsa/boring.go                  |   4 +-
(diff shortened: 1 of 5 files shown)


patches/0003-Add-crypto-backend-foundation.patch
```diff
@@ -83,6 +85,73 @@ index 097c37e343fdb8..1cf43edba40359 100644
...
++	if boring.Enabled && !goexperiment.OpenSSLCrypto {
+ 		bk, err := boring.NewPublicKeyECDH(c.name, k.publicKey)
+ 		if err != nil {
+ 			return nil, err
+@@ -196,7 +197,7 @@ func (c *nistCurve[Point]) ecdh(local *PrivateKey, remote *PublicKey) ([]byte, e
+ 	// only be the result of a scalar multiplication if one of the inputs is the
+ 	// zero scalar or the point at infinity.
+ 
+-	if boring.Enabled {
++	if boring.Enabled && !goexperiment.OpenSSLCrypto {
```
dagood: Shouldn't this `goexperiment.OpenSSLCrypto` usage be in the OpenSSL patch rather than the foundation patch?

qmuntal: Yep, good catch.

qmuntal: Fixed!

dagood (approved):

jaredpar (approved):

PR #862, opened 2023-01-30
Update Windows pool to 2022 for new GCC version required after upstream change

dagood:
* https://github.com/microsoft/go/pull/856 is broken because a newer version of GCC is now needed to build with the race checker.
* Upstream change that requires newer GCC due to dependency on `libsynchronization.a`:  
  420197: runtime/race: update race_windows_amd64.syso | https://go-review.googlesource.com/c/go/+/420197
* Upstream Windows builder change to move from GCC 5.3.0 to 11.2.0 (or some system-provided one?):  
  414475: cmd/racebuild: tweak compilers used for windows syso build | https://go-review.googlesource.com/c/build/+/414475

We get GCC from MinGW-w64 in the `1es-windows-2019` image, which is based on https://github.com/actions/runner-images/blob/main/images/win/Windows2019-Readme.md. That doc lists 8.1.0 as the version.

It looks like https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md has 11.2.0 so the first thing to try is changing to 1es-windows-2022.

Note: those docs seem to show GCC versions, not MinGW-w64 versions. The current latest version is 10.0.0. https://www.mingw-w64.org/downloads/

1ES images names are at https://helix.dot.net/#1esPools

### DIFF
--- a/eng/pipeline/stages/run-stage.yml
+++ b/eng/pipeline/stages/run-stage.yml
@@ -193,6 +193,11 @@ stages:
                   $env:PATH = (
                     $env:PATH -split ';' | Where-Object { $_ -ne 'C:\Program Files\Git\usr\bin' }
                   ) -join ';'
+
+                  Write-Host "Removing Chocolatey shim for SWIG to avoid running misc SWIG tests that would fail."
+                  # The Chocolatey shims are located in a single folder in PATH, so we can't change PATH to exclude it.
+                  # Upstream Windows builders don't have SWIG installed, so this makes coverage even.
+                  Remove-Item (Get-Command swig).Source
                 }
 
                 eng/run.ps1 cmdscan -envprefix GO_CMDSCAN_RULE_ -- `
--- a/eng/pipeline/stages/pool-core.yml
+++ b/eng/pipeline/stages/pool-core.yml
@@ -34,9 +34,9 @@ stages:
         # Pick images. https://helix.dot.net/#1esPools
         demands:
           ${{ if parameters.public }}:
-            windows: ImageOverride -equals 1es-windows-2019-open
+            windows: ImageOverride -equals 1es-windows-2022-open
           ${{ else }}:
-            windows: ImageOverride -equals 1es-windows-2019
+            windows: ImageOverride -equals 1es-windows-2022
 
           ${{ if parameters.public }}:
             linux: ImageOverride -equals 1es-ubuntu-2004-open
--- a/patches/0005-Add-OpenSSL-crypto-backend.patch
+++ b/patches/0005-Add-OpenSSL-crypto-backend.patch
@@ -58,11 +58,11 @@ index a9ec6e6bfefb76..01765f01736ccb 100644
  package api
  
 diff --git a/src/cmd/dist/test.go b/src/cmd/dist/test.go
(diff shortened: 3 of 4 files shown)


dagood:
SWIG test failure:

```
##### ../misc/swig/callback
ok  	misc/swig/callback	0.059s
exit status 0xc0000139
FAIL	misc/swig/callback	0.047s
2023/01/30 18:54:56 Failed: exit status 1
```

qmuntal:
> swig/callback

Can reproduce the `misc/swig/callback` failure, but with a different exit status:

```cmd
exit status 0xc0000135
FAIL    misc/swig/callback      0.072s
FAIL
2023/01/31 09:32:19 Failed: exit status 1
```

Will investigate.

Edit: The error I have having is probably not related to the one in the builder. I could fix it by having all the swig dependent DLLs  reachable via PATH.

qmuntal:
From windows [docs](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/596a1078-e883-4972-9bbc-49e60bebca55):

<html>
<body>


0xC0000139 STATUS_ENTRYPOINT_NOT_FOUND | {Entry Point Not Found} The procedure entry point %hs could not be located in the dynamic link library %hs.
-- | --



</body>
</html>

qmuntal:
Curiously, it only fails on the second `swig_callback` test and the only change is that it builds the C binary using following flags: `-flto -Wno-lto-type-mismatch -Wno-unknown-warning-option`.

dagood:
Unfortunately I haven't been able to get a hold of a VM or VM image from the 1ES team. As far as I can tell, our best bet is to add a bunch of diagnostic logging and artifact publishing to the build and keep running it.

eng/pipeline/stages/run-stage.yml
```diff
@@ -193,6 +193,11 @@ stages:
                   $env:PATH = (
                     $env:PATH -split ';' | Where-Object { $_ -ne 'C:\Program Files\Git\usr\bin' }
                   ) -join ';'
+
+                  Write-Host "Removing Chocolatey shim for SWIG to avoid running misc SWIG tests that would fail."
+                  # The Chocolatey shims are located in a single folder in PATH, so we can't change PATH to exclude it.
+                  # Upstream Windows builders don't have SWIG installed, so this makes coverage even.
+                  Remove-Item (Get-Comment swig).Source
```
dagood: I mentioned in the Go sync that I thought we could use the PATH reassignment trick, but it turns out that the SWIG shim the build finds/uses is installed into the shared Chocolatey bin folder with all the other shims. This is where the build finds gcc, etc. so changing PATH doesn't work. Deleting the shim should be fine--our builds are on one-time-use VMs.

I also considered `choco uninstall`, but that takes a while.

Edit: Looking at https://github.com/actions/runner-images/blob/f0cd188d73624503a493326d6b961d200f510323/images/win/toolsets/toolset-2022.json#L169, it actually looks like MinGW is installed as a msys package. I'm not sure where I got the impression that it's installed via Chocolatey. Either way, removing the shim should still be a valid way forward.

qmuntal (approved):

chsienki (approved):

PR #923, opened 2023-05-15
Update FIPS docs post 1.18 EOL

dagood:
Reading this doc again in connection to some internal policy discussions, I thought it was time to get some changes done. I started with 1.18 removal, but found some more things I wanted to do: 😄

* Now that 1.18 is EOL, remove most mentions. Include a link back in time in case it's necessary for some reason.
* Include CNG more consistently in the doc.
* Add more detail to the dynamic linking section.
* Update the OpenSSL supported versions, and make the notes about the degree of support a bit more specific.
* Some misc doc ease-of-use/ease-of-reading changes.

Rendered: https://github.com/microsoft/go/blob/dev/dagood/fips-docs/eng/doc/fips/README.md

### DIFF
--- a/eng/doc/fips/README.md
+++ b/eng/doc/fips/README.md
@@ -23,34 +23,20 @@ In addition to that, the boringcrypto flag also provides a mechanism to restrict
   import _ "crypto/tls/fipsonly"
 ```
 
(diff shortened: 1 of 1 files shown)


jaredpar (approved):

qmuntal (approved):

dagood:
Came up in the Go sync that we should mention the containers here. I'll submit another PR for that.

PR #924, opened 2023-05-16
FIPS doc: add build config details, Docker

dagood:
I started this change just to mention the `-fips-` Docker images, but I added more, trying to think about what people coming to this doc from the internal Microsoft policy-related pages will want to know. I expect that a number of people will primarily be interested in how to get the "Usage: Build" section implemented in their project and not be as interested in reading the full context.

https://github.com/microsoft/go/blob/dev/dagood/fips-docs/eng/doc/fips/README.md

### DIFF
--- a/eng/doc/fips/README.md
+++ b/eng/doc/fips/README.md
@@ -35,18 +35,56 @@ It is important to note that an application built with Microsoft's Go toolchain 
 
 ## Usage: Build
 
-1. Get the Microsoft build of Go. See [the microsoft/go readme](https://github.com/microsoft/go#binary-distribution) for options.
-1. Enable the desired GOEXPERIMENT and build your program:
-    * To build for Linux/OpenSSL, for example:
-      ```sh
-      GOEXPERIMENT=opensslcrypto go build ./myapp
-      ```
-    * To build for Windows/CNG in PowerShell, for example:
-      ```pwsh
-      $env:GOEXPERIMENT = "cngcrypto"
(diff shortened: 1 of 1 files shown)


eng/doc/fips/README.md
```diff
@@ -35,18 +35,56 @@ It is important to note that an application built with Microsoft's Go toolchain
...
-      ```pwsh
-      $env:GOEXPERIMENT = "cngcrypto"
-      go build ./myapp
-      ```
-1. The built program will use the specified platform-provided cryptographic library whenever it calls a Go standard library crypto API, and FIPS compatibility can be enabled at runtime.
+The `GOEXPERIMENT` environment variable is used at build time to select a cryptographic library backend. This modifies the Go runtime included in the program to use the specified platform-provided cryptographic library whenever it calls a Go standard library crypto API. The `GOEXPERIMENT` values are:
+
+- `opensslcrypto` selects OpenSSL, for Linux
+- `cngcrypto` selects CNG, for Windows
+- Default: uses Go standard library cryptography
```
qmuntal: Worth noting that `GOEXPERIMENT` can contain multiple comma-separated values, e.g., `GOEXPERIMENT=opensslcrypto,loopvar`. It is not allowed to combine `opensslcrypto` and `cngcrypto`, and the standard library crypto is used if non of those appears in the list.

qmuntal (changes requested):

qmuntal (approved):

PR #936, opened 2023-06-01
ed25519: remove stale skip in TestAllocations

qmuntal:
The `TestAllocations` test in `ed25519_test.go` had to be skipped because `boring.Enabled` used to be variable instead of a constant, when using the OpenSSL backend, and that was blocking some inlining optimizations.

We made `boring.Enabled` constant long time ago, so it is no longer necessary to skip that test.


dagood (approved):

qmuntal:
Wops, merged this without a second review... It's just a test change, so I guess it's fine.

PR #943, opened 2023-06-07
Switch openssl backend to golang-fips/openssl

qmuntal:
`github.com/golang-fips/openssl` is almost API-compatible with  `github.com/microsoft/go-crypto-openssl`, so there are only a handful changes:
- `openssl.DecryptRSAOAEPWithMGF1Hash` and `openssl.EncryptRSAOAEPWithMGF1Hash` are not defined, but `openssl.DecryptRSAOAEP` and `openssl.EncryptRSAOAEP` supports the purpose of the former by adding an additional parameter.
- `openssl.Init` no longer supports OpenSSL version auto-selection, as it has too fork-specific. We have to implement the algorithm in our side instead.
- Remove the `!android` build tag.


patches/0003-Add-OpenSSL-crypto-backend.patch
```diff
@@ -225,16 +225,40 @@ index 00000000000000..68ea43ea892e1d
...
++			}
++			if exists && fallbackVersion == "" {
++				fallbackVersion = v
++			}
++		}
++		if version == "" && fallbackVersion != "" {
++			version = fallbackVersion
++		}
++	}
++	err := openssl.Init(version)
```
dagood: Looks like we're getting an extra dlopen+close vs. https://github.com/microsoft/go-crypto-openssl/blob/909778fe3d5df2c72e2f5853a126181985748ffe/openssl/openssl.go#L126 ?

Passing around an `unsafe.Pointer` seems iffy, but maybe openssl-fips could return something like
`type OpenSSLHandle struct { Version string, FIPS bool, handle unsafe.Pointer }` with `Close`.

I'm still not terribly sure dlopen/dlclose have significant overhead, this just jumps out to me because it's something I spotted recently on the go-crypto-openssl side. 😄 

Either way, this isn't something that needs to be addressed before merging this PR IMO.

patches/0003-Add-OpenSSL-crypto-backend.patch
```diff
@@ -250,12 +274,14 @@ index 00000000000000..68ea43ea892e1d
 +	case "0":
 +		if openssl.FIPS() {
 +			if err = openssl.SetFIPS(false); err != nil {
++				println(openssl.VersionText())
 +				panic("opensslcrypto: can't disable FIPS mode: " + err.Error())
```
dagood: Should the version be included with the panic rather than separate?

eng/doc/fips/README.md
```diff
@@ -29,7 +29,7 @@ Prior to Go 1.19, the boringcrypto changes were maintained in the `dev.boringcry
 
 The Microsoft Go fork modifies the Go runtime to implement several crypto primitives using cgo to call into a platform-provided cryptographic library rather than use the standard Go crypto implementations. This allows Go programs to use a platform-provided FIPS 140-2 certified crypto library.
 
-On Linux, the fork uses [OpenSSL](https://www.openssl.org/) through the [go-crypto-openssl] module. On Windows, [CNG](https://docs.microsoft.com/en-us/windows/win32/seccng/about-cng), using [go-crypto-winnative]. Similar to BoringSSL, certain OpenSSL and CNG versions are FIPS 140-2 certified.
+On Linux, the fork uses [OpenSSL](https://www.openssl.org/) through the [golang-fips/openssl-fips] module. On Windows, [CNG](https://docs.microsoft.com/en-us/windows/win32/seccng/about-cng), using [go-crypto-winnative]. Similar to BoringSSL, certain OpenSSL and CNG versions are FIPS 140-2 certified.
```
dagood: Technically we should probably still mention go-crypto-openssl because this doc is meant to describe all the currently supported releases of Go.

```suggestion
On Linux, the fork uses [OpenSSL](https://www.openssl.org/) through the [golang-fips/openssl-fips] module in Go 1.21+ and the [go-crypto-openssl] module in earlier versions. On Windows, [CNG](https://docs.microsoft.com/en-us/windows/win32/seccng/about-cng), using [go-crypto-winnative]. Similar to BoringSSL, certain OpenSSL and CNG versions are FIPS 140-2 certified.
```

eng/doc/fips/README.md
```diff
@@ -145,7 +145,7 @@ The `opensslcrypto` Go runtime supports multiple OpenSSL versions. It discovers
 Not all OpenSSL versions are supported. OpenSSL does not maintain ABI compatibility between different releases, even if only the patch version is increased, it needs specific attention to implement support. The relative importance of each version also results in a different amount of automated testing that has been implemented for various supported version. These are supported versions and the amount of automated validation for each one:
 
 - OpenSSL 1.1.1: the Microsoft CI builds official releases and runs the Go toolset test suite with this version.
-- OpenSSL 1.0.2, 1.1.0, 1.1.1, and 3.0.2: the [go-crypto-openssl] repository CI tests basic operation, but not the integration with the Go runtime.
+- OpenSSL 1.0.2, 1.1.0, 1.1.1, and 3.0.2: the [golang-fips/openssl-fips] repository CI tests basic operation, but not the integration with the Go runtime.
```
```suggestion
- OpenSSL 1.0.2, 1.1.0, 1.1.1, and 3.0.2: the [golang-fips/openssl-fips] and [go-crypto-openssl] repository CI tests basic operation, but not the integration with the Go runtime.
```

eng/doc/fips/README.md
```diff
@@ -189,6 +189,6 @@ The work done to support FIPS compatibility mode leverages code and ideas from o
 
 A program running in FIPS mode can claim it is using a FIPS-certified cryptographic module, but it can't claim the program as a whole is FIPS certified without passing the certification process, nor claim it is FIPS compliant without ensuring all crypto APIs and workflows are implemented in a FIPS-compliant manner.
 
-[go-crypto-openssl]: https://github.com/microsoft/go-crypto-openssl
+[golang-fips/openssl-fips]: https://github.com/golang-fips/openssl
```
```suggestion
[go-crypto-openssl]: https://github.com/microsoft/go-crypto-openssl
[golang-fips/openssl-fips]: https://github.com/golang-fips/openssl
```

patches/0003-Add-OpenSSL-crypto-backend.patch
qmuntal: Good catch.

patches/0003-Add-OpenSSL-crypto-backend.patch
qmuntal: I like the simplicity of `opensll.Init` accepting a version string rather that an `unsafe.Pointer`, it's less error prone and easier to understand, although I'm neither totally sold about the current approach.

On the other hand, I've validated that the init time difference between calling `openssl.CheckVersion` on all well-known versions + `openssl.Init` vs setting `GO_OPENSSL_VERSION_OVERRIDE` + `openssl.Init` is negligible, the former takes ~0.60ms and the later ~0.45ms.

dagood: I was thinking `handle unsafe.Pointer` wouldn't be exported, to make the API reasonably ok. (E.g. `Init` would accept the whole `OpenSSLHandle`, or `Init` becomes a method on `OpenSSLHandle`.) But yeah, avoiding it entirely is simpler and the perf doesn't sound critical.

dagood (approved):

xnox:
Is this going to be merged? It looks like it needs a rebase first. I also would want to update this to a newer snapshot of golang-fips/openssl to gain SHA3.

qmuntal:
> Is this going to be merged? It looks like it needs a rebase first. I also would want to update this to a newer snapshot of golang-fips/openssl to gain SHA3.

I'm waiting for this to happen: https://github.com/golang-fips/openssl/issues/83. I've stopped rebasing this PR till then, else I would have to rebase and fix conflicts constantly.

qmuntal:
Waiting for https://github.com/golang-fips/openssl/pull/92

chsienki (approved):

Issue #912, opened 2023-04-20
Use `-distpack` to create the tar.gz and zip archives of our builds


dagood:
When we set up the project, Go created the binary archives with a series of commands run through their buildlet APIs. Instead of trying to incorporate that in our AzDO builds, it was much easier to create a Go command that reproduced the end results: `eng/_core/archive/archive.go`.

As part of a reproducibility push, Go is putting the ability to create a Go binary archive into the Go repo itself with `-distpack`:

* https://github.com/golang/go/issues/24904
* https://go-review.googlesource.com/c/go/+/470676

We should use this so we don't have to maintain our own packing, which can break when upstream Go makes layout changes. (https://github.com/microsoft/go/issues/904)

dagood:
I tried upstream distpack with our patch files applied. I hit an issue where distpack requires a VERSION file, so I defined a fake `go1.21` one and ran it. We'll probably want to have our build wrapper place a "dev" VERSION file in main branch CI.

The first thing I noticed is that the distpack outputs don't have tar entries for each directory like ours do (and their old archives did). So, I did some cleanup to only show the file diffs.

On linux-amd64, the produced source archive has this diff vs. what we would produce:

```diff
-go/.gitattributes
-go/.github/CODE_OF_CONDUCT.md
-go/.github/ISSUE_TEMPLATE/00-bug.md
-go/.github/ISSUE_TEMPLATE/01-pkgsite.md
-go/.github/ISSUE_TEMPLATE/02-pkgsite-removal.md
-go/.github/ISSUE_TEMPLATE/03-gopls.md
-go/.github/ISSUE_TEMPLATE/04-vuln.md
-go/.github/ISSUE_TEMPLATE/10-proposal.md
-go/.github/ISSUE_TEMPLATE/11-language-change.md
-go/.github/ISSUE_TEMPLATE/config.yml
-go/.github/PULL_REQUEST_TEMPLATE
-go/.github/SUPPORT.md
-go/.gitignore
```

So: just a little more filtering than we do. The source diff looks good to me.

The binary archive has this diff (I ran our pack after upstream distpack, so some junk files are present in our archive):

```diff
--- a/ours.txt
+++ b/theirs.txt
@@ -65,6 +65,7 @@ go/api/next/59339.txt
 go/api/README
 go/bin/go
 go/bin/gofmt
+go/codereview.cfg
 go/CONTRIBUTING.md
 go/doc/asm.html
 go/doc/go1.17_spec.html
[...]

PR #946, opened 2023-06-08
default to GOTOOLCHAIN=local

qmuntal:
This PR does two things:
- Include `go.env` when creating toolchain archives so downstream users installing MSFT Go from those archives have sane default env variables set.
- Patch upstream's `go.env` so `GOTOOLCHAIN` defaults to `local` instead of `auto`. With this we make sure that the MSFT Go toolchain won't silently download and use non-MSFT Go toolchains when compiling applications that require a Go version other than the one being used. See https://github.com/microsoft/go/issues/912#issuecomment-1582255155 for more context.

### DIFF
--- a/patches/0007-Update-default-go.env.patch
+++ b/patches/0007-Update-default-go.env.patch
@@ -5,9 +5,7 @@ Subject: [PATCH] Update default go.env
 
 ---
  go.env                                           | 6 ++++--
- src/cmd/go/script_test.go                        | 1 +
- src/cmd/go/testdata/script/gotoolchain_local.txt | 8 +-------
- 3 files changed, 6 insertions(+), 9 deletions(-)
+ 1 files changed, 4 insertions(+), 2 deletions(-)
 
 diff --git a/go.env b/go.env
 index 6ff2b921d464bc..4aca2ff2c63dbc 100644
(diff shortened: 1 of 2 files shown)


patches/0007-Update-default-go.env.patch
```diff
@@ -0,0 +1,25 @@
...
+ GOPROXY=https://proxy.golang.org,direct
+ GOSUMDB=sum.golang.org
+ 
+-# Automatically download newer toolchains as directed by go.mod files.
+-# See https://go.dev/doc/toolchain for details.
+-GOTOOLCHAIN=auto
++# Use the locally installed Go toolchain, never downloading a different one.
++# Upstream uses `GOTOOLCHAIN=auto` instead, but `auto` can download and switch
++# to a non-MSFT Go toolchain, and we want to avoid that.
++# See https://go.dev/doc/toolchain for details.
```
qmuntal: Note for reviewers: go.dev/doc/toolchain is still not available, draft here https://go-review.googlesource.com/c/website/+/500775.

chsienki (approved):

dagood (approved):
Change LGTM, test still failing as of writing though:

```
=== Failed
=== FAIL: cmd/go TestScript/gotoolchain_local (0.03s)
=== FAIL: cmd/go TestScript (0.08s)
```

PR #935, opened 2023-05-31
Add systemcrypto experiment alias

dagood:
`GOEXPERIMENT=systemcrypto` enables the CNG or OpenSSL backend (or fails the build) depending on `GOOS`. On a Linux host, here are the results for a simple test app:

```
$ GOOS=linux GOEXPERIMENT=systemcrypto go build . ; go version exp
exp: devel go1.21-a37c2efaa4 Wed May 31 16:54:31 2023 -0500 X:opensslcrypto
$ GOOS=windows GOEXPERIMENT=systemcrypto go build . ; go version exp.exe
exp.exe: devel go1.21-a37c2efaa4 Wed May 31 16:54:31 2023 -0500 X:cngcrypto
$ GOOS=freebsd GOEXPERIMENT=systemcrypto go build .
go: the systemcrypto feature provided by the Microsoft build of Go does not support GOOS "freebsd"; for more information, see https://github.com/microsoft/go/tree/microsoft/main/eng/doc/fips
```

I noticed this `ParseGOEXPERIMENT` function is also where we could potentially enable this behavior by default in the future, so I checked that if we did that, `nosystemcrypto` would already work to disable it. (`GOOS=linux GOEXPERIMENT=opensslcrypto;nosystemcrypto` also disables the openssl backend. The experiment values are processed in order.)

The error message is a little lengthy, but I wanted to make it super clear that it's our repo that's doing this and where to find more info about it.

This should make it easier for everyone to integrate into their builds. For anything system-agnostic (e.g. building on the same machine it runs on--no knowledge of GOOS) this might be particularly useful.

PR #950, opened 2023-06-12
Rework systemcrypto to support tags; add better configuration errors

dagood:
Redo most of https://github.com/microsoft/go/pull/935 to support `-tags goexperiment.systemcrypto`.

Now, systemcrypto is a goexperiment proper and all it does is set the `goexperiment.systemcrypto` tag. This changes the meaning of `GOEXPERIMENT=cngcrypto,nosystemcrypto` (note the `no`). `nosystemcrypto` will not disable `cngcrypto` because they are technically unrelated in the code. `nosystemcrypto` would still disable `systemcrypto` if we were to enable `systemcrypto` by default, so I don't think this behavior change should matter.

The change also removed the obvious place to create configuration errors (only supported for `GOEXPERIMENT`) so I went with a different approach that isn't ideal, but it's additive rather than figuring out how to modify more `go build` code and maintaining that. The trick is: if a file has `func init() { "hello" }`, `go build` reports `"hello"` as unused, and includes the string in the message. So, by putting useful info in the string, we can show it to our users if some `//go:build` condition is met. (@qmuntal showed me this in a sync recently. 😄)

I think it might be worth going back and changing the `go build` code to do these checks somewhere rather than relying on this particular quirk of error reporting. (Error reporting might change and start to chop or escape our text in future Go versions?)

Note `src/runtime/backenderr_systemcrypto_nobackend.go` is in `runtime`. The `src/runtime/backenderrs.go` file has a comment explaining why these files are in `runtime` rather than `crypto/internal/backend`. It feels odd putting the files in `runtime` but it seems like the right thing to do to make them properly useful.

<details><summary>Edit: see comments for the latest error messages. Click here to see the error messages before I removed backenderr_unknown.go to fix tests.</summary>

The customized error messages look like this:

```
$ GOOS=linux GOEXPERIMENT=systemcrypto,boringcrypto go build .
# runtime
../go/src/runtime/backendconflict_boring_openssl.go:12:2: `
        The boring and openssl backends are both enabled, but they are mutually exclusive.
        Please make sure only one crypto backend experiment is enabled by GOEXPERIMENT or '-tags'.
        ` (untyped string constant "\n\tThe boring and openssl backends are both enabled, but they are m...) is not used

$ GOOS=linux GOEXPERIMENT=cngcrypto go build .
# runtime
../go/src/runtime/backenderr_unknown.go:12:2: `
        The current combination of build tags removes the use of built-in Go cryptography but doesn't enable a crypto backend.
        Make sure the crypto backend selected by GOEXPERIMENT or '-tags' is supported on the target OS (GOOS).
        ` (untyped string constant "\n\tThe current combination of build tags removes the use of built-i...) is not used

$ GOOS=freebsd GOEXPERIMENT=systemcrypto go build .
# runtime
../go/src/runtime/backenderr_systemcrypto_nobackend.go:12:2: `
        The systemcrypto feature is enabled, but it was unable to enable an appropriate crypto backend for the target GOOS.
        ` (untyped string constant "\n\tThe systemcrypto feature is enabled, but it was unable to enable...) is not used
```

It's not ideal, but better than previous:

```
$ GOOS=linux GOEXPERIMENT=systemcrypto,boringcrypto go build .
# crypto/internal/backend
../go/src/crypto/internal/backend/openssl_linux.go:23:7: Enabled redeclared in this block
        ../go/src/crypto/internal/backend/boring_linux.go:19:7: other declaration of Enabled
[... many of the same type of error ...]
../go/src/crypto/internal/backend/openssl_linux.go:99:6: SHA256 redeclared in this block
        ../go/src/crypto/internal/backend/boring_linux.go:31:6: other declaration of SHA256
../go/src/crypto/internal/backend/openssl_linux.go:99:6: too many errors
```

I tried out a lot more experiment/tag combinations and made a table of the success or error that comes back: https://gist.github.com/dagood/71ba8d93f2c1b20180206df30bda6cbf#file-table-md

</details>

---

This change is done in two patches: the first implements systemcrypto, the second adds the error messages.

To make the tags work, I had to modify a file with contents that have been copied to two places. I added a test to make sure that the two copies remain the same. This test runs in CI, so it should let us know if upstream makes changes that affect these files, too.


qmuntal (commented):
This looks great, specially the `backenderrgenerator` approach, it will safe us from messing the build tags (even more).

patches/0007-Add-informative-backend-configuration-errors.patch
```diff
@@ -0,0 +1,302 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
```
qmuntal, line 1: It would be good to have a CI step that run the generator and fails if new diffs appear, this way they won't get out of sync.

qmuntal (changes requested):

patches/0006-Add-goexperiment.systemcrypto-alias.patch
```diff
@@ -0,0 +1,271 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
```
qmuntal: This patch should go before the `Add-crypto-backend-foundation.patch`, having `goexperiment.systemcrypto` will simplify patching build tag headers down the line.

Both two could even be merged, if it makes sense to you.

qmuntal (changes requested):

patches/0007-Add-informative-backend-configuration-errors.patch
dagood: True, but I'm actually thinking it might be better to go with the `go test [...] -update` approach and keep it inside the Go submodule for simplicity (depending on how you look at it).

patches/0006-Add-goexperiment.systemcrypto-alias.patch
dagood: 👍 

I think it's focused enough on `go build` that it makes sense to keep separate. It seems pretty independent from the foundation patch, too. (I don't see a reason the foundation patch would use the systemcrypto tag.)

patches/0001-Add-systemcrypto-GOEXPERIMENT.patch
```diff
@@ -0,0 +1,282 @@
...
+Date: Wed, 31 May 2023 16:54:31 -0500
+Subject: [PATCH] Add systemcrypto GOEXPERIMENT
+
+Adds systemcrypto GOEXPERIMENT and modifies tag handling to make
+goexperiment.systemcrypto behave as an alias that enables the recommended
+backend for the target GOOS.
+
+The reverse is also true: goexperiment.systemcrypto can be used as an alias in a
+build constraint. It's satisfied when (and only when) the recommended backend
+for the target GOOS is enabled, such as goexperiment.opensslcrypto on Linux.
```
dagood: @qmuntal and I talked about `when (and only when)` a bit offline--I leaned towards symmetry when I was implementing this and lost track of how that makes the tag less useful for constraints:

* "Any backend", regardless of what Microsoft Go recommends, is useful. In our own patches, this would simplify
  `//go:build boringcrypto || goexperiment.opensslcrypto || goexperiment.cngcrypto`. Other library/app maintainers may need conditions just like that.
* Could cause breaks: if we change our Windows backend recommendation from (say) CNG to BetterCNG in Go 1.32, someone using a `//go:build goexperiment.systemcrypto` constraint and `GOEXPERIMENT=cngcrypto` for compatibility/long-term-support reasons would be broken in Go 1.32 because cngcrypto would no longer enable goexperiment.systemcrypto, only bettercngcrypto would. This seems like a needless break.

A quirk here is that `boringcrypto` is actually statically linked rather than being system-provided... but I'm not sure it's worth splitting hairs, just mention in the experiment doc.

We can always create an additional synthetic tag to distinguish "any backend" from "Microsoft-Go-picked automatic backend", but I want to give it a shot with just the one tag name first, and see if it ends up being fine.

That said, this is a tweak, so we don't think it should block this PR.

---

Earlier, I had some concern about what "any backend" might imply:

	// It may seem like it's enough to satisfy "goexperiment.systemcrypto" when
	// "goexperiment.opensslcrypto" is specified, and omit the GOOS=Linux check.
	// However, systemcrypto is a recommendation by the Microsoft Go toolset,
	// and we want to be careful not to misinform the user.

The idea is if you set `GOOS=windows GOEXPERIMENT=opensslcrypto` and your Go file has `//build goexperiment.systemcrypto`, the fact that it's satsified *could* be construed to mean that we recommend it, because passing `goexperiment.systemcrypto` enables our recommendation. But... if we simply document the behavior, I don't see this being an issue.

dagood: Done. After writing it out and changing the code, I don't see any issue with this.

qmuntal (approved):

patches/0007-Add-informative-backend-configuration-errors.patch
```diff
@@ -0,0 +1,322 @@
...
++// would show up first, causing these informative errors to be skipped because
++// there are too many total errors already reported.
++//
++// The errors would also show up if we put these files in the crypto package
++// rather than the runtime package. Crypto is imported before the backend
++// backage, so the errors would show up. However, then these errors would show
++// up only if the Go program is using crypto. This could cause a confusing
++// situation: if the user has a misconfigured backend and doesn't use crypto in
++// their Go app, they will not get any errors. If they start using crypto later,
++// they would only then get an error, but the cause would be much less apparent.
```
dagood, line 243: This is what the tests are (basically) running up against. The tests were running fine before this change because they don't actually perform crypto. This meant `crypto/internal/backend` wasn't being imported. But with `backenderr_unknown.go` being plopped in `runtime`, it's always imported and causes a few tests that perform unusual builds to fail.

### Boring Linux job failure

`cmd/internal/testdir Test/codegen/copy.go` ends up running something like:

```
GOEXPERIMENT=boringcrypto GOOS=linux GOARCH=386 CGO_ENABLED=1 ../go/bin/go build .
```

The boring backend has the `amd64` constraint, and `nobackend.go` doesn't have a condition for `!(boringcrypto && !amd64)`, leaving this case unhandled, which triggers `backenderr_unknown.go` and a build/test failure.

### CNG Windows job failure

`cmd/internal/obj/x86 TestVexEvexPCrelative` tries to run something like this:

```
GOEXPERIMENT=cngcrypto GOOS=linux GOARCH=amd64 ../go/bin/go build .
```

CNG doesn't work on Linux, of course, but `nobackend.go` doesn't have a condition that lets it fill in.

### OpenSSL Linux job (pass)

There wasn't a problem with opensslcrypto because its build constraints are more permissive (allows `386`). Also, if a test needs to do something and doesn't care about the platform (like `TestVexEvexPCrelative`), `linux/amd64` is typically the default choice.

---

It was intentional to make the build fail more proactively than it used to (described in this comment), but I think I need to remove this particular error detector to let the tests succeed.

This did make me realize that maybe I could take the code that generates the `backenderr_unknown.go` constraint and instead have it test/write the constraint for `nobackend.go`. I'll file an issue to think over whether that's a good idea. 😄 (edit: https://github.com/microsoft/go/issues/953)

dagood: The end result after fixing CI:

* `GOOS=linux GOEXPERIMENT=systemcrypto,boringcrypto go build .` 
  "The boring and openssl backends are both enabled, but they are mutually exclusive. Please make sure only one crypto backend experiment is enabled by GOEXPERIMENT or '-tags'."
* `GOOS=freebsd GOEXPERIMENT=systemcrypto go build .`
  "The systemcrypto feature is enabled, but it was unable to enable an appropriate crypto backend for the target GOOS."
* `GOOS=linux GOEXPERIMENT=cngcrypto go build .` 
  ```
  # crypto/internal/backend
  ../go/src/crypto/internal/backend/common.go:41:5: undefined: Enabled
  ../go/src/crypto/internal/backend/common.go:64:5: undefined: Enabled
  ```

Opened https://github.com/microsoft/go/issues/953 to go into the last case further.

qmuntal (approved):
Still LGTM!

chsienki (approved):

Issue #945, opened 2023-06-07
Publish assets.json file to blob storage with aka.ms URL for latest patch version lookup


dagood:
* https://github.com/microsoft/go/issues/939 would let a download tool know what minor versions are currently supported (1.19+1.20). This leaves out the patch version, however.
* The straightforward way to find the latest patch version is to download a Go build and inspect its content, but this is slow.
* Another way to find the latest patch version is to grab all the tags on the repo, filter them, and sort them, but this is complicated.

We can make this straightforward and fast by reusing the assets.json file we publish in each GitHub release: publish it on blob storage and update an aka.ms link to point at it, like the other artifacts.

PR #958, opened 2023-06-21
Add JSON file with same data as download table

### LINKED ISSUE #939: Create a JSON file containing the same information as the download table
https://github.com/microsoft/go/blob/microsoft/main/eng/doc/Downloads.md is good for humans to see that we support 1.20 and 1.19 (as of writing), but it would be good to have a place for robots so they can figure out where to download something like `stable` and `oldstable`.

I think that it would be easy to make `updatelinktable` create a JSON file as well as update the table. It would add no (expected) maintenance cost and have good value for anyone creating a tool to download/install the latest tar.gz/zip.

I still think that it shouldn't include the patch versions--that would increase maintenance. We already have aka.ms links that point at the latest e.g. 1.20.x patch version for 1.20. If we make a database of patch versions (other than the repository tags) I think it needs to be automated.

dagood:
Modify the `updatelinktable` tool that generates the downloads table to also generate a JSON file with basically the same data. This file can be parsed to easily figure out what major versions are supported/released and download them.

The format is essentially https://go.dev/dl/?mode=json, but adjusted to make sense when it only provides links to the latest patch of each Go version.

* Fixes https://github.com/microsoft/go/issues/939

We should consider publishing this file to our own blob storage with an aka.ms link so Go installation tools don't end up relying on e.g. https://raw.githubusercontent.com/microsoft/go/da62c0bec9417596b1ded94c01ca612e502b374a/eng/doc/release-branch-links.json in the long term. This could go along with publishing the assets.json files: https://github.com/microsoft/go/issues/945.

### DIFF
--- a/eng/doc/release-branch-links.json
+++ b/eng/doc/release-branch-links.json
@@ -1,0 +1,112 @@
+[
+  {
+    "version": "go1.20",
+    "stable": true,
+    "files": [
+      {
+        "filename": "go1.20.src.tar.gz",
+        "os": "",
+        "arch": "",
+        "version": "go1.20",
+        "kind": "source",
+        "url": "https://aka.ms/golang/release/latest/go1.20.src.tar.gz",
+        "checksumURL": "https://aka.ms/golang/release/latest/go1.20.src.tar.gz.sha256",
+        "signatureURL": "https://aka.ms/golang/release/latest/go1.20.src.tar.gz.sig"
+      },
+      {
+        "filename": "go1.20.linux-amd64.tar.gz",
+        "os": "linux",
+        "arch": "amd64",
+        "version": "go1.20",
+        "kind": "archive",
+        "url": "https://aka.ms/golang/release/latest/go1.20.linux-amd64.tar.gz",
+        "checksumURL": "https://aka.ms/golang/release/latest/go1.20.linux-amd64.tar.gz.sha256",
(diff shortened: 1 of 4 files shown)


eng/doc/release-branch-links.json
```diff
@@ -0,0 +1,112 @@
...
+        "version": "go1.20",
+        "kind": "archive",
+        "url": "https://aka.ms/golang/release/latest/go1.20.windows-amd64.zip",
+        "checksumURL": "https://aka.ms/golang/release/latest/go1.20.windows-amd64.zip.sha256"
+      }
+    ]
+  },
+  {
+    "version": "go1.19",
+    "oldStable": true,
```
dagood: Hmm, I added `stable` and `oldStable` with https://github.com/actions/setup-go#using-stableoldstable-aliases in mind, but https://go.dev/dl/?mode=json lists both 1.19.x and 1.20.x as `"stable": true`, so this conflicts a bit.

I think I should just make `stable` always `true`. Then, if we ever put an RC release or something similar in this list, we can make it `false`.

I still think someone might find something like `"latestStable": true` and `"previousStable": true` useful, to avoid having to sort this list, or assume the first in the list is always the latest, or assume that if the latest is 1.31, the previous number is necessarily 1.30. (Maybe in some weird future case we skip 1.30 and still support 1.29+1.31.) Nobody is currently asking for this though.

qmuntal (approved):

jaredpar (approved):

Issue #953, opened 2023-06-17
Make `nobackend.go` build constraint an inverse of the other backends (and add test)


dagood:
In https://github.com/microsoft/go/pull/950 I wrote a test/generator that parses the build constraints for each backend and constructs a simple inverse. We could use that code to generate the constraint for `nobackend.go` so that it is *always* enabled when none of the backends' conditions are met.

This could allow "silent" unintended usage of Go crypto. `GOEXPERIMENT=cngcrypto GOOS=linux go build` would not end up using a backend. However, this is in line with some other behavior: `GOEXPERIMENT=opensslcrypto GOOS=linux CGO_ENABLED=0 go build` behaves similarly today.

It seems to me we should still make sure `GOEXPERIMENT=systemcrypto` never silently uses Go crypto, though. Based on the issues I saw in https://github.com/microsoft/go/pull/950, this might mean `GOEXPERIMENT=systemcrypto go tool dist test` won't work (unless we make more test changes), but since systemcrypto is a fairly simple alias, this might be ok.

dagood:
@qmuntal and I discussed this in Go sync:

* We think that generating/testing the `nobackend.go` build constraint makes sense no matter what, simply to make sure the condition is always correct. There's no intentional mismatch.
* In the past we've favored compatibility--seamlessly using Go crypto when the conditions aren't right to use system crypto. With the new crypto policy that we're trying to help people conform to, we think we should switch to more proactive errors, instead.

To fix the test suite, a few ideas:

* Add an "opt out of proactive errors" tag. We could set it in CI (but then `go dist test` "in the wild" may not work) or set it in `go dist test` as part of the patch.
* Individually fix test issues. For the errors I spotted in https://github.com/microsoft/go/pull/950#discussion_r1232893138, it might be enough to detect and remove crypto `GOEXPERIMENT`s when running those `go build` commands.
* Now that I've thought about it: both. I think I like this the most.
  * Add an opt out, and enable it *only* in tests that have issues. (Patch each test individually. Only 2 are known.)
  * It's easier to add an opt-out tag than remove an experiment, adding it in individual tests means most tests are testing the proactive errors, and we might find some inherent value in knowing what scenarios interact poorly with the proactive errors. 
  * In theory, if someone using our toolset needs the opt out for some reason (overwhelmingly simplifies their build process?), it will also be easy for them to enable. I don't think we should advertise that it exists, though, because it's certainly not the right thing to do in any normal case.

PR #960, opened 2023-06-23
Generate nobackend build constraint

dagood:
* For https://github.com/microsoft/go/issues/953

Generate the build constraint in `nobackend.go`. I moved the existing generator into crypto/internal/backend so I could write a single generator with shared code (and only one `go generate` call) rather than two separate ones. This means the generator is "remotely" writing files into `runtime/` from `crypto/internal/backend/`, but I updated the "do not edit" comment so it's clear where it's coming from.

Also made the error file generator a little smarter: for easier maintenance, it detects and deletes old files in case we remove or rename a backend or the generated filename format.

---

This PR just implements the `nobackend` constraint generation part of https://github.com/microsoft/go/issues/953#issuecomment-1599244099 and doesn't yet implement the "unexpected Go crypto usage" guardrail error. I figured it's worth posting this PR first in case that becomes a rabbit hole. (The test failures I hit in https://github.com/microsoft/go/pull/950 might just be a first wave.)


patches/0007-Add-backend-code-gen.patch
```diff
@@ -0,0 +1,427 @@
...
+
+The generators also run during standard test runs in validation mode to
+ensure the generated content is not out of date when backends are
+added/removed or when their build constraints change.
+
+To simplify patch maintenance, the generators are only run once, in this
+patch. An alternative would be adding the generators in an earlier patch
+and re-running the generators immediately for each backend patch. This
+would make the direct connection between changes clearer, but make it
+harder to update the generators and to deal with conflicts.
```
dagood, line 23: I did write the generator so it could work if run per-patch (that's what `go1.1` is about), and actually tried it out. I think it's significantly easier to maintain in its own patch, though. Needing to rebase, run a command on each backend patch, amend, and continue is not great with our current tools, IMO.

If we added a "patch fixup command" patch file feature that would run in a "patch fixup rebase" `git-go-patch` command, that would potentially make it viable. I've always thought this kind of feature would be useful for re-vendoring, too, but it just isn't quite useful enough to be worthwhile vs. simply having the command in the patch message. (Or remembering it--vendoring is pretty straightforward.)

(If you're in the middle of a backend patch, you also need to run with `go generate -mod=readonly ...` because the vendoring dir is stale at that point in time.)

qmuntal (approved):
LGTM!

chsienki (approved):

PR #964, opened 2023-06-27
Call fipstls.Force() when running in FIPS mode

qmuntal:
Given the new crypto policy, downstream users who need to run in FIPS mode would always want to restrict the TLS stack settings to be FIPS-compliant. It's easier for everyone, less downstream changes  if we just enforce it at runtime when we detect that the underlying crypto is running in FIPS mode.

### DIFF
--- a/eng/doc/fips/README.md
+++ b/eng/doc/fips/README.md
@@ -167,13 +167,7 @@ This feature does not require any additional configuration, but it only works wi
 
 ### TLS with FIPS-approved settings
 
-The Go TLS stack will automatically use OpenSSL crypto primitives when running in FIPS mode. Yet, the FIPS 140-2 standard places additional restrictions on TLS communications, mainly on which cyphers and signers are allowed.
-
-A program can import the `crypto/tls/fipsonly` package to configure the Go TLS stack so it is compliant with these restrictions. The configuration is done by an `init()` function, so only importing it is necessary:
-
-```go
-  import _ "crypto/tls/fipsonly"
-```
+The Microsoft Go runtime will automatically enforce that `crypto/tls` and `crypto/x509` only use FIPS-approved settings when running in FIPS mode. These differs from upstream's BoringCrypto backend, in which to get these additional restrictions one has to import `crypto/tls/fipsonly`.
 
 Note that this can reduce compatibility with old devices that do not support modern cryptography techniques such as TLS 1.2.
 
--- a/eng/doc/fips/README.md
+++ b/eng/doc/fips/README.md
@@ -167,7 +167,15 @@ This feature does not require any additional configuration, but it only works wi
 
 ### TLS with FIPS-approved settings
 
-The Microsoft Go runtime will automatically enforce that `crypto/tls` and `crypto/x509` only use FIPS-approved settings when running in FIPS mode. These differs from upstream's BoringCrypto backend, in which to get these additional restrictions one has to import `crypto/tls/fipsonly`.
+The Go TLS stack will automatically use OpenSSL crypto primitives. Yet, the FIPS 140-2 standard places additional restrictions on TLS communications, mainly on which cyphers and signers are allowed.
+
+Since Go 1.22, the Microsoft Go runtime will automatically enforce that `crypto/tls` and `crypto/x509` only use FIPS-approved settings when running in FIPS mode. These differs from upstream's BoringCrypto backend, in which to get these additional restrictions one has to import `crypto/tls/fipsonly`.
+
+Prior to Go 1.22, a program should import the `crypto/tls/fipsonly` package to configure the Go TLS stack so it is compliant with these restrictions. The configuration is done by an `init()` function, so only importing it is necessary:
+
+```go
+  import _ "crypto/tls/fipsonly"
+```
 
 Note that this can reduce compatibility with old devices that do not support modern cryptography techniques such as TLS 1.2.
 
--- a/eng/doc/fips/README.md
+++ b/eng/doc/fips/README.md
@@ -169,9 +169,9 @@ This feature does not require any additional configuration, but it only works wi
 
 The Go TLS stack will automatically use OpenSSL crypto primitives. Yet, the FIPS 140-2 standard places additional restrictions on TLS communications, mainly on which cyphers and signers are allowed.
 
-Since Go 1.22, the Microsoft Go runtime will automatically enforce that `crypto/tls` and `crypto/x509` only use FIPS-approved settings when running in FIPS mode. These differs from upstream's BoringCrypto backend, in which to get these additional restrictions one has to import `crypto/tls/fipsonly`.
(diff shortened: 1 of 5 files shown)


eng/doc/fips/README.md
```diff
@@ -167,13 +167,7 @@ This feature does not require any additional configuration, but it only works wi
...
 ### TLS with FIPS-approved settings
 
-The Go TLS stack will automatically use OpenSSL crypto primitives when running in FIPS mode. Yet, the FIPS 140-2 standard places additional restrictions on TLS communications, mainly on which cyphers and signers are allowed.
-
-A program can import the `crypto/tls/fipsonly` package to configure the Go TLS stack so it is compliant with these restrictions. The configuration is done by an `init()` function, so only importing it is necessary:
-
-```go
-  import _ "crypto/tls/fipsonly"
-```
+The Microsoft Go runtime will automatically enforce that `crypto/tls` and `crypto/x509` only use FIPS-approved settings when running in FIPS mode. These differs from upstream's BoringCrypto backend, in which to get these additional restrictions one has to import `crypto/tls/fipsonly`.
```
dagood: I think for now we need to leave the deleted text because this is the live doc for all supported versions.

Once it's backported, I'm wondering now if this kind of change actually might deserve a "since Go 1.19.11, 1.20.6, and 1.21" note in this doc until 1.22 is released.

eng/doc/fips/UserGuide.md
```diff
@@ -1231,8 +1231,6 @@ Does not contain crypto primitives, out of FIPS scope.
 
 Package tls partially implements TLS 1.2, as specified in RFC 5246, and TLS 1.3, as specified in RFC 8446.
 
-Package tls will automatically use FIPS compliant primitives implemented in other crypto packages, but it will accept non-FIPS ciphers and signature algorithms unless `crypto/tls/fipsonly` is imported.
```
dagood, line 1234: (Same here.)

qmuntal:
I've submitted an [upstream CL](https://go-review.googlesource.com/c/go/+/506795) that implements all test fixes from this PR. If/when it is submitted we will be able to drop all the new patch changes.

eng/doc/fips/README.md
```diff
@@ -167,9 +167,11 @@ This feature does not require any additional configuration, but it only works wi
 
 ### TLS with FIPS-approved settings
 
-The Go TLS stack will automatically use OpenSSL crypto primitives when running in FIPS mode. Yet, the FIPS 140-2 standard places additional restrictions on TLS communications, mainly on which cyphers and signers are allowed.
+The Go TLS stack will automatically use OpenSSL crypto primitives. Yet, the FIPS 140-2 standard places additional restrictions on TLS communications, mainly on which cyphers and signers are allowed.
 
-A program can import the `crypto/tls/fipsonly` package to configure the Go TLS stack so it is compliant with these restrictions. The configuration is done by an `init()` function, so only importing it is necessary:
+Since Go 1.22, the Microsoft Go runtime will automatically enforce that `crypto/tls` and `crypto/x509` only use FIPS-approved settings when running in FIPS mode. These differs from upstream's BoringCrypto backend, in which to get these additional restrictions one has to import `crypto/tls/fipsonly`.
```
dagood: A few small things (preferring present tense, using "you" since it's present elsewhere in the doc and "one" hasn't shown up yet, a little rearranging, writing down the reason for the diff):

```suggestion
Since Go 1.22, the Microsoft Go runtime automatically enforces that `crypto/tls` and `crypto/x509` only use FIPS-approved settings when running in FIPS mode. This differs from upstream's BoringCrypto backend, which requires you to import `crypto/tls/fipsonly` to apply the FIPS-mandated restrictions. The Microsoft Go crypto backends do this automatically to reduce the source code changes necessary to produce a FIPS-compliant Go application.
```

eng/doc/fips/README.md
```diff
@@ -167,9 +167,11 @@ This feature does not require any additional configuration, but it only works wi
...
+
+Prior to Go 1.22, a program should import the `crypto/tls/fipsonly` package to configure the Go TLS stack so it is compliant with these restrictions. The configuration is done by an `init()` function, so only importing it is necessary:
```
dagood: I think `must` is nicely direct here if we add the condition (using the TLS stack):

```suggestion
Prior to Go 1.22, a program using the Go TLS stack must import the `crypto/tls/fipsonly` package to be compliant with these restrictions. The configuration is done by an `init()` function, so only importing it is necessary:
```

patches/0002-Add-crypto-backend-foundation.patch
```diff
@@ -702,6 +956,118 @@ index be8ac30f9daf9f..3356154c7d1082 100644
...
+diff --git a/src/net/http/client_test.go b/src/net/http/client_test.go
+index 0fe555af38f08a..5b2c8051a8ba83 100644
+--- a/src/net/http/client_test.go
++++ b/src/net/http/client_test.go
+@@ -960,7 +960,7 @@ func testResponseSetsTLSConnectionState(t *testing.T, mode testMode) {
+ 
+ 	c := ts.Client()
+ 	tr := c.Transport.(*Transport)
+-	tr.TLSClientConfig.CipherSuites = []uint16{tls.TLS_RSA_WITH_3DES_EDE_CBC_SHA}
++	tr.TLSClientConfig.CipherSuites = []uint16{tls.TLS_RSA_WITH_AES_256_GCM_SHA384}
```
dagood, line 752: To make sure I understand the change: it isn't all that important to this test which cipher is chosen, so this change is simply making it use a cipher compatible with FIPS TLS because there's no reason not to, and it's simpler than Abandon/Force? (Especially since these changes are making their way upstream.)

dagood (approved):

patches/0002-Add-crypto-backend-foundation.patch
qmuntal: Yep, that's it.

qmuntal:
Well, the number of tests to fix when running in FIPS mode is too high to patch them one by one. I've reverted most fixes and instead called `fipstls.Abadon` when executing `crypto/tls` and `crypto/x509` tests. On the other hand, we can't call that function in `net/http` and `net/smpt`, as it is internal of `crypto`, so I've went ahead and fix the two failing tests in those packages.

eng/doc/fips/UserGuide.md
```diff
@@ -1231,7 +1231,10 @@ Does not contain crypto primitives, out of FIPS scope.
 
 Package tls partially implements TLS 1.2, as specified in RFC 5246, and TLS 1.3, as specified in RFC 8446.
 
-Package tls will automatically use FIPS compliant primitives implemented in other crypto packages, but it will accept non-FIPS ciphers and signature algorithms unless `crypto/tls/fipsonly` is imported.
+Package tls will automatically use FIPS compliant primitives implemented in other crypto packages. 
+
+Since Go 1.22, the Microsoft Go runtime will automatically enforce that tls only use FIPS-approved settings when running in FIPS mode.
+Prior to Go 1.22, a program should import the `crypto/tls/fipsonly` package to configure tls so it is compliant with these restrictions.
```
dagood: Nitpicking--could be more similar to the main doc: 😄 

```suggestion
Since Go 1.22, the Microsoft Go runtime automatically enforces that tls only uses FIPS-approved settings when running in FIPS mode.
Prior to Go 1.22, a program using tls must import the `crypto/tls/fipsonly` package to be compliant with these restrictions.
```

dagood (approved):

qmuntal:
Closing this PR as won't do for now. We will revisit it when we have more usage insights.

PR #965, opened 2023-06-27
Prevent unexpected Go crypto fallback: fail instead

### LINKED ISSUE #953: Make `nobackend.go` build constraint an inverse of the other backends (and add test)

dagood:
* Resolves https://github.com/microsoft/go/issues/953

Once this is merged, something like `GOEXPERIMENT=opensslcrypto GOOS=windows go build .` will fail with:

```
# runtime
../go/go/src/runtime/backenderr_gen_nofallback_openssl.go:12:2: `
        The goexperiment.opensslcrypto tag is specified, but other tags required to enable that backend were not met.
        Required build tags:
          goexperiment.opensslcrypto && linux && cgo && !android && !cmd_go_bootstrap && !msan
        Please check your build environment and build command for a reason one or more of these tags weren't specified.
        ` (untyped string constant "\n\tThe goexperiment.opensslcrypto tag is specified, but other tags ...) is not used
```

Adding `-tags=allow_missing_crypto_backend_fallback` removes the error, allowing fallback. This is only intended for specific cases in our CI job where we e.g. enable `GOEXPERIMENT=opensslcrypto` globally and then run the whole test suite. A number of tests have inner `go build` commands that either disable cgo or use a different GOOS and needed the tag.

If someone needs to use Go crypto in their own app, it makes more sense to remove the `GOEXPERIMENT` value. Then, it's clear what the intent is by looking at the build command, and our toolset then makes sure the intent is followed.

This PR also changes the cngcrypto/opensslcrypto/boringcrypto CI jobs to build normally and run the tests with the target GOEXPERIMENT. The toolchain (at various levels of bootstrapping) builds without cgo, so the "accidental fallback" error would show up. Since the toolchain is using fallback Go crypto, we aren't meaningfully testing the goexperiment in these cases, so it makes more sense disable the goexperiment for the initial build rather than feed `allow_missing_crypto_backend_fallback` into many places.

(The build of Go we ship is built with no GOEXPERIMENT, so it isn't necessarily important to test that building Go with a crypto experiment works, only that tests built with our Go toolset and the crypto experiments work.)

### DIFF
--- a/eng/_core/cmd/build/build.go
+++ b/eng/_core/cmd/build/build.go
@@ -133,10 +133,6 @@ func build(o *options) error {
 		return err
 	}
 
-	if o.Experiment != "" {
-		buildutil.AppendExperimentEnv(o.Experiment)
-	}
-
 	if !o.SkipBuild {
 		// If we have a stage 0 copy of Go in an env variable (as set by run.ps1), use it in the
 		// build command by setting GOROOT_BOOTSTRAP. The upstream build script "make.bash" uses
@@ -228,6 +224,16 @@ func build(o *options) error {
 			// The stderr output isn't used to determine whether the tests succeeded or not. (The
 			// redirect doesn't cause an issue where tests succeed that should have failed.)
 			testCmd.Stderr = os.Stdout
+
+			if o.Experiment != "" {
+				buildutil.AppendExperimentEnv(o.Experiment)
+				// We aren't able to rebuild Go standard library packages under a crypto experiment,
+				// but "cmd/dist/test.go" will normally do this for local test runs to make sure the
+				// build is clean. The problem is that the build doesn't include cgo, which is
+				// required for some crypto backends. Set this variable to promise that our build is
+				// fresh and avoid the non-cgo build.
+				os.Setenv("GO_BUILD_FRESH", "1")
+			}
 
 			return runCmd(testCmd)
 		}
--- a/eng/_core/cmd/build/build.go
+++ b/eng/_core/cmd/build/build.go
@@ -230,9 +230,10 @@ func build(o *options) error {
 				// We aren't able to rebuild Go standard library packages under a crypto experiment,
 				// but "cmd/dist/test.go" will normally do this for local test runs to make sure the
 				// build is clean. The problem is that the build doesn't include cgo, which is
-				// required for some crypto backends. Set this variable to promise that our build is
-				// fresh and avoid the non-cgo build.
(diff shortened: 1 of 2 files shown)


eng/_core/cmd/build/build.go
```diff
@@ -229,6 +225,16 @@ func build(o *options) error {
...
 			testCmd.Stderr = os.Stdout
 
+			if o.Experiment != "" {
+				buildutil.AppendExperimentEnv(o.Experiment)
+				// We aren't able to rebuild Go standard library packages under a crypto experiment,
+				// but "cmd/dist/test.go" will normally do this for local test runs to make sure the
+				// build is clean. The problem is that the build doesn't include cgo, which is
+				// required for some crypto backends. Set this variable to promise that our build is
+				// fresh and avoid the non-cgo build.
+				os.Setenv("GO_BUILD_FRESH", "1")
```
qmuntal: `GO_BUILD_FRESH` seems like an upstream env var and it keeps me wondering if we are reusing some already existing functionality. I'll prefer to name is something like `GO_MSFT_BUILD_FRESH` so the scope is clearer.

qmuntal (approved):

eng/_core/cmd/build/build.go
dagood: Ended up as `GO_MSFT_SCRIPTED_BUILD`--realized that "BUILD_FRESH" sounds more like a request to run a fresh build, not something that indicates current status.

qmuntal (approved):

chsienki (approved):

PR #968, opened 2023-06-30
Add requirefips tag

### LINKED ISSUE #928: Add compile-time flag that enables FIPS mode and ignores GOFIPS=0 at runtime
A partner team has the responsibility to deliver FIPS-mode-only images containing a Go app to its customers. This could normally be done by setting GOFIPS=1 in the image's environment, but this image is meant to be used in a customer-defined AKS cluster, so the whole thing may be open to mistakes:

* The user could set `GOFIPS=0` to make some non-FIPS crypto algorithm work, perhaps not even aware they are working on a product that's meant to be FIPS-compliant.
* The user could clear out environment variables (perhaps attempting to have a "clean slate") causing `GOFIPS` to no longer be specified at all.

In these cases, instead of respecting the runtime `GOFIPS`, it would be better to fail. The images will have FIPS-specific tags, so the user is definitely interested in FIPS, and this is an opportunity to let the user know they made a configuration mistake.

We could add to goexperiment, so the app would be built with e.g. `GOEXPERIMENT=opensslcrypto,compelfips`.

---

This means the user can't toggle off FIPS mode to narrow down a runtime error. However, this technique is unlikely to be all that useful: they would still be using OpenSSL/CNG under the covers. It's also unlikely someone will try this: the images are more like appliances, so they will likely try to just swap out the tag to a non-FIPS tag--which would work fine.

dagood:
* Resolves https://github.com/microsoft/go/issues/928
* Doc PR: https://github.com/microsoft/go/pull/970

This change makes `GOEXPERIMENT=systemcrypto go build -tags=requirefips . && ./exp` use `GOFIPS=1` behavior, and panics if `GOFIPS=0` or `GOLANG_FIPS=0`:

```
$ GOEXPERIMENT=systemcrypto go build -tags=requirefips . && GOFIPS=0 ./exp
panic: the 'requirefips' build tag is enabled, but it conflicts with the detected env variable GOFIPS=0 which would disable FIPS mode

goroutine 1 [running]:
crypto/internal/backend.envGoFIPS()
        /work/go/src/crypto/internal/backend/common.go:40 +0xf4
crypto/internal/backend.init.0()
        /work/go/src/crypto/internal/backend/common.go:16 +0xf
```

I tested on a Windows VM, and it behaves as expected there, failing if `HKLM\System\CurrentControlSet\Control\Lsa\FipsAlgorithmPolicy` isn't configured. Behavior on Mariner is enabling FIPS mode (I added a println to the backend to make sure) and on my dev Fedora machine (without FIPS OpenSSL) it results in an OpenSSL error as expected.

---

Updates the `crypto/internal/backend/common.go` logic to more generally catch when no backend is enabled and FIPS mode is requested. Now, you get this on Windows:

```
$ $env:GOFIPS='1'
$ go build . && .\exp                                           
panic: FIPS mode requested (environment variable GOFIPS=1) but no supported crypto backend is enabled
```

This follows the theme of some of the recent changes: it technically reduces our toolset's compatibility, but simplifies respecting the *intent* of the app's user and the chance of mistakes in less clear-cut situations.

It will also help if someone intended to add FIPS support to their build, but their change to the build process didn't work as they intended. E.g. tags specified in an env var are overridden if the `-tags` flag is set:

```
$ $env:GOFIPS='1'
$ $env:GOFLAGS='-tags=goexperiment.systemcrypto,requirefips'
$ go build -tags 'fun' . && .\exp
panic: FIPS mode requested (environment variable GOFIPS=1) but no supported crypto backend is enabled
```

It's hard to get into this state in other ways. The added build error does this:

```
$ $env:GOFIPS=''
$ go build -tags='requirefips' .
# runtime
..\go\go\src\runtime\backenderr_gen_requirefips_nosystemcrypto.go:12:2: `
        The requirefips tag is enabled, but no crypto backend is enabled.
        A crypto backend is required to enable FIPS mode.
        ` (untyped string constant "\n\tThe requirefips tag is enabled, but no crypto backend is enabled...) is not used
```

Generally speaking, though, I think failing at runtime if `GOFIPS=1` and `!backend.Enabled` is a reasonable backstop if there's anything we missed when coding the compile errors or updating build tags.

(This PR also moves the logic entirely into the backend foundation patch rather than part of it being in the CNG patch. It was inconsistent: the CNG part was implemented in the CNG patch, but the OpenSSL part was implemented in the foundation patch.)

---

The reason to panic on `GOFIPS=0` rather than just ignoring it is to make sure nobody's intent is being silently ignored--the binary's author or its user.

Say someone is building a system that uses a handful of prebuilt Go binaries. Some are compiled with `requirefips` and some aren't. The user needs the system to communicate with a legacy server using crypto that the FIPS standard doesn't allow. The user sees failures related to FIPS TLS restrictions and sets `GOFIPS=0`. If this were silently ignored, parts of the system without `requirefips` would function and parts with `requirefips` would keep failing, and there's no surface-level way to tell why. (If you dig, `go version -m` on a binary has enough information to see `requirefips`, but you'd have to know a lot about this system to get this far and recognize the tag.)

Having the affected binaries panic, instead, gives them a way forward. They could either compile the binaries themselves, or (particularly in an internal or Docker scenario) look or ask for a non-FIPS version of the binaries from whoever provided the ones they were originally using.

---

I'll submit another PR to update the FIPS docs to describe this change and also `systemcrypto`.

### DIFF
--- a/patches/0002-Add-crypto-backend-foundation.patch
+++ b/patches/0002-Add-crypto-backend-foundation.patch
@@ -17,11 +17,9 @@ Subject: [PATCH] Add crypto backend foundation
  src/crypto/hmac/hmac_test.go                 |   2 +-
  src/crypto/internal/backend/backend_test.go  |  30 +++++
  src/crypto/internal/backend/bbig/big.go      |  17 +++
- src/crypto/internal/backend/common.go        |  81 +++++++++++++
- src/crypto/internal/backend/iscgo.go         |  10 ++
+ src/crypto/internal/backend/common.go        |  78 +++++++++++++
  src/crypto/internal/backend/isrequirefips.go |   9 ++
  src/crypto/internal/backend/nobackend.go     | 116 +++++++++++++++++++
- src/crypto/internal/backend/nocgo.go         |  10 ++
  src/crypto/internal/backend/norequirefips.go |   9 ++
  src/crypto/internal/backend/stub.s           |  10 ++
  src/crypto/rand/rand_unix.go                 |   2 +-
@@ -40,14 +38,12 @@ Subject: [PATCH] Add crypto backend foundation
  src/crypto/tls/cipher_suites.go              |   2 +-
  src/go/build/deps_test.go                    |   2 +
  src/runtime/runtime_boring.go                |   5 +
- 36 files changed, 328 insertions(+), 29 deletions(-)
+ 34 files changed, 305 insertions(+), 29 deletions(-)
  create mode 100644 src/crypto/internal/backend/backend_test.go
  create mode 100644 src/crypto/internal/backend/bbig/big.go
  create mode 100644 src/crypto/internal/backend/common.go
- create mode 100644 src/crypto/internal/backend/iscgo.go
  create mode 100644 src/crypto/internal/backend/isrequirefips.go
  create mode 100644 src/crypto/internal/backend/nobackend.go
- create mode 100644 src/crypto/internal/backend/nocgo.go
  create mode 100644 src/crypto/internal/backend/norequirefips.go
  create mode 100644 src/crypto/internal/backend/stub.s
 
@@ -260,10 +256,10 @@ index 00000000000000..85bd3ed083f5b2
 +}
 diff --git a/src/crypto/internal/backend/common.go b/src/crypto/internal/backend/common.go
 new file mode 100644
-index 00000000000000..c69ee3b8b53c9e
+index 00000000000000..efdd080a1b7708
 --- /dev/null
 +++ b/src/crypto/internal/backend/common.go
-@@ -0,0 +1,81 @@
+@@ -0,0 +1,78 @@
 +// Copyright 2022 The Go Authors. All rights reserved.
 +// Use of this source code is governed by a BSD-style
 +// license that can be found in the LICENSE file.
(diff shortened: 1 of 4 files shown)


qmuntal (approved):

patches/0002-Add-crypto-backend-foundation.patch
```diff
@@ -273,25 +277,41 @@ index 00000000000000..007d8070538247
...
-+	if v, _ := envGoFIPS(); v == "1" {
-+		if runtime.GOOS != "linux" {
-+			panic("FIPS mode requested (GOFIPS=1) but not supported on " + runtime.GOOS)
-+		} else if !iscgo {
-+			panic("FIPS mode requested (GOFIPS=1) but not supported with cgo disabled")
++	if v, r, ok := envGoFIPS(); ok && v == "1" {
++		if runtime.GOOS != "linux" && runtime.GOOS != "windows" {
++			panic("FIPS mode requested (" + r + ") but not supported on " + runtime.GOOS)
++		}
++		if runtime.GOOS == "linux" && !iscgo {
```
qmuntal: Can this check be removed? We now panic at compile time when `opensslcrypto` is used without cgo instead of silently falling back to Go crypto, so it would be more informative to just say that no supported crypto backend is enabled.

If this check is removed, then we can also remove `iscgo.go` and `nocgo.go`.

dagood: True, you really have to know what you're doing to force this error to pop up now:

```
$ CGO_ENABLED=0 go build -tags=goexperiment.opensslcrypto,requirefips,allow_missing_crypto_backend_fallback . && ./exp
panic: FIPS mode requested (requirefips tag set) but not supported with cgo disabled
```

I also moved the `runtime.GOOS` check into `if !Enabled` because I think it shows the intent better: it's just an explanation, and `Enabled` is really the only thing that really needs to be enforced.

qmuntal (approved):

jaredpar:
Where do we document the new behavior though?

dagood:
I submitted https://github.com/microsoft/go/pull/970 to document this and also the `systemcrypto` change.

jaredpar (approved):

PR #970, opened 2023-07-04
Document systemcrypto, requirefips

dagood:
Add docs for:

* https://github.com/microsoft/go/pull/935 / https://github.com/microsoft/go/pull/950
* https://github.com/microsoft/go/pull/968

I tidied up a little, generally. I moved the detailed multi-GOEXPERIMENT into its own section (where explaining `-tags` fits too) and put the example build commands in collapsed sections to tighten up the section a little after adding the new examples.

Rendered: https://github.com/microsoft/go/blob/dev/dagood/doc-sys-req/eng/doc/fips/README.md


qmuntal (approved):

eng/doc/fips/README.md
```diff
@@ -37,59 +37,100 @@ It is important to note that an application built with Microsoft's Go toolchain
 
 The `GOEXPERIMENT` environment variable is used at build time to select a cryptographic library backend. This modifies the Go runtime included in the program to use the specified platform-provided cryptographic library whenever it calls a Go standard library crypto API. The `GOEXPERIMENT` values that pick a crypto backend are:
 
+- *Recommended:* `systemcrypto` automatically selects the crypto backend we suggest for the target platform
```
jaredpar:
```suggestion
- *Recommended:* `systemcrypto` automatically selects the suggested crypto backend for the target platform
```

jaredpar: Do we ever explicitly state that `systemcrypto` selects OpenSSL for Linux, CNG for windows, etc ... We give a few examples but there isn't really a full table that says what combo of values selects which crypto libraries. I was wondering if we should have something like 

|Target Platform| systemcrypto mapping | library |
| -- | -- | --| 
| Linux | opensslcrypto | OpenSSL |
| Windows | cngcrypto | Windows |

It's fairly implicit though so not sure if it's needed

eng/doc/fips/README.md
```diff
@@ -117,9 +158,25 @@ To make the Go runtime panic during program initialization if FIPS mode is not e
 
 > Unlike `opensslcrypto`, a Windows program built with `cngcrypto` doesn't include the ability to enable/disable FIPS, only ensure it's enabled. Windows FIPS mode is not a per-process setting, and changing it may require elevated permissions. Adding this feature would likely have unintended consequences.
 
-## Usage: Extra runtime configuration options
+## Usage: Extra configuration options
+
+### Enable FIPS mode at compile time
```
jaredpar: Enable or requires?

jaredpar (approved):

jaredpar:
Now there are a lot of different options to control FIPS behavior:

- GOEXPERIMENT to select the backend 
- requiresfips build tag 
- GOFIPS environment variable

Do we think it's clear what happens for all the varations on these flags?

eng/doc/fips/README.md
dagood: Hmm, yeah, the "we suggest" was meant to point to the Microsoft fork/project doing the suggesting (rather than upstream Go, which suggests using none of these 😄) but this isn't a great way to convey that. I like the idea of trying out a table. I think it would be good to mention that the mapping matches the internal crypto library usage policy requirements for Go. I'm thinking of even dropping the somewhat vague "recommend" term, which I think raises questions about why it's a recommendation (and is it *just* a recommendation?).

eng/doc/fips/README.md
dagood: I struggled with this header. 😄 I started with "Require FIPS mode at compile time", but I see that as saying the *compilation environment* is what should run in FIPS mode. "Enable FIPS mode" seemed a little better because it uses the terms set up in the previous section ("enable FIPS mode" could mean "GOFIPS=1"), but it's not really any good, seeing it again now. I think this fixes it:

```suggestion
### Build option to require FIPS mode
```

dagood (commented):
> Do we think it's clear what happens for all the varations on these flags?

Hmm, maybe not. I'll play with this. One thing in particular that I think will help is moving the "tutorial" info to a separate doc to keep these sections focused on the features we've added and how they interact.

qmuntal (approved):
LGTM++

dagood:
Last set of changes for sure this time. 😄 I want to get this in, then we can go back to smaller PRs.

I think having a dedicated changelog section is good--a simple place to put info about changes over time rather than just keeping notes around based on whether they seem useful or not. Anyone who might need to fix up a FIPS build after it sits for some amount of time could find it useful to get up to speed, too, especially if they don't follow our mailing list. (Or maybe they do, but this would be easier than looking back through to find what's relevant to building for FIPS.)

eng/doc/fips/README.md
```diff
@@ -350,6 +352,36 @@ The work done to support FIPS compatibility mode leverages code and ideas from o
...
+- Removes automatic Go standard library crypto fallback. If a crypto backend is selected but isn't supported, the build fails. Before 1.21, the build automatically uses the Go standard library crypto implementation, which may not be clear.
+- Adds [`systemcrypto` experiment alias](#usage-build).
+- Adds [`requirefips` build tag](#build-option-to-require-fips-mode).
+- Adds [`allowcryptofallback` experiment](#build-option-to-allow-go-standard-library-crypto-fallback).
+
+### Go 1.20.6 and 1.19.11 (Jul 2023)
+
+- When multiple versions of OpenSSL are present on the machine at runtime, a version with FIPS mode enabled now has higher priority than others. [microsoft/go-crypto-openssl@v0.2.8](https://github.com/microsoft/go-crypto-openssl/releases/tag/v0.2.8)
+
+### Go 1.19 (Aug 2022)
```
qmuntal, line 372: Worth noting that the CNG backend was introduced in Go 1.19

qmuntal (approved):

eng/doc/fips/README.md
```diff
@@ -33,69 +31,165 @@ On Linux, the fork uses [OpenSSL](https://www.openssl.org/) through the [go-cryp
...
+| `GOEXPERIMENT=systemcrypto` | `GOFIPS=1` | Compliant | Can be used to create a compliant app. The app either enables FIPS mode or ensures it is already enabled. Otherwise, the app panics. |
+| `GOEXPERIMENT=systemcrypto` | `GOFIPS=0` | Compliant | Crypto usage is not FIPS compliant. The app attempts to disable FIPS mode and panics if it isn't possible. |
+| `GOEXPERIMENT=systemcrypto` | `GO_OPENSSL_VERSION_OVERRIDE=1.1.1k-fips` | Compliant | Can be used to create a compliant app. If the app is built for Linux, `systemcrypto` chooses `opensslcrypto`, and the environment variable causes it to load `libcrypto.so.1.1.1k-fips` instead of using the automatic search behavior. This environment variable has no effect with `cngcrypto`. |
+| `GOEXPERIMENT=systemcrypto` and `-tags=requirefips` | Default | Compliant | Can be used to create a compliant app. The behavior is the same as `GOFIPS=1`, but no runtime configuration is necessary. See [the `requirefips` section](#build-option-to-require-fips-mode) for more information on when this "locked-in" approach may be useful rather than the flexible approach. |
```
jaredpar, line 65: Think we should list the `allowcryptofallback` case here. Customers found it before, they'll find it again, best to acknowledge the combo exists and call it out as explicitly not compliant.

eng/doc/fips/README.md
```diff
@@ -33,69 +31,165 @@ On Linux, the fork uses [OpenSSL](https://www.openssl.org/) through the [go-cryp
...
+
 ## Usage: Build
 
 The `GOEXPERIMENT` environment variable is used at build time to select a cryptographic library backend. This modifies the Go runtime included in the program to use the specified platform-provided cryptographic library whenever it calls a Go standard library crypto API. The `GOEXPERIMENT` values that pick a crypto backend are:
 
+- *Recommended:* `systemcrypto` automatically selects the suggested crypto backend for the target platform
+  > Prior to Go 1.21, this alias is not available and the backend must be selected manually
 - `opensslcrypto` selects OpenSSL, for Linux
 - `cngcrypto` selects CNG, for Windows
 - `boringcrypto` selects the upstream BoringCrypto backend, which is **not supported**
```
jaredpar: Consider `not supported nor compliant`. Above we use compliant but here supported. Should enforce that it's not supported nor is it compliant

eng/doc/fips/README.md
```diff
@@ -33,69 +31,165 @@ On Linux, the fork uses [OpenSSL](https://www.openssl.org/) through the [go-cryp
...
+- If no option is selected, Go standard library cryptography is used.
+
+The options are exclusive and must not be enabled at the same time as one another.
+
+`systemcrypto` matches the internal Microsoft crypto policy for Go:
+
+| Target platform | `systemcrypto` selection | Library |
+| --- | --- | --- |
+| Linux | `opensslcrypto` | OpenSSL |
+| Windows | `cngcrypto` | CNG |
```
jaredpar, line 106: Consider adding a line for Mac stating it's not supported for completeness.

jaredpar (approved):

eng/doc/fips/README.md
dagood: It actually is compliant if used correctly, it just introduces risk. In any case, sure, adding.

dagood: I put it into its own table at the end of this section instead to provide a little more context.

PR #972, opened 2023-07-05
Add the active crypto backend to build info (`go version -m app`)

### LINKED ISSUE #963: Add `debug.BuildSetting` entries corresponding to crypto backend usage
`go version -m example` reports the go experiment and some other info that can most likely be used to figure out whether a crypto module was used to build `example`, but it isn't straightforward. I think we can fairly easily add this info in a more consistent way:

https://pkg.go.dev/runtime/debug#BuildSetting

```diff
diff --git a/src/cmd/go/internal/load/pkg.go b/src/cmd/go/internal/load/pkg.go
index c0e6265e29d065..fdeb82b0484ec6 100644
--- a/src/cmd/go/internal/load/pkg.go
+++ b/src/cmd/go/internal/load/pkg.go
@@ -2430,6 +2430,9 @@ func (p *Package) setBuildInfo(ctx context.Context, autoVCS bool) {
 	if key, val := cfg.GetArchEnv(); key != "" && val != "" {
 		appendSetting(key, val)
 	}
+	if ok := cfg.BuildContext.MatchTag("goexperiment.opensslcrypto"); ok {
+		appendSetting("goexperiment.opensslcrypto", "on")
+	}
 
 	// Add VCS status if all conditions are true:
 	//
diff --git a/src/go/build/build.go b/src/go/build/build.go
index 48adcfed5cf3cb..8f757a87c09257 100644
--- a/src/go/build/build.go
+++ b/src/go/build/build.go
@@ -2021,6 +2021,10 @@ func (ctxt *Context) matchTag(name string, allTags map[string]bool) bool {
 	return false
 }
 
+func (ctxt *Context) MatchTag(tag string) bool {
+	return ctxt.matchTag(tag, nil)
+}
+
 // goodOSArchFile returns false if the name contains a $GOOS or $GOARCH
 // suffix which does not match the current system.
 // The recognized name formats are:
```

This works as expected with `go version -m`, and other tools can probably be adapted to read the data directly from Go binaries if necessary (e.g. a Go toolset isn't present).

dagood:
* Resolves https://github.com/microsoft/go/issues/963
  * There is a different mechanism we could use (see [comment on #963](https://github.com/microsoft/go/issues/963#issuecomment-1622686013)) but I think this probably the way to go.

This change makes it easier to tell what backend a program ended up being built with by adding lines like `cryptobackend=openssl` to the build info key-value pairs:

```
$ GOEXPERIMENT=systemcrypto go build . && go version -m exp
exp: devel go1.21-a011bc5ece Mon Jul 3 13:03:24 2023 -0500 X:systemcrypto
        build   GOEXPERIMENT=systemcrypto
        build   GOOS=linux
        build   cryptobackend=openssl
... [more pairs above and below this output]

$ go build -tags=goexperiment.systemcrypto . && go version -m exp
exp: devel go1.21-a011bc5ece Mon Jul 3 13:03:24 2023 -0500
        build   -tags=goexperiment.systemcrypto
        build   cryptobackend=openssl
...

$ GOOS=windows GOEXPERIMENT=systemcrypto go build . && go version -m exp.exe
exp.exe: devel go1.21-a011bc5ece Mon Jul 3 13:03:24 2023 -0500 X:systemcrypto
        build   GOEXPERIMENT=systemcrypto
        build   GOOS=windows
        build   cryptobackend=cng
...
```

Note that `X:systemcrypto` shows up when using `GOEXPERIMENT`, but `-tags` doesn't show up there. The `-tags` line only shows up when the `-tags` flag is used for `go build`. Technically, someone (or a program) can look for these signs and figure out what the program is using by understanding the build logic, but this could be fairly fragile. Instead, we can include the info directly.

buildinfo was added in 1.18.

The built-in pairs are defined at https://pkg.go.dev/runtime/debug#BuildSetting, and I haven't seen any reason we *shouldn't* add our own pairs. The most likely break I can think of is that an external tool will fail if it sees an unrecognized pair, but it seems more likely to me that it would skip them.

### DIFF
--- a/patches/0001-Add-systemcrypto-GOEXPERIMENT.patch
+++ b/patches/0001-Add-systemcrypto-GOEXPERIMENT.patch
@@ -8,12 +8,18 @@ goexperiment.systemcrypto behave as an alias that enables the recommended
 backend for the target GOOS. See src/internal/goexperiment/flags.go for more
 information about the behavior.
 
+Includes active crypto backend in the build info accessible in the
+binary, for example by "go version -m". This makes it easy to determine
+which backend is being used by a compiled Go program no matter which
+GOEXPERIMENT or build tag was used to enable it.
+
 Includes new tests in "build_test.go" and "buildbackend_test.go" to help
 maintain this feature. For more information, see the test files.
 ---
+ src/cmd/go/internal/load/pkg.go               |  5 ++
  src/cmd/go/internal/modindex/build.go         | 54 ++++++++++++++
  src/cmd/go/internal/modindex/build_test.go    | 73 +++++++++++++++++++
- src/go/build/build.go                         | 54 ++++++++++++++
+ src/go/build/build.go                         | 70 ++++++++++++++++++
  src/go/build/buildbackend_test.go             | 66 +++++++++++++++++
  .../testdata/backendtags_openssl/main.go      |  3 +
  .../testdata/backendtags_openssl/openssl.go   |  3 +
@@ -22,7 +28,7 @@ maintain this feature. For more information, see the test files.
  .../goexperiment/exp_systemcrypto_off.go      |  9 +++
  .../goexperiment/exp_systemcrypto_on.go       |  9 +++
  src/internal/goexperiment/flags.go            | 15 ++++
- 11 files changed, 292 insertions(+)
+ 12 files changed, 313 insertions(+)
  create mode 100644 src/cmd/go/internal/modindex/build_test.go
  create mode 100644 src/go/build/buildbackend_test.go
  create mode 100644 src/go/build/testdata/backendtags_openssl/main.go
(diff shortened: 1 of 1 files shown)


karianna:
@dagood We intending to keep this PR? I noted it was a year old :-)

dagood:
I would still argue that we should do this. I think now, it's even more important given https://github.com/microsoft/go/issues/1352 (@qmuntal). We should probably also add some straightforward "used Microsoft Go" data (because there will no longer be an explicit GOEXPERIMENT or tag).

dagood:
Technically slightly different from https://github.com/microsoft/go/pull/1712, but https://github.com/microsoft/go/pull/1712 means that if we do need anything more stored in the buildinfo, we have an example and this draft PR is not useful. Closing.

PR #975, opened 2023-07-11
Remove !cmd_go_bootstrap and !msan build tags

qmuntal:
This PR removes the `!cmd_go_bootstrap` and `!msan` build tags from the CNG and the OpenSSL backends. The main benefit is that the error messages added in #965 are much more cleaner. It will be eventually simplified even more when https://github.com/golang-fips/openssl/pull/90 is merged and we start using that module.

The `!msan` build tag was inherited from the Boring backend, which statically links against a precompiled syso file, and syso files are not supported with `msan`. The other backends don't use this approach, so it's not necessary.

The `!cmd_go_bootstrap` was necessary when the Go compiler depended on `crypto/sha256`, and there were some issues using Boring in the compiler. Since [CL 402595](https://go-review.googlesource.com/c/go/+/402595), the Go compiler doesn't depend on any `crypto` package, so the build tag can be removed. I've kept it on the Boring backend, as upstream still uses it. I'll submit an upstream PR to fix it. Done: https://go-review.googlesource.com/c/go/+/508402

While here, I've updated the Boring backend tags to match upstream, which recently added `arm64` support for it.


chsienki (approved):

patches/0003-Add-BoringSSL-crypto-backend.patch
```diff
@@ -30,15 +30,15 @@ index 00000000000000..0b62cef68546d0
...
+index 00000000000000..3b5504b6afc5c6
 --- /dev/null
 +++ b/src/crypto/internal/backend/boring_linux.go
 @@ -0,0 +1,135 @@
 +// Copyright 2017 The Go Authors. All rights reserved.
 +// Use of this source code is governed by a BSD-style
 +// license that can be found in the LICENSE file.
 +
-+//go:build goexperiment.boringcrypto && linux && cgo && amd64 && !android && !cmd_go_bootstrap && !msan
++//go:build goexperiment.boringcrypto && linux && cgo && (amd64 || arm64) && !android && !msan
```
dagood, line 41: It looks like we might want to apply the platform change to 1.20 as well since this is included in some of those releases: https://github.com/golang/go/commit/39ec97da15e9167e7a17746c643f264baae6ce7b. However, doesn't seem like a priority because we don't anticipate anyone really depending on boring with our fork.

dagood (approved):

patches/0003-Add-BoringSSL-crypto-backend.patch
qmuntal: Good catch, will do.

PR #982, opened 2023-07-13
[microsoft/release-branch.go1.21] Update 1.21 branch

dagood:
Update the submodule to the current 1.21 commit and fix patches. The usual trivial conflicts in go.mod/go.sum and vendoring.

I mentioned in Go sync that I saw this branch and remembered that upstream typically branches for the first `rc` release. This has been the case before, but as far as I can tell from https://github.com/golang/go/commits/release-branch.go1.21, the branch was just created today. (I don't know a reasonable way to tell when a branch was actually created on GitHub, but what is easy to see is that today is when the first branch-specific commit was pushed.)

### DIFF
--- a/eng/_core/cmd/build/build.go
+++ b/eng/_core/cmd/build/build.go
@@ -133,6 +133,10 @@ func build(o *options) error {
 		return err
 	}
 
+	if o.Experiment != "" {
+		buildutil.AppendExperimentEnv(o.Experiment)
+	}
+
 	if !o.SkipBuild {
 		// If we have a stage 0 copy of Go in an env variable (as set by run.ps1), use it in the
 		// build command by setting GOROOT_BOOTSTRAP. The upstream build script "make.bash" uses
@@ -224,17 +228,6 @@ func build(o *options) error {
 			// The stderr output isn't used to determine whether the tests succeeded or not. (The
 			// redirect doesn't cause an issue where tests succeed that should have failed.)
 			testCmd.Stderr = os.Stdout
-
-			if o.Experiment != "" {
-				buildutil.AppendExperimentEnv(o.Experiment)
-				// We aren't able to rebuild Go standard library packages under a crypto experiment,
-				// but "cmd/dist/test.go" will normally do this for local test runs to make sure the
-				// build is clean. The problem is that the build doesn't include cgo, which is
-				// required for some crypto backends. Set this variable specific to Microsoft Go to
-				// indicate that our build is fresh because it's running within our scripts and
-				// doesn't need a rebuild. A patch in the test runner detects this.
-				os.Setenv("GO_MSFT_SCRIPTED_BUILD", "1")
-			}
 
 			return runCmd(testCmd)
 		}
--- a/eng/_core/buildutil/buildutil.go
+++ b/eng/_core/buildutil/buildutil.go
@@ -9,6 +9,7 @@ import (
 	"log"
 	"os"
 	"strconv"
+	"strings"
 )
 
 // Retry runs f until it succeeds or the attempt limit is reached.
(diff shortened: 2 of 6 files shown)


jaredpar (approved):

eng/_core/cmd/build/build.go
```diff
@@ -234,6 +236,27 @@ func build(o *options) error {
...
+						// Failed to read for an unexpected reason.
+						return err
+					}
+				} else {
+					// As of Go 1.21, the VERSION file has multiple lines. The first line is just
+					// the version number.
+					before, _, _ := strings.Cut(string(data), "\n")
+					os.Setenv("TESTGO_TOOLCHAIN_VERSION", before)
+					fmt.Printf("---- VERSION file found with first line %q. Assigning TESTGO_TOOLCHAIN_VERSION for test run.\n", before)
+				}
```
dagood: Notes: I believe this problem showed up in this release because I moved `buildutil.AppendExperimentEnv(o.Experiment)` for 1.21. Now, the toolchain is built *without* `opensslcrypto` and tested *with* `opensslcrypto`. I think this mismatch is part of the problem.

In earlier releases, we're able to build Go with `opensslcrypto` because it's able to fall back to Go crypto. The check I added in this release makes that into a failure instead of fallback. There was no clear way to wire up the bootstrap builds to allow fallback while preventing it for "real app" builds, so building Go without `opensslcrypto` then testing with `opensslcrypto` was the solution I settled with.

As of writing, I don't know if this workaround works. Another approach is to revisit the possibility of allowing the Go build to fall back to Go crypto so we can build+test with `opensslcrypto` again.

qmuntal: > Notes: I believe this problem showed up in this release because I moved buildutil.AppendExperimentEnv(o.Experiment) for 1.21. Now, the toolchain is built without opensslcrypto and tested with opensslcrypto. I think this mismatch is part of the problem.

I guess you are talking about #965, no?

> As of writing, I don't know if this workaround works. Another approach is to revisit the possibility of allowing the Go build to fall back to Go crypto so we can build+test with opensslcrypto again.

Difficult problem. The toolchain and the tests fight against the approach implemented in #965. What if we relax the check allowing the fallback to plain Go unless `GOFIPS=1` or `-tags=requirefips` are set? We would be a little bit better than in Go 1.20, which didn't had the `requirefips` knob. We can try to come up with something better for go1.22.

qmuntal: Falling back to plain Go has been disruptive for us, so I would expect some downstream users to also have unintentional compilation or test failures. For example, https://github.com/Azure/ClusterConfigurationAgent/pull/1683 intentionally modified all Docker files to compile Go binaries using `CGO_ENABLED=0 GOEXPERIMENT=opensslcrypto` meanwhile they figure out the performance impact of enabling cgo. When they move to go1.21 as is today, all they pipelines will break.

dagood: > I guess you are talking about #965, no?

Right.

> What if we relax the check allowing the fallback to plain Go unless `GOFIPS=1` or `-tags=requirefips` are set?

Maybe, but I think a problem is the (large number of?) programs that will need to comply with internal Microsoft crypto policy, which may never have FIPS mode enabled. But maybe this isn't the right way to give them a guardrail. (More about it in Go sync chat.)

> intentionally modified all Docker files to compile Go binaries using `CGO_ENABLED=0 GOEXPERIMENT=opensslcrypto`

I think this is the sort of potentially misleading config that we want to prevent, although you're right, it's a breaking change and that deserves weight.

---

I think I applied the workaround to the wrong place and it only applied to devscript builders. I'm going to try it again just to see if it would theoretically work, but I do think reverting this change is probably what will end up making the most sense.

dagood: The workaround still doesn't work. Good to know. 😄 Running a revert, and keeping in mind that I need to do it on `microsoft/main` as well.

dagood: We decided in the Go sync chat to keep preventing accidental Go fallback. The situation changed since we started allowing fallback (pre-1.19) and it seems like now it helps without much downside.

I added a note to the fallback compile error message that directly addresses the breaking change.

Meanwhile, I figured out that if we change the `allow_missing_crypto_backend_fallback` tag to a `allowcryptofallback` GOEXPERIMENT, the complicated workarounds aren't necessary because `opensslcrypto,allowcryptofallback` follow each other around and take effect in the same places (and they are omitted in the same places).

dagood: * Here's the main PR for this improvement: https://github.com/microsoft/go/pull/984

This 1.21 PR has the same changes (clean cherry-pick) so once the main PR goes in, I'll consider this PR ready to go.

qmuntal (approved):

PR #984, opened 2023-07-14
Go crypto fallback: use GOEXPERIMENT for opt-in

dagood:
https://github.com/microsoft/go/pull/965 changed Microsoft Go behavior so that instead of falling back to Go crypto, the build will fail if a backend is enabled and its requirements aren't met.

This caused issues in the Go build and test because in many places, cgo isn't enabled and cross-builds are being done. So, https://github.com/microsoft/go/pull/965 added `-tags=allow_missing_crypto_backend_fallback` to opt out of this behavior, and disabled the crypto backend during the build part of CI, only enabling it during test. Passing the tags to the right parts of the tests is tricky, and other workarounds were needed too to account for using different GOEXPERIMENT values for build vs. test. Once the repo has a `VERSION` file, (https://github.com/microsoft/go/pull/982) it gets even more complicated, and ultimately I didn't get this approach working in the 1.21 release branch.

This PR changes the `allow_missing_crypto_backend_fallback` tag to a `allowcryptofallback` GOEXPERIMENT. This means the complicated workarounds aren't necessary because `opensslcrypto,allowcryptofallback` follow each other around and take effect in the same places (and they are omitted in the same places).

I also added a note to the fallback compile error message that directly addresses the breaking change, for anyone depending on the fallback behavior.

Open to names other than `allowcryptofallback`, although I'm happy with it. Experiments are inherently easier to find than build tags, and someone might have a convoluted build script where they would actually want to use it (although I plan to discourage using it in the FIPS doc), so I figure a reasonable name would be good to have.

### DIFF
--- a/patches/0007-Add-backend-code-gen.patch
+++ b/patches/0007-Add-backend-code-gen.patch
@@ -25,28 +25,25 @@ harder to update the generators and to deal with conflicts.
 Use "go/bin/go generate crypto/internal/backend" after recently building
 the repository to run the generators.
 ---
- src/cmd/dist/build.go                         |   2 +-
- src/cmd/dist/test.go                          |   9 +-
- src/cmd/internal/obj/x86/pcrelative_test.go   |   4 +
- src/cmd/internal/testdir/testdir_test.go      |   4 +
- src/cmd/link/internal/ld/stackcheck_test.go   |   2 +-
- src/cmd/link/link_test.go                     |   4 +-
- src/cmd/vet/vet_test.go                       |   2 +-
  src/crypto/internal/backend/backendgen.go     |  20 ++
- .../internal/backend/backendgen_test.go       | 279 ++++++++++++++++++
+ .../internal/backend/backendgen_test.go       | 284 ++++++++++++++++++
  src/crypto/internal/backend/nobackend.go      |   2 +-
+ .../exp_allowcryptofallback_off.go            |   9 +
+ .../exp_allowcryptofallback_on.go             |   9 +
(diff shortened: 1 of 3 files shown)


patches/0007-Add-backend-code-gen.patch
```diff
@@ -392,7 +264,12 @@ index 00000000000000..1cc4952e7ca9c2
 +		"The "+expTag.String()+" tag is specified, but other tags required to enable that backend were not met.",
 +		"Required build tags:",
 +		"  "+backend.constraint.String(),
-+		"Please check your build environment and build command for a reason one or more of these tags weren't specified.")
++		"Please check your build environment and build command for a reason one or more of these tags weren't specified.",
++		"",
++		"If you only performed a Go toolset upgrade and didn't expect this error, your code was likely depending on fallback to Go standard library crypto.",
++		"As of Go 1.21, Go crypto fallback is a build error. This helps prevent accidental fallback.",
++		"Removing "+backend.name+"crypto will restore pre-1.21 behavior by intentionally using Go standard library crypto.",
```
dagood, line 271: Note that I'm not mentioning `allowcryptofallback` here. I think it's probably exceptionally rare that someone actually *needs* to use it rather than just removing the crypto backend they aren't using. I don't want to inadvertently encourage people to try it out without understanding the context--I think it would be better to describe it in the FIPS readme where there's more room for detail and we can update it more freely.

chsienki (approved):

qmuntal (approved):

PR #990, opened 2023-07-26
Use distpack to create archives

### LINKED ISSUE #140: Include Go version in name of artifact (tar.gz, zip)
We should consider including the Go version in these names (when we are building a versioned branch):
* go.20210716.4.linux-amd64.tar.gz
* go.20210716.4.windows-amd64.zip

The official tar.gz/zip files look like this:
* go1.16.6.linux-amd64.tar.gz
* go1.16.6.windows-amd64.zip

Upstream checks in a `VERSION` file when they fork a branch, then update it with every patch release:
https://github.com/microsoft/go/blob/bc51e930274a5d5835ac8797978afc0864c9e30c/VERSION#L1

We could use that file to include *both* the Go version and Microsoft build number: `go1.16.6.20210716.4.linux-amd64.tar.gz`.

Or we could remove the build number to match upstream perfectly, however we should be sure that you can trace a given Go build back to the AzDO build that produced it. (We could simply put our own new file with this info into `GOROOT`, as a sibling of the existing `VERSION` file in `GOROOT`.)

I don't know of any *particularly* strong reasons to do this, though:
* Potentially easier drop-in replacement if we match the upstream artifact name exactly.
* Easier to tell which `tar.gz` is which if you've downloaded a few of them over time. (Build number is not mappable back to a branch name without looking for the build on AzDO and seeing which branch it used.)

### LINKED ISSUE #912: Use `-distpack` to create the tar.gz and zip archives of our builds

dagood:
* Resolves #912 
* Resolves https://github.com/microsoft/go/issues/140

Remove our own code to create build/source archives and use upstream's distpack tool instead.

distpack only works when a VERSION file is present (release branches), but we build `microsoft/main` tracking upstream `master`, which doesn't have a VERSION. In this case, I generate a temp VERSION file. This results in local outputs like:

* go1.22-82d80f8baa-dev.linux-amd64.tar.gz
* go1.22-82d80f8baa-dev.src.tar.gz

I tested a build with the submodule set to 1.21rc3 without any patches, and the local build produced:

* go1.21rc3-dev.linux-amd64.tar.gz
* go1.21rc3-dev.src.tar.gz

There was no file diff vs. upstream's build on go.dev/dl. After adding patches, only the source files we added showed up in the file diff.

This change adds the AzDO build number (locally, `dev`) so we don't end up with the same tar.gz name for multiple builds that use the same submodule commit. This isn't important for our own infra to work properly, but other infra might cache our artifacts based on filename and it's worth adding uniqueness to keep that working. Also, for devs downloading multiple builds to the same download folder over time, it helps differentiate the files without dealing with `(1)`, `(2)`, etc.

Latest test build here: https://dev.azure.com/dnceng/internal/_build/results?buildId=2230229&view=results


qmuntal (approved):
Nice!

eng/_core/cmd/build/build.go
```diff
@@ -237,18 +244,121 @@ func build(o *options) error {
...
+	}
+	fields := strings.Fields(string(vBytes))
+	if len(fields) < 2 {
+		return "", fmt.Errorf("expected at least 2 fields in dist version output, got %q in %q", len(fields), string(vBytes))
+	}
+	if fields[0] != "devel" {
+		return "", fmt.Errorf("expected first field 'devel' in dist version, got %q", fields[0])
+	}
+	// The second field should be something like "go1.21-abcde1234", and the remaining fields are a
+	// timestamp. Just using the second field is: the full VERSION file string is placed into the
```
chsienki: nit:

```suggestion
	// timestamp. Just using the second field as is: the full VERSION file string is placed into the
```

chsienki (approved):
LGTM

PR #994, opened 2023-08-01
Add ecdh to FIPS user guide

qmuntal:
`crypto/ecdh` was added in go1.20, but it is still not documented in our FIPS user guide. Fix that now.

### DIFF
--- a/eng/doc/fips/UserGuide.md
+++ b/eng/doc/fips/UserGuide.md
@@ -23,6 +23,12 @@ The Go crypto documentation is available online at https://pkg.go.dev/crypto.
       - [func StreamWriter.Close](#func-streamwriterclose)
     - [crypto/des](#cryptodes)
     - [crypto/dsa](#cryptodsa)
+    - [crypto/ecdh](#cryptoecdh)
+      - [func P256](#func-p256)
+      - [func P384](#func-p384)
+      - [func P521](#func-p521)
+      - [func X25519](#func-x25519)
+      - [func PrivateKey.ECDH](#func-privatekeyecdh)
     - [crypto/ecdsa](#cryptoecdsa)
       - [func Sign](#func-sign)
       - [func SignASN1](#func-signasn1)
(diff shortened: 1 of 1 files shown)


eng/doc/fips/UserGuide.md
```diff
@@ -30,6 +30,12 @@ The Go crypto documentation is available online at https://pkg.go.dev/crypto.
       - [func VerifyASN1](#func-verifyasn1)
       - [func GenerateKey](#func-generatekey)
       - [func PrivateKey.Sign](#func-privatekeysign)
+    - [crypto/ecdh](#cryptoecdh)
```
dagood: I think `ecdh` should be before `ecdsa` to keep alphabetical order.

eng/doc/fips/UserGuide.md
```diff
@@ -523,13 +529,140 @@ The message is signed using [BCryptSignHash].
...
+
+ecdh.X25519 is not implemented by any backend.
+
+#### func [PrivateKey.ECDH](https://pkg.go.dev/crypto/ecdh#PrivateKey.ECDH)
+
+```go
+func (k *ecdh.PrivateKey) ECDH(remote *ecdh.PublicKey) ([]byte, error)
+```
+
+ECDH performs a ECDH exchange and returns the shared secret. The PrivateKey and PublicKey must use the same curve.
```
```suggestion
ECDH performs an ECDH exchange and returns the shared secret. The PrivateKey and PublicKey must use the same curve.
```

It looks like https://pkg.go.dev/crypto/ecdh#PrivateKey.ECDH uses "a", but https://pkg.go.dev/crypto/ecdh#PublicKey and https://pkg.go.dev/crypto/ecdh#PrivateKey use "an". 😄

eng/doc/fips/UserGuide.md
```diff
@@ -523,13 +529,140 @@ The message is signed using [BCryptSignHash].
...
+</details>
+
 ### [crypto/ed25519](https://pkg.go.dev/crypto/ed25519)
 
 Not implemented by any backend.
 
 ### [crypto/elliptic](https://pkg.go.dev/crypto/elliptic)
 
-Not implemented by any backend.
+Not implemented by any backend, but the following `elliptic.Curve` constructors can be used together with `crypto/ecdsa`.
```
dagood: I think this would be easier to understand if it were more specific... here's a shot at it. (Otherwise it seems to me like a contradiction to say it's not implemented by a backend yet it's ok to do.)

```suggestion
Not implemented by any backend, but to use `ecdsa.GenerateKey`, one of the following `elliptic.Curve` constructors must be used to specify the curve. See [`ecdsa.GenerateKey`](#func-generatekey) for additional requirements. As long as the requirements are met, only the name of the curve is used, not the curve parameters or methods implemented by standard Go crypto, allowing FIPS compliance.
```

eng/doc/fips/UserGuide.md
qmuntal: Done!

dagood (approved):

chsienki (approved):

PR #1017, opened 2023-08-10
Update SDL script's path-unifying regex

dagood:
* Fixes https://github.com/microsoft/go/issues/1016

Tested it out in https://dev.azure.com/dnceng/internal/_build/results?buildId=2241148&view=results:

```
Renaming 'D:\a\_work\1\a\artifacts\go1.22-cd589c8-20230810.3.linux-amd64.tar.gz.extracted' to 'D:\a\_work\1\a\artifacts\go.linux-amd64.tar.gz.extracted'
Renaming 'D:\a\_work\1\a\artifacts\go1.22-cd589c8-20230810.3.linux-arm.tar.gz.extracted' to 'D:\a\_work\1\a\artifacts\go.linux-arm.tar.gz.extracted'
Renaming 'D:\a\_work\1\a\artifacts\go1.22-cd589c8-20230810.3.linux-arm64.tar.gz.extracted' to 'D:\a\_work\1\a\artifacts\go.linux-arm64.tar.gz.extracted'
Renaming 'D:\a\_work\1\a\artifacts\go1.22-cd589c8-20230810.3.src.tar.gz.extracted' to 'D:\a\_work\1\a\artifacts\go.src.tar.gz.extracted'
Renaming 'D:\a\_work\1\a\artifacts\go1.22-cd589c8-20230810.3.windows-amd64.zip.extracted' to 'D:\a\_work\1\a\artifacts\go.windows-amd64.zip.extracted'
```

Also removes `-BreakOnFailure $true`. Every pipeline run fails: the tool detects the same set of test secrets and reports them to the TSA service. The pipeline doesn't know whether or not the issues will be suppressed. Removing the constant source of failure means that more important failures (infra breaks) will show up. We can rely on TSA to show us new/important reported failures.

PR #1021, opened 2023-08-11
Change "arm" to "armv6l" during tar.gz prep

dagood:
distpack produces files like `go1.21.0.linux-arm.tar.gz`, but upstream published it as `go1.21.0.linux-armv6l.tar.gz`. This matches the historical name. I _believe_ this line is responsible for upstream's rename: https://github.com/golang/build/blob/a929908712e90b6259357273fafcfc08f7215193/internal/relui/workflows.go#L1455

Unfortunately for us, since upstream seems to rename the file in `x/build` rather than distpack, we have to reimplement the name swap in our own code.

I noticed this because some SDL issue suppressions were no longer applying due to the filename change, but it actually would have broken the aka.ms links at https://github.com/microsoft/go/blob/microsoft/main/eng/doc/Downloads.md. Luckily we aren't using distpack yet in 1.21, so the 1.21.0 release's aka.ms links are fine and we have time to fix this up for 1.22.

### DIFF
--- a/eng/_core/cmd/build/build.go
+++ b/eng/_core/cmd/build/build.go
@@ -282,9 +282,16 @@ func build(o *options) error {
 		// patches but build the same submodule commit multiple times.
 		buildID := getBuildID()
 		if o.PackBuild {
+			// distpack calls GOARCH=arm "arm" in its tar.gz filename, but the upstream release
+			// process changes it to "armv6l" on https://go.dev/dl/ to match the historical name.
+			// Do the same here.
+			brandingTargetArch := targetArch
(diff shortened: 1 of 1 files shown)


qmuntal (approved):

dagood:
> aka.ms links

After mulling this over, also realized that the microsoft/go-infra aka.ms infra is not set up to handle the new filenames due to the version number now being included, similar to what I fixed in https://github.com/microsoft/go/pull/1017/files. Working on that too. (It's independent of this PR.)

chsienki (approved):

PR #1024, opened 2023-08-22
Use openssl/cng to implement HKDF in crypto/tls

qmuntal:
This PR updates `crypto/tls` to use the HKDF implementations provided by the CNG/OpenSSL backends instead of the one in `x/crypto/hkdf`. The Boring backend doesn't implement HKDF, so it will fallback to `x/crypto/hkdf`.

I've upgraded the CNG backend to latest, the previous version didn't support HKDF.

### DIFF
--- a/patches/0002-Add-crypto-backend-foundation.patch
+++ b/patches/0002-Add-crypto-backend-foundation.patch
@@ -17,9 +17,9 @@ Subject: [PATCH] Add crypto backend foundation
  src/crypto/hmac/hmac_test.go                 |   2 +-
  src/crypto/internal/backend/backend_test.go  |  30 +++++
  src/crypto/internal/backend/bbig/big.go      |  17 +++
- src/crypto/internal/backend/common.go        |  78 +++++++++++++
+ src/crypto/internal/backend/common.go        |  78 ++++++++++++
  src/crypto/internal/backend/isrequirefips.go |   9 ++
- src/crypto/internal/backend/nobackend.go     | 116 +++++++++++++++++++
+ src/crypto/internal/backend/nobackend.go     | 125 +++++++++++++++++++
  src/crypto/internal/backend/norequirefips.go |   9 ++
  src/crypto/internal/backend/stub.s           |  10 ++
  src/crypto/rand/rand_unix.go                 |   2 +-
@@ -38,7 +38,7 @@ Subject: [PATCH] Add crypto backend foundation
  src/crypto/tls/cipher_suites.go              |   2 +-
  src/go/build/deps_test.go                    |   2 +
  src/runtime/runtime_boring.go                |   5 +
- 34 files changed, 305 insertions(+), 29 deletions(-)
+ 34 files changed, 314 insertions(+), 29 deletions(-)
  create mode 100644 src/crypto/internal/backend/backend_test.go
  create mode 100644 src/crypto/internal/backend/bbig/big.go
  create mode 100644 src/crypto/internal/backend/common.go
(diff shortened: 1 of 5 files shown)


dagood (approved):

patches/0002-Add-crypto-backend-foundation.patch
```diff
@@ -355,10 +355,10 @@ index 00000000000000..e5d7570d6d4363
 +const isRequireFIPS = true
 diff --git a/src/crypto/internal/backend/nobackend.go b/src/crypto/internal/backend/nobackend.go
 new file mode 100644
-index 00000000000000..ad6081552af15d
+index 00000000000000..3698bbac41e5c8
 --- /dev/null
 +++ b/src/crypto/internal/backend/nobackend.go
-@@ -0,0 +1,116 @@
+@@ -0,0 +1,125 @@
 +// Copyright 2017 The Go Authors. All rights reserved.
```
karianna, line 362: I assume by convention we would not update the copyright date?

patches/0004-Add-OpenSSL-crypto-backend.patch
```diff
@@ -550,6 +560,54 @@ index f8485dc3ca1c29..9c1d3d279c472f 100644
...
++++ b/src/crypto/tls/key_schedule.go
+@@ -7,9 +7,11 @@ package tls
+ import (
+ 	"crypto/ecdh"
+ 	"crypto/hmac"
++	boring "crypto/internal/backend"
+ 	"errors"
+ 	"fmt"
+ 	"hash"
++	"internal/goexperiment"
```
karianna, line 575: I assume this is a MSFT only internal package where we have experimental functions/features etc?  Do we feature flag experimental features?

patches/0004-Add-OpenSSL-crypto-backend.patch
```diff
@@ -550,6 +560,54 @@ index f8485dc3ca1c29..9c1d3d279c472f 100644
...
+ 	"io"
+ 
+ 	"golang.org/x/crypto/cryptobyte"
+@@ -59,7 +61,16 @@ func (c *cipherSuiteTLS13) expandLabel(secret []byte, label string, context []by
+ 		panic(fmt.Errorf("failed to construct HKDF label: %s", err))
+ 	}
+ 	out := make([]byte, length)
+-	n, err := hkdf.Expand(c.hash.New, secret, hkdfLabelBytes).Read(out)
++	var r io.Reader
```
karianna, line 584: I see we have a naming convention of a single character for 'obvious' types (such as 'r' for Reader), can I assume 'n' is for number and 'v' is value or void?

patches/0004-Add-OpenSSL-crypto-backend.patch
qmuntal: `internal/goexperiment` is an internal package also available upstream, which basically contains feature flags for experimental features. We patch that package to add MSFT-only features. You can see the FIPS-related ones documented in here: https://github.com/microsoft/go/tree/microsoft/main/eng/doc/fips#usage-build

patches/0004-Add-OpenSSL-crypto-backend.patch
qmuntal: It is idiomatic to use short names, up to a point. You got `r` and `n` right. `v` is not that common, I was thinking of it as `value`, but I now see that it is confusing, it would make more sense to give it a meaningful name, such as `prk` (pseudorandom key), which is what `hkdf.Extract` conceptually returns.

patches/0002-Add-crypto-backend-foundation.patch
qmuntal: Yep, that's the upstream convention too.

patches/0006-Vendor-crypto-backends.patch
```diff
@@ -5773,9 +5777,469 @@ index 00000000000000..a77ff97bb8f521
...
++type hashX struct {
++	h         bcrypt.ALG_HANDLE
++	ctx       bcrypt.HASH_HANDLE
++	size      int
++	blockSize int
++	buf       []byte
++	key       []byte
++}
++
++func newHashX(id string, flag bcrypt.AlgorithmProviderFlags, key []byte) *hashX {
```
chsienki, line 5970: Do we ever pass anything other than `bcrypt.ALG_NONE_FLAG` to this?

chsienki (approved):

PR #1036, opened 2023-09-08
Implement crypto/tls PRF using OpenSSL/CNG

qmuntal:
This PR updates `crypto/tls` to implement PRF using OpenSSL/CNG instead of plain Go.

The Go implementation doesn't return an error, but OpenSSL/CNG do. I had to update a bunch of functions so that the error is correctly handled.

I've also updated both backends to their latest versions, as TLS1PRF support landed soon in there.

The final implementation will probably change a bit once https://github.com/golang-fips/openssl/pull/116 and https://github.com/microsoft/go-crypto-winnative/pull/45 are merged, but I prefer to submit this now, get some reviews and let the pipeline exercise the new code.

### DIFF
--- a/patches/0002-Add-crypto-backend-foundation.patch
+++ b/patches/0002-Add-crypto-backend-foundation.patch
@@ -17,9 +17,9 @@ Subject: [PATCH] Add crypto backend foundation
  src/crypto/hmac/hmac_test.go                 |   2 +-
  src/crypto/internal/backend/backend_test.go  |  30 +++++
  src/crypto/internal/backend/bbig/big.go      |  17 +++
- src/crypto/internal/backend/common.go        |  78 ++++++++++++
+ src/crypto/internal/backend/common.go        |  78 +++++++++++
  src/crypto/internal/backend/isrequirefips.go |   9 ++
- src/crypto/internal/backend/nobackend.go     | 127 +++++++++++++++++++
+ src/crypto/internal/backend/nobackend.go     | 135 +++++++++++++++++++
  src/crypto/internal/backend/norequirefips.go |   9 ++
  src/crypto/internal/backend/stub.s           |  10 ++
  src/crypto/rand/rand_unix.go                 |   2 +-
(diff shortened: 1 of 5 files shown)


patches/0004-Add-OpenSSL-crypto-backend.patch
```diff
@@ -226,26 +226,36 @@ index 00000000000000..9995444d1f1e9f
...
 +var knownVersions = [...]string{"3", "1.1", "11", "111", "1.0.2", "1.0.0", "10"}
 +
++const lcryptoPrefix = "libcrypto.so."
++
 +func init() {
-+	version, _ := syscall.Getenv("GO_OPENSSL_VERSION_OVERRIDE")
-+	if version == "" {
-+		var fallbackVersion string
++	var lcrypto string
++	if version, _ := syscall.Getenv("GO_OPENSSL_VERSION_OVERRIDE"); version != "" {
```
qmuntal, line 233: I had to update this code because `openssl.Init` now accepts a file path instead of a version suffix.

patches/0002-Add-crypto-backend-foundation.patch
```diff
@@ -702,6 +713,191 @@ index 589e8b6fafbba3..0a6d665ee3096d 100644
...
+ }
+ 
+ // prf12 implements the TLS 1.2 pseudo-random function, as defined in RFC 5246, Section 5.
+-func prf12(hashFunc func() hash.Hash) func(result, secret, label, seed []byte) {
+-	return func(result, secret, label, seed []byte) {
++func prf12(hashFunc func() hash.Hash) func(result, secret, label, seed []byte) error {
++	return func(result, secret, label, seed []byte) error {
++		if boring.Enabled && boring.SupportsTLS1PRF() {
++			// No need to check if hashFunc is supported by the backend,
++			// since it can only by crypto.SHA256 or crypto.SHA384,
```
```suggestion
+			// since it can only be crypto.SHA256 or crypto.SHA384,
```

dagood (approved):

qmuntal:
Submitted an OpenSSL PR to fix the CI failures: https://github.com/golang-fips/openssl/pull/121.

patches/0002-Add-crypto-backend-foundation.patch
```diff
@@ -741,8 +877,218 @@ index d7f082c9ee1e04..14a85fbf1bd465 100644
...
++func prfAndHashForVersion(version uint16, suite *cipherSuite) (func(result, secret, label, seed []byte) error, crypto.Hash) {
+ 	switch version {
+ 	case VersionTLS10, VersionTLS11:
+ 		return prf10, crypto.Hash(0)
+ 	case VersionTLS12:
+ 		if suite.flags&suiteSHA384 != 0 {
+-			return prf12(sha512.New384), crypto.SHA384
++			h := crypto.SHA384
++			return prf12(h, sha512.New384), h
+ 		}
```
dagood, line 948: Why is `h` pulled out rather than passing directly?

dagood (approved):

patches/0002-Add-crypto-backend-foundation.patch
qmuntal: To avoid having to duplicate the identifier, like in here:

```go
return prf12(crypto.SHA384, sha512.New384), crypto.SHA384
```

dagood: Ha, I somehow missed that `h` was being used twice. 👍

gdams (approved):

PR #1052, opened 2023-09-28
Implement support for crypto/md5 using OpenSSL/CNG

qmuntal:
This PR patches crypto/md5 so it is implemented using the OpenSSL and CNG backends.


patches/0005-Add-CNG-crypto-backend.patch
```diff
@@ -483,6 +462,28 @@ index 4e629a4db8f7c7..a7cd24a0d15647 100644
...
+ 	"io"
+ 	"testing"
+ 	"unsafe"
+@@ -87,6 +88,9 @@ func TestGolden(t *testing.T) {
+ }
+ 
+ func TestGoldenMarshal(t *testing.T) {
++	if goexperiment.CNGCrypto {
++		t.Skip("CNGCrypto does not support hash marshalling")
++	}
```
dagood, line 499: Needed for `TestLargeHashes` too (causing test failures in latest run).

dagood (approved):

gdams (approved):

PR #1068, opened 2023-10-16
Allow TLS 1.3 when crypto/tls/fipsonly is imported

qmuntal:
This PR allows to use clients and server to use TLS 1.3 when `crypto/tls/fipsonly` is imported.

The major change is that `tls.TLS_CHACHA20_POLY1305_SHA256` won't appear in the supported TLS 1.3 cipher suites, as it is not approved by FIPS 140-3.

### DIFF
--- a/patches/0010-allow-TLS-1.3-when-crypto-tls-fipsonly-is-imported.patch
+++ b/patches/0010-allow-TLS-1.3-when-crypto-tls-fipsonly-is-imported.patch
@@ -1,0 +1,128 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Quim Muntal <qmuntaldiaz@microsoft.com>
+Date: Mon, 16 Oct 2023 12:04:32 +0000
+Subject: [PATCH] allow TLS 1.3 when crypto/tls/fipsonly is imported
+
+---
+ src/crypto/tls/boring.go                 | 2 +-
+ src/crypto/tls/boring_test.go            | 8 +++++---
+ src/crypto/tls/cipher_suites.go          | 5 +++++
+ src/crypto/tls/handshake_client.go       | 4 +++-
+ src/crypto/tls/handshake_client_tls13.go | 4 ----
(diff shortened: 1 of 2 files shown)


gdams (approved):

patches/0010-allow-TLS-1.3-when-crypto-tls-fipsonly-is-imported.patch
```diff
@@ -0,0 +1,128 @@
...
+ 
+ func isBoringCipherSuite(id uint16) bool {
+@@ -66,7 +66,9 @@ func isBoringCipherSuite(id uint16) bool {
+ 		TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
+ 		TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,
+ 		TLS_RSA_WITH_AES_128_GCM_SHA256,
+-		TLS_RSA_WITH_AES_256_GCM_SHA384:
++		TLS_RSA_WITH_AES_256_GCM_SHA384,
++		TLS_AES_128_GCM_SHA256,
++		TLS_AES_256_GCM_SHA384:
```
dagood, line 57: Seeing TLS 1.3 suites bunched up with 1.2 suites, I wonder if it would make the tests more specific to separate out when 1.2 vs. 1.3 suites are expected. However: I don't know if this would actually catch anything, and it could be a significant chunk of work that never detects any problem and simply increases conflicts vs. upstream. I think we're better off waiting for upstream to reimplement this test with 1.3 in mind.

dagood (approved):

patches/0010-allow-TLS-1.3-when-crypto-tls-fipsonly-is-imported.patch
qmuntal: Good point. I'll split it, this way `TestBoringServerCipherSuites` will also test that we are not returning TLS 1.3 ciphers in TLS 1.2 connections.

PR #1080, opened 2023-11-08
Update FIPS README.md: use new syntax for notes

dagood:
I saw this new syntax being used elsewhere and thought it might be interesting to try it out here.

https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts

> > [!NOTE]
> > Highlights information that users should take into account, even when skimming.
> 
> > [!IMPORTANT]
> > Crucial information necessary for users to succeed.
> 
> > [!WARNING]
> > Critical content demanding immediate user attention due to potential risks.

> [!NOTE]
> Highlights information that users should take into account, even when skimming.

> [!IMPORTANT]
> Crucial information necessary for users to succeed.

> [!WARNING]
> Critical content demanding immediate user attention due to potential risks.

On the other hand, I think I may have overused the idea of adding notes when I was writing these sections... this feature is making it more obvious.


gdams (approved):
Seems like a good formatting improvement

Issue #1114, opened 2024-01-22
Re-enable test output conversion after fixing truncated test results


dagood:
The JSONification and parsing of test data also cut off a significant amount of important info in a recent upstream sync PR failure, making it harder to diagnose:

![image](https://github.com/microsoft/go/assets/12819531/ee4d4b0f-d67a-4247-bcca-ad7bb848cd23)

Generating these is why verbose mode is enabled, and if it doesn't work in the first place, there's even less reason to keep test results verbose.

For this issue in particular I suspect there's some underlying issue with the log output or parsing--the order, individual lines being misattributed (due to parallelism?), etc.

_Originally posted by @dagood in https://github.com/microsoft/go/issues/1049#issuecomment-1783329906_

PR #1115, opened 2024-01-22
Improve Windows builder retry diagnosability

### LINKED ISSUE #1049: Investigate use of verbose mode causing slower test runs in CI
@qmuntal noticed the Windows jobs in CI are slower than they could be because verbose mode is disabling parallel runs for a particular test.

I seem to remember that implementing the JSON mode and conversion to the AzDO test schema is related to verbose mode in some way, but it doesn't make sense for this to be required.

### LINKED ISSUE #1060: Change Windows CI auto-retry to improve UX
In the last release, there were a couple issues I hit:

* Test logs weren't uploaded when a timeout occurred.
* When a job is running, attempting to scroll up in the currently-running step doesn't work because every few seconds, the general job-init log is displayed and then it returns to showing the running log. Scroll, search, etc. is reset. (I believe this is a point-in-time bug, but this kind of behavior has occurred in other situations to me before.) A workaround is to cancel the job, but at this point in the job, I don't know for sure whether the job is simply slow, or the job has actually attempted the tests several times due to a non-flaky failure. I don't want to waste time by canceling a job that may succeed.

---

The first thing that comes to mind is reimplementing the retries as a sequence of pipeline steps. We do 5 retries right now, so we could have 5 "build" steps that each tries to build the repo if it hasn't been successful yet, or skip if it has worked.

Similarly, 5 "test" steps would run, and a "upload results" step could be added after each test step. This also would give us results sooner than we currently get them, to make investigation easier to do early.

A "complete" pipeline step has more stable logs, the ability to download raw logs, etc., so splitting up steps could generally help quite a bit.

This may also make it easier to set up tooling that keeps track of how much retrying is being done in our builds, by elevating it to the AzDO API level rather than needing to parse logs.

dagood:
* Switch to pipeline logic for test retries. Makes it easier to find the failure and see if retries were necessary.
  * Fixes https://github.com/microsoft/go/issues/1060
* Remove junit test output conversion: this seemed to cut off test output in some cases.
  * Fixes https://github.com/microsoft/go/issues/1049
  * Opened https://github.com/microsoft/go/issues/1114 to get this (or something like it) working again in the future.

### DIFF
--- a/eng/_core/README.md
+++ b/eng/_core/README.md
@@ -18,10 +18,13 @@ The high-level execution flow looks roughly like this when running in CI:
 
 * `eng/pipeline/jobs/run-stage.yml`  
   runs:
-* `eng/run.ps1 run-builder -builder linux-amd64-test -junitfile [...]`  
+* `eng/run.ps1 run-builder -test -builder linux-amd64-test -junitfile [...]`  
   which runs the Go function:
 * `gotestsum.Run(... eng/run.ps1 build -test -json ...)`  
   which runs and captures the output of:
 * `eng/run.ps1 build -test -json`  
   which runs [`cmd/build/build.go`](cmd/build/build.go) in this module.
 
+This is not currently used in our CI because this process seems to cut off
+some test output:
+[microsoft/go#1114](https://github.com/microsoft/go/issues/1114).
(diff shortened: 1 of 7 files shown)


qmuntal (approved):
LGTM!

gdams (approved):

Issue #1107, opened 2024-01-11
Add crypto/tls/fipsonly functionality that doesn't require source changes


dagood:
A while ago, we decided not to enable `crypto/tls/fipsonly` by default: https://github.com/microsoft/go/issues/277#issuecomment-974245122.

Adding `import _ "crypto/tls/fipsonly"` isn't hard, but our internal partners are trying to rebuild apps with minimized source changes. Implementing a tag that enables a small source file with the import would help.

dagood:
* https://github.com/microsoft/go/pull/1116 isn't enough, because it produces a binary that is always limited to FIPS TLS. To make our toolset produce compatible binaries, we need make it possible to decide at runtime. Tying this to FIPS mode (GOFIPS) seems like a good idea: https://github.com/microsoft/go/pull/1127.

dagood:
Done in 1.22.0. https://github.com/microsoft/go/blob/microsoft/release-branch.go1.22/docs/go1.22.md#automatically-configure-tls-settings-in-fips-mode

Issue #999, opened 2023-08-02
Query from application whether crypto backend is in FIPS mode


corhere:
Applications can use the `"crypto/boring".Enabled()` function to query whether or not a crypto backend is handling cryptographic operations. `boring.Enabled() == true` implies that the BoringCrypto backend is operating in FIPS mode as [that backend unconditionally forces FIPS mode enabled](https://github.com/golang/go/blob/a09ea59198faf85d41da6699220319089c98d86b/src/crypto/internal/boring/boring.go#L25-L31) but the same is not true for the OpenSSL and CNG backends unless the application is built with `-tags requirefips`. There is currently no way for an application to accurately determine the FIPS-mode status of the backend without also requiring that FIPS mode be enabled.

The crypto modules' FIPS-mode status is made available to the crypto backend internals through their respective bindings ([`openssl.FIPS()`](https://pkg.go.dev/github.com/golang-fips/openssl/v2#FIPS), [`cng.FIPS()`](https://pkg.go.dev/github.com/microsoft/go-crypto-winnative@v0.0.0-20230502061212-6eb98854418e/cng?GOOS=windows#FIPS)) but there is no safe way for the application to access that information. The application can't `import "github.com/golang-fips/openssl/v2"` to call `openssl.FIPS()` directly [as it would import a separate copy with a different package path which shares no state with the module vendored into the runtime](https://github.com/golang/go/blob/ce5e37ec21442c6eb13a43e68ca20129102ebac0/src/README.vendor#L7-L16) and therefore does not have access to the OpenSSL library handle in use by the crypto backend. Also, the application would fail to build with "multiple definition" link errors from the two conflicting copies of the cgo symbols. There is a hacky way to let the application interrogate the crypto modules' bindings directly, but it is unsafe and brittle:
```go
import "unsafe"

//go:linkname FIPS vendor/github.com/golang-fips/openssl/v2.FIPS
func FIPS() bool
```

-----

I propose that, when `GOEXPERIMENT=*crypto`, the following API is made available to applications:
```go
package boring // import "crypto/boring"

// FIPS returns whether the cryptographic backend is operating in FIPS mode.
func FIPS() bool
```

I will gladly open a PR to implement this if there is interest.

dagood:
What would this API be used for in practice?

[...]

PR #1116, opened 2024-01-24
Add `ms.tlsfipsonly` tag to reduce source changes

dagood:
* For https://github.com/microsoft/go/issues/1107

This is the first `ms.` prefix, but I've been thinking we should use one for a while: https://github.com/microsoft/go/issues/999#issuecomment-1664742297. If upstream added a `tlsfipsonly` tag, it may work differently, and this helps us be prepared for that.

For now I wrote that it's in 1.23 since that's what `main` is right now, but we might want to put it in 1.22.

This is a very simple feature--a lot of this PR is for the error checking and writing up the usage in the FIPS doc.

Doc preview: https://github.com/microsoft/go/blob/dev/dagood/fipsonly-tag/eng/doc/fips/README.md


gdams (approved):

qmuntal (approved):

dagood:
After some more internal discussion, we don't think this is enough to actually address our needs.

* New approach: https://github.com/microsoft/go/pull/1127

PR #1117, opened 2024-01-24
Add crypto/md5 to user guide

qmuntal:
`crypto/md5` was added in #1052 but didn't update the user guide. Do it now.

### DIFF
--- a/eng/doc/fips/UserGuide.md
+++ b/eng/doc/fips/UserGuide.md
@@ -869,6 +869,8 @@ The hash.Hash methods are implemented as follows:
 
 Package md5 implements the MD5 hash algorithm as defined in RFC 1321.
 
+MD5 is cryptographically broken and should not be used for secure applications.
+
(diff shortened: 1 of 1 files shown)


eng/doc/fips/UserGuide.md
```diff
@@ -859,7 +867,50 @@ The hash.Hash methods are implemented as follows:
 
 ### [crypto/md5](https://pkg.go.dev/crypto/md5)
 
-Not implemented by any backend.
+Package md5 implements the MD5 hash algorithm as defined in RFC 1321.
```
dagood, line 870: Add a note similar to the one about sha1 (but not approved at all for any security purposes)?

> SHA-1 is an approved FIPS 140-2 hash algorithm although it is cryptographically broken and should not be used for secure applications.

dagood (approved):

gdams (approved):

Issue #1119, opened 2024-01-27
Fix patches after "crypto: rollback BoringCrypto fips-20220613 update"


dagood:
Once this goes into upstream:

* https://github.com/golang/go/issues/65321

We'll probably hit patch or test conflicts in microsoft/main and 1.22, 1.21, and 1.20. Essentially, we need to revert parts of:

* https://github.com/microsoft/go/pull/1097
* https://github.com/microsoft/go/pull/1110
* https://github.com/microsoft/go/pull/1104
* https://github.com/microsoft/go/pull/1103

Once the cherry-picks go through to the release branches, we should proactively merge to latest to get past the conflict, so the next release can go smoothly on each branch.

gdams:
Quim says this is done

PR #1121, opened 2024-01-30
Support TLS 1.3 in fipstls mode

qmuntal:
This PR adds support for TLS 1.3 when running in fipstls mode.

The changes have been taken from https://go-review.googlesource.com/c/go/+/558796 and put in its own patch file so they are easier to revert once upstream finally support TLS 1.3 in fipstls mode.

This will have to be ported to the go1.22 release branch.

Updates #1119.

### DIFF
--- a/patches/0010-Support-TLS-1.3-in-fipstls-mode.patch
+++ b/patches/0010-Support-TLS-1.3-in-fipstls-mode.patch
@@ -1,0 +1,353 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: qmuntal <qmuntaldiaz@microsoft.com>
+Date: Tue, 30 Jan 2024 11:40:31 +0100
+Subject: [PATCH] Support TLS 1.3 in fipstls mode
+
+---
+ src/crypto/tls/boring.go                 | 14 +++--
+ src/crypto/tls/boring_test.go            | 66 ++++++++++++++++++------
+ src/crypto/tls/cipher_suites.go          | 10 ++--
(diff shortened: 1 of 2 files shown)


patches/0010-Support-TLS-1.3-in-fipstls-mode.patch
```diff
@@ -0,0 +1,358 @@
...
+ 
+ // fipsMaxVersion replaces c.maxVersion in FIPS-only mode.
+ func fipsMaxVersion(c *Config) uint16 {
+-	// FIPS requires TLS 1.2.
+-	return VersionTLS12
++	// FIPS requires TLS 1.2 or TLS 1.3.
++	return VersionTLS13
+ }
+ 
+ // default defaultFIPSCurvePreferences is the FIPS-allowed curves,
```
dagood, line 38: In the var here, `var defaultFIPSCurvePreferences = []CurveID{CurveP256, CurveP384, CurveP521}`, the `CurveP521` entry is still there--this is good, per https://github.com/golang/go/issues/62372#issuecomment-1898746485.

(Just taking a note on what's different here vs. the reverted upstream changes.)

dagood (approved):
Nice! I like this approach, keeping our implementation intact rather than simply following upstream and removing it altogether.

dagood:
> This will have to be ported to the go1.22 release branch.

Also go1.20/go1.21, right? (Although it's only practical to actually do it once https://github.com/golang/go/issues/65322 and https://github.com/golang/go/issues/65323 get merged.)

gdams (approved):

PR #1127, opened 2024-02-05
Enable fipstls when running in FIPS mode

qmuntal:
Automatically enforces that `crypto/tls` and `crypto/x509` only use FIPS-approved settings when running in FIPS mode. This differs from upstream's BoringCrypto backend, which requires you to import `crypto/tls/fipsonly` to apply the FIPS-mandated restrictions. 

Do this automatically to reduce the source code changes necessary to produce a FIPS-compliant Go application.

Code taken from #964.

### DIFF
--- a/eng/doc/fips/README.md
+++ b/eng/doc/fips/README.md
@@ -17,7 +17,7 @@ The Go `crypto` package is not FIPS certified, and the Go team has stated that i
 
 On the other hand, Google maintains the [goexperiment](https://pkg.go.dev/internal/goexperiment) `boringcrypto`, that uses cgo and BoringSSL to implement various crypto primitives. As BoringSSL is FIPS 140-2 certified, an application built using this flag is more likely to be FIPS 140-2 compliant, yet Google does not provide any liability about the suitability of this code in relation to the FIPS 140-2 standard.
 
-In addition to that, the boringcrypto flag also provides a mechanism to restrict all TLS configuration to FIPS-approved settings. The effect is triggered by importing the fipsonly package anywhere in a program, as in:
+In addition to that, the boringcrypto flag also provides a mechanism to restrict all TLS configuration to FIPS-compliant settings. The effect is triggered by importing the fipsonly package anywhere in a program, as in:
 
 ```go
   import _ "crypto/tls/fipsonly"
@@ -40,7 +40,7 @@ The Microsoft Go fork provides several ways to configure the crypto backend and 
   - [`goexperiment.<backend>crypto` build tag](#usage-build)
   - [`requirefips` build tag](#build-option-to-require-fips-mode)
   - [`GOEXPERIMENT` `allowcryptofallback`](#build-option-to-use-go-crypto-if-the-backend-compatibility-check-fails)
-  - [`import _ "crypto/tls/fipsonly"` source change](#tls-with-fips-approved-settings)
+  - [`import _ "crypto/tls/fipsonly"` source change](#tls-with-fips-compliant-settings)
 - Runtime configuration:
   - [`GOFIPS` environment variable](#usage-runtime)
   - (OpenSSL backend) [`GO_OPENSSL_VERSION_OVERRIDE` environment variable](#runtime-openssl-version-override)
@@ -67,7 +67,7 @@ There are typically two goals that lead to this document. Creating a FIPS compli
 
 Other notes for common configurations:
 
-- If the app uses TLS, `import _ "crypto/tls/fipsonly"` is also necessary for FIPS compliance. See [TLS with FIPS-approved settings](#tls-with-fips-approved-settings)
+- Prior to Go 1.22, if the app uses TLS, `import _ "crypto/tls/fipsonly"` is also necessary for FIPS compliance. See [TLS with FIPS-compliant settings](#tls-with-fips-compliant-settings)
 - A Docker image is available that includes suitable build-time config in the environment. See [Dockerfile base image](#dockerfile-base-image)
 
 Some configurations are invalid and intentionally result in a build error or runtime panic:
@@ -358,17 +358,17 @@ The OpenSSL version present when building a program does not have to match the O
 
 This feature does not require any additional configuration, but it only works with OpenSSL versions known and supported by the Go toolchain.
 
-### TLS with FIPS-approved settings
+### TLS with FIPS-compliant settings
 
-The Go TLS stack will automatically use OpenSSL crypto primitives when running in FIPS mode. Yet, the FIPS 140-2 standard places additional restrictions on TLS communications, mainly on which cyphers and signers are allowed.
+The Go TLS stack will automatically use crypto primitives from the selected crypto backend. Yet, this isn't enough for FIPS compliance: the FIPS 140-2 standard places additional restrictions on TLS communications, mainly on which cyphers and signers are allowed. Note that this can reduce compatibility with old devices that do not support modern cryptography techniques such as TLS 1.2.
 
-A program can import the `crypto/tls/fipsonly` package to configure the Go TLS stack so it is compliant with these restrictions. The configuration is done by an `init()` function, so only importing it is necessary:
+Since Go 1.22, the Microsoft Go runtime automatically enforces that `crypto/tls` and `crypto/x509` only use FIPS-compliant settings when running in FIPS mode. This differs from upstream's BoringCrypto backend, which requires you to import `crypto/tls/fipsonly` to apply the FIPS-mandated restrictions. The Microsoft Go crypto backends do this automatically to reduce the source code changes necessary to produce a FIPS-compliant Go application, and to make it easier to use the same binary in both FIPS and non-FIPS environments.
+
+Prior to Go 1.22, a program using the Go TLS stack must import the `crypto/tls/fipsonly` package to be compliant with these restrictions. The configuration is done by an `init()` function, so only importing it is necessary:
 
 ```go
   import _ "crypto/tls/fipsonly"
 ```
-
-Note that this can reduce compatibility with old devices that do not support modern cryptography techniques such as TLS 1.2.
 
 ## Acknowledgements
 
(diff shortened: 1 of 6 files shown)


eng/doc/fips/README.md
```diff
@@ -360,9 +360,11 @@ This feature does not require any additional configuration, but it only works wi
...
+The Go TLS stack will automatically use OpenSSL crypto primitives. Yet, the FIPS 140-2 standard places additional restrictions on TLS communications, mainly on which cyphers and signers are allowed.
 
-A program can import the `crypto/tls/fipsonly` package to configure the Go TLS stack so it is compliant with these restrictions. The configuration is done by an `init()` function, so only importing it is necessary:
+Since Go 1.22, the Microsoft Go runtime automatically enforces that `crypto/tls` and `crypto/x509` only use FIPS-approved settings when running in FIPS mode. This differs from upstream's BoringCrypto backend, which requires you to import `crypto/tls/fipsonly` to apply the FIPS-mandated restrictions. The Microsoft Go crypto backends do this automatically to reduce the source code changes necessary to produce a FIPS-compliant Go application.
+
+Prior to Go 1.22, a program using the Go TLS stack must import the `crypto/tls/fipsonly` package to be compliant with these restrictions. The configuration is done by an `init()` function, so only importing it is necessary:
```
dagood, line 367: Add a line in the changelog?

eng/doc/fips/README.md
```diff
@@ -360,9 +360,11 @@ This feature does not require any additional configuration, but it only works wi
 
 ### TLS with FIPS-approved settings
 
-The Go TLS stack will automatically use OpenSSL crypto primitives when running in FIPS mode. Yet, the FIPS 140-2 standard places additional restrictions on TLS communications, mainly on which cyphers and signers are allowed.
+The Go TLS stack will automatically use OpenSSL crypto primitives. Yet, the FIPS 140-2 standard places additional restrictions on TLS communications, mainly on which cyphers and signers are allowed.
```
dagood: Good catch on the old FIPS mode clause. 😄 

I think OpenSSL can be expanded to any backend and the "yet" can be a little more direct while we're here:

```suggestion
The Go TLS stack will automatically use crypto primitives from the selected crypto backend. Yet, this isn't enough for FIPS compliance: the FIPS 140-2 standard places additional restrictions on TLS communications, mainly on which cyphers and signers are allowed.
```

eng/doc/fips/README.md
```diff
@@ -360,9 +360,11 @@ This feature does not require any additional configuration, but it only works wi
...
 
-A program can import the `crypto/tls/fipsonly` package to configure the Go TLS stack so it is compliant with these restrictions. The configuration is done by an `init()` function, so only importing it is necessary:
+Since Go 1.22, the Microsoft Go runtime automatically enforces that `crypto/tls` and `crypto/x509` only use FIPS-approved settings when running in FIPS mode. This differs from upstream's BoringCrypto backend, which requires you to import `crypto/tls/fipsonly` to apply the FIPS-mandated restrictions. The Microsoft Go crypto backends do this automatically to reduce the source code changes necessary to produce a FIPS-compliant Go application.
```
dagood: > The Microsoft Go crypto backends do this automatically to reduce the source code changes necessary to produce a FIPS-compliant Go application.

Hmm. Probably worth mentioning that this makes a single Go binary able to run with/without FIPS-restricted TLS, too.

But I wonder if this concept in general is worth its own new Features section, e.g. first we have "No code changes required" (which is arguably truer once we have this feature 😄), then "Build one binary for FIPS and non-FIPS" or something like htat. I take this for granted now, but for teams that are poking through our docs to figure out if it works for their scenario, I think this is worth highlighting in itself.

dagood: I think it's worth adding something in the intro to clarify that even though boringcrypto provides `import _ "crypto/tls/fipsonly"` and you'd need to add that to your project, you don't need to with Microsoft Go.

(We could potentially remove that intro paragraph entirely, although I think we need to be careful that people using Go <1.22 still easily see/understand that they need the import.)

dagood (commented):
Took a look at the doc changes to start with.

eng/doc/fips/README.md
qmuntal: I've updated the `Usage: Common configurations` section to mention this change too.

qmuntal: > Hmm. Probably worth mentioning that this makes a single Go binary able to run with/without FIPS-restricted TLS, too.

Done!

> But I wonder if this concept in general is worth its own new Features section...

IMO this is why we have `Usage: Common configurations`, and in there we already explain that FIPS or no-FIPS is a runtime decision unless `-tags=requirefips` is set.

dagood: Yeah, reading that section again, I agree it's enough.

eng/doc/fips/README.md
```diff
@@ -358,11 +358,13 @@ The OpenSSL version present when building a program does not have to match the O
 
 This feature does not require any additional configuration, but it only works with OpenSSL versions known and supported by the Go toolchain.
 
-### TLS with FIPS-approved settings
+### TLS with FIPS-compliant settings
 
-The Go TLS stack will automatically use OpenSSL crypto primitives when running in FIPS mode. Yet, the FIPS 140-2 standard places additional restrictions on TLS communications, mainly on which cyphers and signers are allowed.
+The Go TLS stack will automatically use crypto primitives from the selected crypto backend. Yet, this isn't enough for FIPS compliance: the FIPS 140-2 standard places additional restrictions on TLS communications, mainly on which cyphers and signers are allowed.
```
dagood: Sorry, one more thing in the doc. 😄 I think the `compatibility with old devices` note should be moved below this paragraph to make it 100% clear that the FIPS restrictions themselves cause the compatibility issue, not our approach to it. (And definitely not just the `import` approach!)

patches/0002-Add-crypto-backend-foundation.patch
```diff
@@ -1527,26 +1547,164 @@ index 8233985a62bd22..f46d4636557714 100644
...
+diff --git a/src/net/smtp/smtp_test.go b/src/net/smtp/smtp_test.go
+index 259b10b93d9e36..2ee46483954fcb 100644
+--- a/src/net/smtp/smtp_test.go
++++ b/src/net/smtp/smtp_test.go
+@@ -1105,40 +1105,59 @@ func sendMail(hostPort string) error {
+ 
+ // localhostCert is a PEM-encoded TLS cert generated from src/crypto/tls:
+ //
+-//	go run generate_cert.go --rsa-bits 1024 --host 127.0.0.1,::1,example.com \
++//	go run generate_cert.go --rsa-bits 2048 --host 127.0.0.1,::1,example.com \
```
dagood, line 1628: I think this deserves its own comment too--I assume it's the same as the `client_test.go` change. For now the patches are adjacent, but might lose the context if we need to change another file later that sorts between `client_test.go` and `smtp_test.go`.

patches/0004-Add-OpenSSL-crypto-backend.patch
```diff
@@ -281,6 +282,10 @@ index 00000000000000..d929b3e516bdb0
 +			}
 +		}
 +	}
++	if openssl.FIPS() {
```
dagood, line 285: Maybe call this once before the `switch` to avoid potentially calling it twice?

patches/0004-Add-OpenSSL-crypto-backend.patch
```diff
@@ -281,6 +282,10 @@ index 00000000000000..d929b3e516bdb0
...
++		fipstls.Force()
++		sig.FIPSOnly()
```
dagood: This could be misleading... since `openssl.FIPS()` isn't a const, my understanding is this will always be linked in whether FIPS mode is on or off, then a static scanner (coded with boringcrypto in mind) would claim that the whole binary uses FIPS-only TLS.

The sig file's comment is pretty specific:

```
// FIPSOnly indicates that package crypto/tls/fipsonly is present.
func FIPSOnly()
```

However I do think that in spirit, linking in `sig.FIPSOnly()` makes sense.

As far as the clarity of the patch goes, I think it's worth a comment to clarify the intent and move it out of the `if` so there can't be any misimpression that it'll actually be conditionally included.

That all said, https://pkg.go.dev/rsc.io/goversion already doesn't seem to work on our binaries (at least on Windows?): `... X:cngcrypto (standard crypto)`. So, actual interactions with scanners is probably something to figure out later. The `go tool nm foo.exe` output looks correct to me, including both BoringCrypto and FIPSOnly.

patches/0005-Add-CNG-crypto-backend.patch
```diff
@@ -207,6 +208,12 @@ index 00000000000000..3d05fa70409666
 +			panic("cngcrypto: not in FIPS mode")
 +		}
 +	}
++	if enabled, _ := cng.FIPS(); enabled {
```
dagood, line 211: (I think the comments from the OpenSSL side also apply here.)

dagood (approved):
This is great. 🙂 I don't think any of my comments are worth blocking for or have deep ramifications.

gdams (approved):

eng/doc/fips/README.md
qmuntal: Good point. Thanks!

patches/0004-Add-OpenSSL-crypto-backend.patch
qmuntal: Hmm, interesting. Given that we can't really make `sig.FIPSOnly` work correctly, I would rather not add it. If someone is using an scanner looking for `sig.FIPSOnly` as a matter of compliance, I prefer to have a false negative (aka binary reported as not compliant when it is compliant) that can be fixed with an exception rather than a false positive (aka binary reported as compliant when it is not compliant).

patches/0004-Add-OpenSSL-crypto-backend.patch
qmuntal: The code in the `switch` potentially call `openssl.SetFIPS`, so it can't be directly reused. We could add some logic to workaround that, but this code is already quite tricky and I would prefer not to add more complexity just to save a couple of cgo calls that are only ever called once.

PR #1138, opened 2024-02-08
Add crypto/des section to FIPS user guide

dagood:
Noticed that our 1.22 release notes point at this doc for this API, but it wasn't documented yet.

Didn't add **Implementation** sections in the initial PR. I know where to dig around and find the info, but I think @qmuntal could do it a lot faster with familiarity. 😄


qmuntal:
> I think @qmuntal could do it a lot faster with familiarity. 😄

Done!

PR #1141, opened 2024-02-14
Support executing the Go toolchain with GOFIPS set

### LINKED ISSUE #1106: Improve behavior of `GOFIPS=1 GOEXPERIMENT=systemcrypto go run .`
Currently, with a simple hello world program, we get this result:

```
$ GOFIPS=1 GOEXPERIMENT=systemcrypto ~git/go/go/bin/go run .
panic: FIPS mode requested (environment variable GOFIPS=1) but no supported crypto backend is enabled

goroutine 1 [running]:
crypto/internal/backend.init.0()
        /home/dagood/git/go/go/src/crypto/internal/backend/common.go:21 +0x65
```

The reason for this panic that the microsoft/go artifacts are not built with a backend. When `go` runs, it finds `GOFIPS=1` and panics immediately. This is confusing in a few ways:

* With `go run`, it isn't really clear from the error where the panic happened (in `go` or in the running app), which makes it harder to figure out.
* The panic message seems to directly contradict `GOEXPERIMENT=systemcrypto` unless you already know what's going on. It seems like `GOEXPERIMENT=systemcrypto` is being ignored, but really the panic is simply unrelated.

In a script, the cause might not be as direct:

```sh
export GOFIPS=1 # Intent: run the Go app in FIPS mode. I don't care about the build.

GOEXPERIMENT=systemcrypto go build .
./app
```

---

I think having the toolset binaries behave differently to make this situation clearer would be useful. If we can, though, we should try to avoid ending up in a situation where people seemingly-successfully use `GOFIPS=1` when they intended to use `-tags=requirefips`.

I don't have this totally thought out yet--filing this to track.

qmuntal:
It is currently not possible to execute the Go toolchain with most combinations of `GOFIPS`+`GOEXPERIMENT`. For example, this command fails: `GOFIPS=1 GOEXPERIMENT=systemcrypto go test ./...`:

```cmd
$ GOFIPS=1 GOEXPERIMENT=systemcrypto ~git/go/go/bin/go run .
panic: FIPS mode requested (environment variable GOFIPS=1) but no supported crypto backend is enabled

goroutine 1 [running]:
crypto/internal/backend.init.0()
        /home/dagood/git/go/go/src/crypto/internal/backend/common.go:21 +0x65
```

The `GOFIPS` env is intended to be used in the temporary binary generated by `go test` (and `go run`), but it also affects the Go toolchain itself, as it imports `crypto/sha256`, which triggers the `GOFIPS` check even before building the binary.

This PR updates the Go toolchain initialization to unset `GOFIPS` for it's own processes, while keeping it set for the user binaries. There is no need to add new tests, as `go test` and `go run` are already heavily tested. We haven't seen any test failure in our fips builders yet because we were setting `GOFIPS=true` instead of `GOFIPS=1`, which caused the tests to ignore it.

Notice that this change doesn't preclude the Go toolchain to run in FIPS mode. When `GOFIPS` is unset, the default behavior is to try to use FIPS mode whenever possible. The only change is that the process won't panic when FIPS mode can't be enabled. If someone wants to make sure that the Go toolchain is executed in FIPS mode, then the only solution is to build it with `-tags requirefips`.

Fixes #1106.

### DIFF
--- a/eng/_util/cmd/run-builder/run-builder.go
+++ b/eng/_util/cmd/run-builder/run-builder.go
@@ -143,7 +143,7 @@ func main() {
 		}
 
 		if *fipsMode {
-			env("GOFIPS", "true")
+			env("GOFIPS", "1")
 			// Enable system-wide FIPS if supported by the host platform.
 			restore, err := enableSystemWideFIPS()
 			if err != nil {
--- a/patches/0011-unset-GOFIPS-when-running-the-Go-toolchain.patch
+++ b/patches/0011-unset-GOFIPS-when-running-the-Go-toolchain.patch
@@ -1,0 +1,72 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: qmuntal <qmuntaldiaz@microsoft.com>
+Date: Wed, 14 Feb 2024 11:03:01 +0100
+Subject: [PATCH] unset GOFIPS when running the Go toolchain
+
+---
+ src/cmd/go/internal/gofips/gofips.go | 32 ++++++++++++++++++++++++++++
+ src/cmd/go/main.go                   |  5 +++++
+ 2 files changed, 37 insertions(+)
+ create mode 100644 src/cmd/go/internal/gofips/gofips.go
+
+diff --git a/src/cmd/go/internal/gofips/gofips.go b/src/cmd/go/internal/gofips/gofips.go
+new file mode 100644
+index 00000000000000..009eece5b6c080
+--- /dev/null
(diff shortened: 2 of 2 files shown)


dagood (approved):
Looks good! At first glance the dependency on init order seems fragile, but because packages are basically initialized breadth-first, it's fine. If we mess it up in the future, the error may still be confusing, but the fixed test should at least keep it confined to the PR that breaks it.

I tried to think of a way to make the error clearer in case the ordering does somehow get mixed up, but I couldn't come up with anything because communication between `cmd/` and `crypto/` is limited (can't use each others' internal packages).

gdams (approved):

PR #1163, opened 2024-03-19
Add support for logging used Windows APIs

qmuntal:
This PR instruments the Go runtime so it does a best effort to log the Windows API used at runtime.

The instrumented functions are:

- [syscall.FindProc](https://pkg.go.dev/syscall?GOOS=windows#DLL.FindProc)
- [windows.FindProc](https://pkg.go.dev/golang.org/x/sys/windows?GOOS=windows#DLL.FindProc)
- Functions internally loaded by the Go runtime

This should cover almost all cases for applications not using cgo.

To enable this logging, one should set the `MS_GO_SYSCALL_TRACE` env var to a file path (existing or not), to which the APIs will be logged.

The output format is `<processName>,<moduleName>,<functionName>\n`. Example:

```
gotest.exe,C:\WINDOWS\System32\KERNEL32.DLL,LoadLibraryExW
gotest.exe,C:\WINDOWS\System32\KERNEL32.DLL,LoadLibraryW
gotest.exe,C:\WINDOWS\System32\bcryptprimitives.dll,ProcessPrng
```

Note that the output file can contain entries for different processes in case the main processes run child Go processes.

There are some limitations in the implementation:

- Doesn't support file paths longer than MAX_PATH.
- Only supports ANSI encoding in `<processName>` and `<moduleName>`

This is because the logging function is executed deep inside the Go runtime, possibly in a stage that doesn't allow memory allocations.


patches/0012-add-support-for-logging-used-Windows-APIs.patch
```diff
@@ -0,0 +1,372 @@
...
++		// because the first time syscallTrace is called is during
++		// runtime initialization, when there is only one goroutine.
++		if !initSyscallTrace() {
++			return
++		}
++	}
++	// We can't allocate memory here, as syscallTrace might be called
++	// before the memory allocator is initialized. Instead, we use a
++	// fixed-size buffer on the stack.
++	buf := make([]byte, _MAX_PATH*3+3)
```
dagood, line 146: I believe `*3+3` is because there are three entries on the line, two commas, and one `\n`, but maybe worth mentioning?

patches/0012-add-support-for-logging-used-Windows-APIs.patch
```diff
@@ -0,0 +1,372 @@
...
++	if n1 == 0 || n1 == _MAX_PATH {
++		// Failed or short buffer.
++		buf[n] = '?'
++		n += 1
++	} else {
++		n += int(n1)
++	}
++	buf[n] = ','
++	n += 1
++	procNameLength := findnull(procName)
```
dagood, line 163: Is `procName` (always?) null terminated? Based on `syscall_getprocaddress` accepting `*byte` it certainly seems like it, but figured it might be worth asking.

patches/0012-add-support-for-logging-used-Windows-APIs.patch
```diff
@@ -0,0 +1,372 @@
...
++
++var (
++	_CreateFileW,
++	_GetEnvironmentVariableW,
++	_GetModuleFileNameA,
++	_ stdFunction
++)
++
++var (
++	syscallTraceHandler uintptr
```
dagood: This seems like a weird name for a file handle, but `syscallTraceHandle` would make sense. Typo? Or `syscallTraceFileHandle` could be even clearer.

dagood (approved):
Nice! Haven't tried it locally, but it makes sense.

gdams (approved):

patches/0012-add-support-for-logging-used-Windows-APIs.patch
qmuntal: Type, and I like your proposal. Thanks!

patches/0012-add-support-for-logging-used-Windows-APIs.patch
qmuntal: Yep, always, else we would have to pass the string length around. Anyway, I'm adding a comment on the function docs.

PR #1178, opened 2024-03-25
Specify that only certain OpenSSL configurations are supported

qmuntal:
We are already stating that only certain OpenSSL versions are supported/tested. We should go a little bit further and clarify that only some configurations of those versions are really supported.

### DIFF
--- a/eng/doc/fips/README.md
+++ b/eng/doc/fips/README.md
@@ -342,6 +342,8 @@ Not all OpenSSL versions are supported. OpenSSL does not maintain ABI compatibil
 
 Versions not listed above are not supported at all.
 
+Note that one can enable or disable certain [OpenSSL features] when building it, diverging from the default configuration. The Go runtime does not support all possible configurations, and some may cause the Go runtime to panic during initialization or not work as expected. The Go runtime is tested with the default configuration of the supported versions and with the OpenSSL configuration shipped in [Azure Linux]. 
+
 ### Dynamic linking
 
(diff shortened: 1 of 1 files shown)


eng/doc/fips/README.md
```diff
@@ -342,6 +342,8 @@ Not all OpenSSL versions are supported. OpenSSL does not maintain ABI compatibil
 
 Versions not listed above are not supported at all.
 
+Note that one can enable or disable certain [OpenSSL features] when building it, diverging from the default configuration. The Go runtime does not support all possible configurations, and some may cause the Go runtime to panic during initialization or not work as expected. The Go runtime is tested with the default configuration of the supported versions and with the OpenSSL configuration shipped in [Azure Linux].
```
gdams: Guessing that we probably want to use this syntax for consitency:

```suggestion
> [!NOTE]
> Note that one can enable or disable certain [OpenSSL features] when building it, diverging from the default configuration. The Go runtime does not support all possible configurations, and some may cause the Go runtime to panic during initialization or not work as expected. The Go runtime is tested with the default configuration of the supported versions and with the OpenSSL configuration shipped in [Azure Linux]. 
```

> [!NOTE]
> Note that one can enable or disable certain [OpenSSL features] when building it, diverging from the default configuration. The Go runtime does not support all possible configurations, and some may cause the Go runtime to panic during initialization or not work as expected. The Go runtime is tested with the default configuration of the supported versions and with the OpenSSL configuration shipped in [Azure Linux].

dagood (approved):

PR #1200, opened 2024-04-10
Remove -d=dclstack compiler flag for ssacheck builder

qmuntal:
Our outerloop sshacheck builder have been consistently failing for a year (e.g. https://dev.azure.com/dnceng/internal/_build/results?buildId=2426158&view=logs&s=e08da66d-f904-5a02-d622-d38459b08878) because it sets the `-d=dclstack` argument when invoking the Go toolchain, but that argument was removed in https://go-review.googlesource.com/c/build/+/463775.

We should check outerloop builders more often...

### DIFF
--- a/eng/_util/cmd/run-builder/run-builder.go
+++ b/eng/_util/cmd/run-builder/run-builder.go
@@ -90,7 +90,7 @@ func main() {
 	case "regabi":
 		buildutil.AppendExperimentEnv("regabi")
 	case "ssacheck":
-		env("GO_GCFLAGS", "-d=ssa/check/on,dclstack")
+		env("GO_GCFLAGS", "-d=ssa/check/on")
 	case "staticlockranking":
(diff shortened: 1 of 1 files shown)


gdams (approved):

dagood (approved):
A long while ago I set up https://dev.azure.com/dnceng/internal/_dashboards/dashboard/9371d09b-76a4-41d7-b287-7ba66cd4928a to try to keep an eye on it without *loud* noise like emails. I think it suffered the usual fate of flakiness on a low priority pipeline: it wasn't worth checking and confirming every single failure was flakiness, but it also wasn't worth filtering out flakiness or turning off specific tests, so it ended up ignored.

Now that we're far enough that we're working on scenario tests, it makes sense to me that getting this part of the house cleaned up (and keeping it that way) might have enough priority.

PR #1210, opened 2024-04-26
Generate symbols for all Go toolchain binaries

qmuntal:
> [!NOTE]
> This PR uses a PDB generation tool, gopdb, which is still experimental and not available publicly.

This PR adds support for publishing symbols in the form of PDBs. It is enabled by default for the `rolling-internal` pipeline.

This is a summary of the changes:
- Download and build gopdb from the internal AzDO mirror.
- Update the build command to run the `gopdb` tool for all the Go toolchain binaries.
- Store all PDBs as pipeline artifacts
- Update the publish stage to also publish the PDB files using [PublishSymbols@2](https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/publish-symbols-v2?view=azure-pipelines). Symbols are published into the private dnceng symbol server, which is the only server that our pipeline has access to.
- To achieve the previous task, I had to move the `publish-stage.yml` execution to `builders-to-stages.yml`, as it now needs to know which builders were run in order to download the symbol artifacts they published.

See these changes in action in https://dev.azure.com/dnceng/internal/_build/results?buildId=2439083&view=results.

### DIFF
--- a/eng/_core/cmd/build/build.go
+++ b/eng/_core/cmd/build/build.go
@@ -45,7 +45,7 @@ func main() {
 	flag.BoolVar(&o.JSON, "json", false, "Runs tests with -json flag to emit verbose results in JSON format. For use in CI.")
 	flag.BoolVar(&o.PackBuild, "packbuild", false, "Enable creating an archive of this build using upstream 'distpack' and placing it in eng/artifacts/bin.")
 	flag.BoolVar(&o.PackSource, "packsource", false, "Enable creating a source archive using upstream 'distpack' and placing it in eng/artifacts/bin.")
-	flag.BoolVar(&o.CreatePDB, "pdb", false, "Create PDB files for all the binaries in the bin and tool directories and place them in eng/artifacts/symbols.")
+	flag.BoolVar(&o.CreatePDB, "pdb", false, "Create PDB files for all the PE binaries in the bin and tool directories. The PE files are modified in place and PDBs are placed in eng/artifacts/symbols.")
 
 	flag.BoolVar(
 		&o.Refresh, "refresh", false,
--- a/eng/_core/cmd/build/build.go
+++ b/eng/_core/cmd/build/build.go
@@ -45,6 +45,7 @@ func main() {
 	flag.BoolVar(&o.JSON, "json", false, "Runs tests with -json flag to emit verbose results in JSON format. For use in CI.")
 	flag.BoolVar(&o.PackBuild, "packbuild", false, "Enable creating an archive of this build using upstream 'distpack' and placing it in eng/artifacts/bin.")
 	flag.BoolVar(&o.PackSource, "packsource", false, "Enable creating a source archive using upstream 'distpack' and placing it in eng/artifacts/bin.")
+	flag.BoolVar(&o.CreatePDB, "pdb", false, "Create PDB files for all the binaries in the bin and tool directories and place them in eng/artifacts/symbols.")
 
 	flag.BoolVar(
 		&o.Refresh, "refresh", false,
@@ -82,6 +83,7 @@ type options struct {
 	JSON       bool
 	PackBuild  bool
 	PackSource bool
+	CreatePDB  bool
 	Refresh    bool
 	Experiment string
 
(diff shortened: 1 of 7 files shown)


gdams (approved):

eng/pipeline/stages/builders-to-stages.yml
```diff
@@ -9,10 +9,14 @@ parameters:
   builders: []
   # If true, include a signing stage+job that depends on all 'buildandpack' builder jobs finishing.
   sign: false
+  # If true, the publish build artifacts to blob storage.
```
dagood:
```suggestion
  # If true, publish build artifacts to blob storage.
```

eng/_core/cmd/build/build.go
```diff
@@ -45,6 +45,7 @@ func main() {
 	flag.BoolVar(&o.JSON, "json", false, "Runs tests with -json flag to emit verbose results in JSON format. For use in CI.")
 	flag.BoolVar(&o.PackBuild, "packbuild", false, "Enable creating an archive of this build using upstream 'distpack' and placing it in eng/artifacts/bin.")
 	flag.BoolVar(&o.PackSource, "packsource", false, "Enable creating a source archive using upstream 'distpack' and placing it in eng/artifacts/bin.")
+	flag.BoolVar(&o.CreatePDB, "pdb", false, "Create PDB files for all the binaries in the bin and tool directories and place them in eng/artifacts/symbols.")
```
dagood: It seems worth noting that this modifies the PE binaries in-place (resulting in `packbuild` being able to grab the binaries without any changes to how it works).

```suggestion
	flag.BoolVar(&o.CreatePDB, "pdb", false, "Create PDB files for all the PE binaries in the bin and tool directories. The PE files are modified in place and PDBs are placed in eng/artifacts/symbols.")
```

eng/pipeline/stages/builders-to-stages.yml
```diff
@@ -45,3 +50,29 @@ stages:
...
+    - template: pool.yml
+      parameters:
+        inner:
+          template: publish-stage.yml
+          parameters:
+            builder: { os: linux, arch: amd64 }
+            official: true
+            public: false
+            builders:
+            - ${{ each builder in parameters.builders }}:
```
dagood: Nitpick: indent the list to help IDEs.

eng/pipeline/stages/publish-stage.yml
```diff
@@ -60,6 +66,13 @@ stages:
                 artifact: BuildAssets
               ${{ else }}:
                 artifact: BuildAssetsInternal
+            - ${{ if eq(parameters.publishSymbols, true) }}:
```
dagood: Because this param has explicit type `boolean` you can shorten it:

```suggestion
            - ${{ if parameters.publishSymbols }}:
```

(eng/pipeline/stages/builders-to-stages.yml has dynamic types.)

eng/pipeline/stages/publish-stage.yml
```diff
@@ -124,3 +137,35 @@ stages:
...
+                displayName: 'Copy to flat dir: ${{ builder.id }}'
+                workingDirectory: '$(Pipeline.Workspace)'
+            - task: PublishSymbols@2
+              inputs:
+                SymbolsFolder: $(Pipeline.Workspace)/Symbols
+                SearchPattern: '*.pdb'
+                SymbolServerType: TeamServices
+                # Source indexing doesn't work for us. It needs the source files to be available
+                # in the AzDO repo, but we pull them at build time using a git submodule.
+                IndexSources: false
```
dagood: Maybe link to an issue in go-lab to track?

eng/pipeline/stages/run-stage.yml
```diff
@@ -124,16 +138,21 @@ stages:
               - pwsh: Write-Host "##vso[task.setvariable variable=GOARM]6"
                 displayName: Set GOARM for cross-compile
 
+          - ${{ if eq(variables.createPDB, true) }}:
```
dagood, line 141: Interesting... wasn't aware job variables would be accessible here. (I experienced pipeline variables *not* being accessible inside a stage template in go-images, but it seems now that that's because that stage's variables are the ones that are in scope at that point, excluding the outer variables.)

https://learn.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#variables seems very vague about it, but another thing it mentions is that they're evaluated *in order*... so maybe some of the value reuse we're currently doing through layers of templates could be done this way too.

No suggestion to change anything now, just some context about why this isn't showing up anywhere else. 😄

eng/pipeline/steps/install-gopdb.yml
```diff
@@ -0,0 +1,36 @@
...
+  - pwsh: |
+      . eng/utilities.ps1
+      $gobin = Get-Stage0GoRoot # Make sure we have a Go toolchain available
+      cd go-pdb
+      & $gobin/bin/go.exe build -o $(pdbPath)/gopdb.exe ./cmd/gopdb
+    displayName: Install gopdb
+
+  - pwsh: |
+      Remove-Item -Path go-pdb -Force -Recurse
+    displayName: Cleanup go-pdb
```
dagood:
```suggestion
    displayName: Cleanup gopdb
```

eng/pipeline/steps/install-gopdb.yml
```diff
@@ -0,0 +1,36 @@
...
+    type: string
+    default: 0.2.1
+
+steps:
+  - pwsh: |
+      $pdbPath = "$(System.ArtifactsDirectory)/gopdb"
+      New-Item $pdbPath -ItemType Directory -ErrorAction Ignore
+      Write-Host "##vso[task.setvariable variable=pdbPath;]$pdbPath"
+      Write-Host "##vso[task.prependpath]$pdbPath"
+    displayName: Setup gopdb path
```
dagood:
```suggestion
    displayName: Set up gopdb path
```

dagood (approved):

PR #1230, opened 2024-05-24
For CG: update eng/_util; omit unused modules

dagood:
Fix potential issues detected by Component Governance (CG) by upgrading some dependencies of `_util` and removing some unused upstream go.mod/go.sum files from consideration.

Test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2460123&view=results

Followup issue if we want to remove the workaround patch: https://github.com/microsoft/go-lab/issues/76


karianna (approved):
LGTM but can we raise a tracking issue in case this can get fixed in CG proper later on.

PR #1234, opened 2024-05-31
Update submodule to latest; fix patch conflicts: move TLS defaults

dagood:
* Builds on https://github.com/microsoft/go/pull/1232

Update to latest version of Go and fix patch conflicts. Mostly automatic, but:

* patches/0002-Add-crypto-backend-foundation.patch
  * Upstream incorporated the change in src/net/http/client_test.go so that no longer appears.
* patches/0004-Add-OpenSSL-crypto-backend.patch
  * Trivial: neighboring line to an import.
* patches/0008-Update-default-go.env.patch
  * Skip a test with logic that conflicts with this change: https://github.com/golang/go/issues/67793
* patches/0010-Support-TLS-1.3-in-fipstls-mode.patch
  * https://github.com/golang/go/commit/0b57881571a78e3ee9fe9001d1d7d4cc48fdc32d moved some FIPS settings to a more centralized location and made them simpler to configure.
  * Had to manually resolve some conflicts with the https://github.com/microsoft/go/pull/1232 fix for `TestBoringServerCurves/curve=29/fipstls` because the return values and neighboring lines were changed, but behavior is the same.

### DIFF
--- a/eng/pipeline/stages/run-stage.yml
+++ b/eng/pipeline/stages/run-stage.yml
@@ -206,6 +206,11 @@ stages:
                   # The Chocolatey shims are located in a single folder in PATH, so we can't change PATH to exclude it.
                   # Upstream Windows builders don't have SWIG installed, so this makes coverage even.
                   RemovePathBinary 'swig'
+                  
+                  Write-Host "Removing 'patch' to avoid parts of cmd/go TestScript/mod_tidy_diff on Windows."
+                  # patch here doesn't seem to be the same as a Linux-style patch tool, and fails:
+                  # FAIL: testdata\script\mod_tidy_diff.txt:78: exec patch -p1 -i diff.patch: exit status 57005
+                  RemovePathBinary 'patch'
                 displayName: Remove unexpected tools
 
               - pwsh: |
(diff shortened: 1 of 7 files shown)


karianna (changes requested):
Since CI failed I'm not going to review in depth just yet - setting to request changes just for workflow purposes.

eng/pipeline/stages/run-stage.yml
```diff
@@ -206,6 +206,11 @@ stages:
                   # The Chocolatey shims are located in a single folder in PATH, so we can't change PATH to exclude it.
                   # Upstream Windows builders don't have SWIG installed, so this makes coverage even.
                   RemovePathBinary 'swig'
+                  
+                  Write-Host "Removing 'patch' to avoid parts of cmd/go TestScript/mod_tidy_diff on Windows."
```
karianna, line 210: Do we need to report this issue upstream in the poweshell/Windows patch tool?

karianna (approved):

eng/pipeline/stages/run-stage.yml
dagood: IMO: no. This patch.exe seems to come from Strawberry Perl, looking at CI logs: `Removing C:\Strawberry\c\bin\patch.exe`. (I couldn't tell until seeing it in CI, and didn't repro locally.) I anticipate the Strawberry Perl team telling us this binary isn't intended to implement the same set of features and match the Linuxy output/result 1:1.

Before going any further, I'd want to grab the Strawberry Perl version of patch.exe and see the results for myself so I know what's actually going on. `exit status 57005` is pretty odd. We might even need to change CI to export its `patch.exe` specifically because it might be hard to reproduce how it's set up.

I think there is potential for some value if we looked into it, but I don't think it's worth the effort currently.

PR #1270, opened 2024-07-17
Use Release Studio for public publish

dagood:
Includes a more detailed assets.json so the new, more unique URLs are discoverable. (Automatic by updating go-infra dependency.)

The pipeline yml is the main difference. This is implemented as a series of stages (send a request -> wait for it to complete -> send a request -> wait for it to complete) by the suggestion of the DDFun team providing us with the NuGet package. This makes it so that we can retry individual stages if the publish request takes an abnormal amount of time and succeeds, but exceeds our stage's timeout. We use a different retry mechanism in microsoft/go-infra, but stage retry makes sense in microsoft/go where we don't have the parameter-driven retry logic of go-infra.

This adds another csproj. Like the signing tooling, the Release Studio tooling is also delivered to us as a .NET NuGet package.

* [Example run that publishes an existing build](https://dev.azure.com/dnceng/internal/_build/results?buildId=2497132&view=results)
* [Example run that builds its own Go version](https://dev.azure.com/dnceng/internal/_build/results?buildId=2497131&view=results)
* [Example run without public publish enabled](https://dev.azure.com/dnceng/internal/_build/results?buildId=2497144&view=results)
* As of writing, those are in progress, so here's an [earlier run that publishes an existing build](https://dev.azure.com/dnceng/internal/_build/results?buildId=2496236&view=results) that should be functionally identical. I tweaked some text and squashed before submitting this PR.

---

We plan to move this to microsoft/go-infra. Publishing in the release pipelines will allow us to avoid publishing bad builds to an even greater degree than the checkbox being added by this PR. It will also let us use Release Studio's aka.ms link generation capabilities. However, currently the top priority is to get our builds on the CDN that Release Studio is targeting, so this throwaway work is worthwhile.


eng/pipeline/steps/find-PublishBranchAlias-task.yml
```diff
@@ -0,0 +1,29 @@
...
+      # out ourselves.
+      $branch = TrimStart $branch "refs/heads/"
+
+      # Don't include "microsoft/" virtual directory prefix: we are already in the
+      # "microsoft" container, so this would result in "microsoft/microsoft/main".
+      $branch = TrimStart $branch "microsoft/"
+
+      Write-Host "PublishBranchAlias is: $branch"
+      Write-Host "##vso[task.setvariable variable=PublishBranchAlias;]$branch"
+    displayName: Find publish branch alias
```
gdams: newline missing here?

gdams (approved):
LGTM (other than one newline nit)

dagood:
Now that the example builds are done, I ran a [test release build pipeline](https://dev.azure.com/dnceng/internal/_build/results?buildId=2497229&view=results) against one of them to make sure it'll handle the aka.ms links correctly, and it looks good to me.

PR #1279, opened 2024-07-23
initial changes to build macOS

gdams:
We don't intend to release this in the near future but it makes sense to build/test it


gdams:
I'll need to work out what the label is for the macOS agents. I'm not even sure if there is a 1ES pool for this. CC @dagood?

dagood:
According to https://github.com/dotnet/arcade/blob/main/Documentation/AzureDevOps/AzureDevOpsOnboarding.md#agent-queues we can use the hosted pool for macOS in particular. However, for production builds, we could simply cross-build from a Linux agent. I don't know if it's *actually* preferable to use the 1ES pool, or if using macOS hosted pool is totally fine. We should probably ask dnceng.

eng/signing/Sign.proj
```diff
@@ -68,6 +76,20 @@
...
       <FilesToSign Remove="%(GoZipFile.ExtractedDir)**\testdata\**" />
     </ItemGroup>
+    <!-- macOS codesign -->
+    <Exec Command="if not exist %(GoMacOSTarFile.ExtractedDir) mkdir %(GoMacOSTarFile.ExtractedDir)" />
+    <Exec Command="tar -xzf %(GoMacOSTarFile.Identity) -C %(GoMacOSTarFile.ExtractedDir)" />
+    <ItemGroup>
+      <FilesToSign Include="%(GoMacOSTarFile.ExtractedDir)**\*">
+        <Authenticode>MacDeveloperHarden</Authenticode>
+      </FilesToSign>
+      <!--
```
gdams, line 83: not 100% sure we need this, it will need some testing

gdams:
Delegating to Davis who is taking this over ;)

PR #1281, opened 2024-07-24
[microsoft/release-branch.go1.23] Update submodule for 1.23

dagood:


gdams:
I see a test failure on windows

dagood:
That's an interesting one... when testing with `GOEXPERIMENT=cngcrypto` in particular, telemetry isn't working (according to these test cases). Taking a look.

dagood:
It repros upstream with `$env:GOEXPERIMENT='aliastypeparams'` (or anything else, it seems). Filing an issue.
(Edit: filed https://github.com/golang/go/issues/68579)

dagood:
For now, I added a patch to let the test case pass anyway, although the patch is specific to our CI case (cngcrypto).

karianna (approved):

PR #1304, opened 2024-09-03
[microsoft/release-branch.go1.22] Update golang-fips/openssl to db0a77fc6fcc

dagood:
* Incorporate cherry-picked fix: https://github.com/golang-fips/openssl/pull/157

Update to use https://github.com/golang-fips/openssl/commit/db0a77fc6fcca38dcf8df0ab2e28b778d2d21551, mod tidy, mod vendor, and conflict resolution for neighboring lines.

I did notice that trying to update the go.mod/go.sum/vendor with Microsoft Go 1.23.0 and upstream Go 1.23.0 yielded different results than using Go 1.22.6 (and different from each other!). The changes in this PR use 1.22.6 and look normal. I'll open issues for what I saw to figure out if this is something we should/can do something about. I *generally* expect that you can work on a Go module using any greater version of Go.


dagood:
* The root cause of the tidy/vendor oddities seem to be related to:
  https://github.com/golang/go/issues/69235

karianna (approved):

Issue #1305, opened 2024-09-03
Building app with Go 1.23 OpenSSL backend emits a build warning


dagood:
When I use `mcr.microsoft.com/oss/go/microsoft/golang:1.23-fips-cbl-mariner2.0` to build a simple app targeting linux-amd64 I get this output:

```
# vendor/github.com/golang-fips/openssl/v2
thread_setup_unix.c: In function 'thread_id':
thread_setup_unix.c:32:12: warning: 'pthread_setspecific' expecting 1 byte in a region of size 0 [-Wstringop-overread]
   32 |     (void) pthread_setspecific(destructor_key, (void*)1);
      |            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from thread_setup_unix.c:5:
/usr/lib/gcc/x86_64-pc-linux-gnu/11.2.0/include-fixed/pthread.h:1317:12: note: in a call to function 'pthread_setspecific' declared with attribute 'access (none, 2)'
 1317 | extern int pthread_setspecific (pthread_key_t __key,
      |            ^~~~~~~~~~~~~~~~~~~
```

Dockerfile repro: https://github.com/dagood/msftgoex/blob/dev/dagood/1.23-warn/sum/Dockerfile

Repeating the repro with `mcr.microsoft.com/oss/go/microsoft/golang:1.22-fips-cbl-mariner2.0` does not emit the warning.

Note: the build succeeds.

dagood:
`1.23-fips-bookworm` and `1.23-fips-azurelinux3.0` **don't** emit the warning. So far I've only seen `1.23-fips-cbl-mariner2.0` (and equivalent tags) emit it.

qmuntal:
> 1.23-fips-bookworm and 1.23-fips-azurelinux3.0 don't emit the warning. So far I've only seen 1.23-fips-cbl-mariner2.0 (and equivalent tags) emit it.

This is libc version dependent. See https://github.com/golang-fips/openssl/pull/162 for a complete explanation.

PR #1307, opened 2024-09-05
[microsoft/release-branch.go1.23] Update openssl to ms-go1.23-support, 17d05d3f692c

dagood:
* For https://github.com/microsoft/go/issues/1305
* https://github.com/golang-fips/openssl/pull/168

### DIFF
--- a/patches/0004-Add-OpenSSL-crypto-backend.patch
+++ b/patches/0004-Add-OpenSSL-crypto-backend.patch
@@ -58,10 +58,10 @@ index f0e3575637c62a..0e9aceeb832d3b 100644
  package main
  
 diff --git a/src/cmd/dist/test.go b/src/cmd/dist/test.go
-index d7cbadf7b17aef..61685c5359bfde 100644
+index 0ffcabe4164af6..056f3ea5ae9beb 100644
 --- a/src/cmd/dist/test.go
 +++ b/src/cmd/dist/test.go
-@@ -1238,12 +1238,11 @@ func (t *tester) registerCgoTests(heading string) {
+@@ -1225,12 +1225,11 @@ func (t *tester) registerCgoTests(heading string) {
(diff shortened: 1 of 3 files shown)


patches/0004-Add-OpenSSL-crypto-backend.patch
```diff
@@ -714,29 +714,29 @@ index c83a7272c9f01f..a0548a7f9179c5 100644
...
  )
 diff --git a/src/go.sum b/src/go.sum
-index c5eab7d4b2f93b..fac8a0626146bc 100644
+index a75ea98c7312df..7729cca493647c 100644
 --- a/src/go.sum
 +++ b/src/go.sum
 @@ -1,3 +1,5 @@
-+github.com/golang-fips/openssl/v2 v2.0.3 h1:9+J2R0BQio6Jz8+dPZf/0ylISByl0gZWjTEKm+J+y7Y=
-+github.com/golang-fips/openssl/v2 v2.0.3/go.mod h1:7tuBqX2Zov8Yq5mJ2yzlKhpnxOnWyEzi38AzeWRuQdg=
++github.com/golang-fips/openssl/v2 v2.0.4-0.20240905155948-17d05d3f692c h1:q7tCgQctS2aXmDVWjTV0951iRioF7Svx/hSQxymkZYo=
```
karianna, line 733: We seemed to go from a simple tag of 2.0.3 - just making sure that the new 2.0.4+hash tag is what we really wanted here.

dagood: Yeah, this is a move from the `v2` branch, which has tags, to the `ms-go1.23-support` branch, which doesn't, and in that case [Go makes up a pseudo-version](https://go.dev/ref/mod#pseudo-versions).

karianna (approved):

PR #1314, opened 2024-09-09
FIPS user guide: explain "bcrypt", simplify

dagood:
Address "bcrypt" directly in the FIPS user guide, because it shows up later in identifiers in this doc. I believe we're good at referring to it as CNG everywhere else, but it's not possible when talking about the actual calls.

~~Also, try removing the manual TOC and point readers at the GitHub Outline feature. I think it's relatively new that it actually opens a persistent sidebar with a search feature. IMO that makes it usable. I don't think there was a particular reason we had the manual TOC other than that feature not being ready/usable. I did it in this PR so that people are more likely to see/read the first few sections before the API list begins.~~


karianna (commented):
What if folks ready this ourtside of GitHub?

qmuntal:
>  I don't think there was a particular reason we had the manual TOC other than that feature not being ready/usable

(Talking as the primary maintainer of the `fips/UserGuide.md`) That TOC is automatically regenerated by the [Markdown All in One](https://markdown-all-in-one.github.io/docs/guide/#features) VS Code extension whenever a section is added or updated, so keeping up to date is trivial.

I would vote for keeping it given the previous fact and that it does provide value when viewing the file outside GitHub.

dagood:
> (Talking as the primary maintainer of the `fips/UserGuide.md`) That TOC is automatically regenerated by the [Markdown All in One](https://markdown-all-in-one.github.io/docs/guide/#features) VS Code extension whenever a section is added or updated, so keeping up to date is trivial.

Interesting, I had no idea... I remember seeing broken links, was that just a matter of running it again?

I'll put it back in and add a comment to mention that it's generated (and how).

I didn't think that anyone reads this outside of the hosting on GitHub. (We don't e.g. ship it inside our Go builds.)

qmuntal (approved):
This new sections seems like a must-have, thanks!

PR #1285, opened 2024-07-31
Update submodule to latest `master` in `microsoft/main`

ghost:
Hi! I'm a bot, and this is an automatically generated upstream sync PR. 🔃

After submitting the PR, I will attempt to enable auto-merge in the "merge commit" configuration.

For more information, visit [sync documentation in microsoft/go-infra](https://github.com/microsoft/go-infra/tree/main/docs/automation/sync.md).

PR #1320, opened 2024-09-18
Bump stage0 to go1.22.6

qmuntal:
Upstream Go master branch requires at least go1.22.6 for bootstrapping: https://go-review.googlesource.com/c/go/+/606156.

Will help unblock #1285.

### DIFF
--- a/eng/utilities.ps1
+++ b/eng/utilities.ps1
@@ -24,17 +24,17 @@ function Get-Stage0GoRoot() {
   # pre-installed. This CI script installs a consistent, official version of Go to a directory in
   # $HOME to handle this. This also makes it easier to locally repro issues in CI that involve a
   # specific version of Go. The downloaded copy of Go is called the "stage 0" version.
-  $stage0_go_version = '1.20.6'
+  $stage0_go_version = '1.22.6'
 
   $proc_arch = ([System.Runtime.InteropServices.RuntimeInformation]::ProcessArchitecture).ToString().ToLowerInvariant()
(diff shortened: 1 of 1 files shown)


dagood:
CodeQL error looks like it's the same as this one--different version in the issue title, but because all we updated in this PR is the Go version, I don't doubt it's the same root cause:

* https://github.com/github/codeql-action/issues/2467.

(I commented in the issue with a link to our run so they can confirm and hopefully help me understand when this will be fixed.)

karianna:
So is our GH Action for CodeQL failure expected?

dagood:
They don't see a connection to that issue, actually. They suggested re-running it in debug mode and filing a new one. I figured out how to do that and it's going now. (There's a checkbox when you trigger the run in a specific way. Unfortunately, I don't know which one, because I hit start before thinking to note it down. 😄)

Edit: this one:

> image

dagood:
* Filed this issue: https://github.com/github/codeql/issues/17526

karianna (approved):

karianna:
Feel free to merge if you're comfortable (I'm good with it, thanks for the explanation).

qmuntal:
> Feel free to merge if you're comfortable (I'm good with it, thanks for the explanation).

I'll try to understand a little bit better the codeql issue before merging this PR just in case we are missing something important.

qmuntal:
Submitted https://github.com/microsoft/go/pull/1329 with what I think a fix that will unblock this PR.

PR #1329, opened 2024-09-20
Run GitHub codeql workflow only for eng/_utils

qmuntal:
`github/codeql-action` doesn't work well if the `go` binary is replaced during the build command, and that's exactly what happens when building the Go toolchain. See https://github.com/github/codeql/issues/17526 for more info.

This PR reduces the scope of the GitHub codeql workflow to only run for `eng/_utils` to avoid the issue. This folder contains all the code that is really not part of the Microsoft Go toolchain but is used to build the toolchain itself. The Go toolchain will still be analyzed when running the CI in Azure Pipelines, as 1ES is tightly integrated with CodeQL.

In fact, the GitHub CodeQL workflow was not doing that much. This is the report from its last successful run:

image

It only analyzed 3 files from the thousands of files that take part in the Go toolchain compilation, and all of them where from `eng/_utils`.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -23,7 +23,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@v3
+        uses: github/codeql-action/init@294a9d92911152fe08befb9ec03e240add280cb3 # v3.26.8
         with:
           languages: go
 
@@ -33,6 +33,6 @@ jobs:
           working-directory: eng/_util
 
       - name: Perform CodeQL Analysis
-        uses: github/codeql-action/analyze@v3
+        uses: github/codeql-action/analyze@294a9d92911152fe08befb9ec03e240add280cb3 # v3.26.8
         with:
           category: /language:go
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -18,13 +18,6 @@ jobs:
     permissions:
       security-events: write
 
-    strategy:
-      fail-fast: false
-      matrix:
-        language:
-          - 'cpp'
-          - 'go'
-
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


qmuntal:
This run analyzed more files that the last run in `main` 😄 

image

.github/workflows/codeql.yml
```diff
@@ -18,27 +18,21 @@ jobs:
...
         uses: github/codeql-action/init@v3
         with:
-          languages: ${{ matrix.language }}
-          build-mode: manual
+          languages: go
 
-      - run: pwsh eng/run.ps1 submodule-refresh -shallow
-      - run: pwsh eng/run.ps1 build
+      - name: Autobuild
+        uses: github/codeql-action/autobuild@v3
```
karianna: Prefer to use the hash commit of the latest release (v3.26.8) for immutability but I'm struggling to find it!

karianna: gh api --jq '.sha' /repos/github/codeql-action/commits/v3.26.8 is the sort of CL thing you can do to find it.

karianna:
```suggestion
        uses: github/codeql-action/autobuild@294a9d92911152fe08befb9ec03e240add280cb3 # v3.26.8
```

karianna: I've given suggested change below for one of these, dependabot knows how to deal with updating the hash and comment going forwards

karianna (changes requested):

.github/workflows/codeql.yml
qmuntal: Good point, and thanks for the hash 😄

dagood (approved):
Thanks! I'll provide an update in the CodeQL issue.

karianna (approved):

.github/workflows/codeql.yml
```diff
@@ -18,27 +18,21 @@ jobs:
...
-    strategy:
-      fail-fast: false
-      matrix:
-        language:
-          - 'cpp'
-          - 'go'
-
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
```
karianna: Would be good to update this one as well.

karianna (approved):

.github/workflows/codeql.yml
qmuntal: Done!

PR #1346, opened 2024-10-01
patches: add support for coverage in FIPS mode

### LINKED ISSUE #1253: `go tool cover` doesn't work with FIPS mode enabled on Mariner 2.0
The Go runtime/coverage library uses an MD5 hash in its metadata calculations ([code for reference](https://cs.opensource.google/go/go/+/refs/tags/go1.22.4:src/runtime/coverage/emit.go;l=214)). 
The "vendor crypto backends" patch moves this to use `EVP_DigestInit_ex` on a fresh evp ctx when using OpenSSL as the backend ([code for reference](https://github.com/microsoft/go/blob/microsoft/main/patches/0006-Vendor-crypto-backends.patch#L3034)).

On Mariner 2.0, use of md5 is is explicitly disabled in FIPS mode ([code for reference](https://github.com/microsoft/azurelinux/blob/2.0/SPECS/openssl/openssl-1.1.1-fips.patch#L896)) unless `EVP_MD_CTX_FLAG_NON_FIPS_ALLOW` is set on the evp ctx.

Resultant stack trace:
```
panic: EVP_DigestInit_ex
openssl error(s):
error:060800C8:digital envelope routines:EVP_DigestInit_ex:disabled for FIPS
	crypto/evp/digest.c:135

goroutine 1 [running]:
vendor/github.com/golang-fips/openssl/v2.newEvpHash(0x2, 0x10, 0x40)
	/usr/local/go/src/vendor/github.com/golang-fips/openssl/v2/hash.go:132 +0x12e
vendor/github.com/golang-fips/openssl/v2.NewMD5(...)
	/usr/local/go/src/vendor/github.com/golang-fips/openssl/v2/hash.go:248
crypto/internal/backend.NewMD5(...)
	/usr/local/go/src/crypto/internal/backend/openssl_linux.go:135
crypto/md5.New()
	/usr/local/go/src/crypto/md5/md5.go:104 +0x3a
runtime/coverage.prepareForMetaEmit()
	/usr/local/go/src/runtime/coverage/emit.go:214 +0x197
```

Of course, this isn't that big a deal, since code coverage behaviour with FIPS mode disabled should be the same as if enabled, but it would be nice if this worked.

I don't remember what the FIPS implementations are like on other distros' OpenSSL releases, but wouldn't be surprised if this is a more general problem. Also worth remarking that Mariner 3.0 will move to using SymCrypt as the provider for FIPS mode which will heavily change this picture - in all expecting this Issue to be a LiveWith.

xnox:
Binaries built with `go build -cover` require access to MD5 at build
time and runtime, as coverage format relies on MD5 for non-security
purposes.

crypto/md5 may be blocked at runtime, as the checks there for runtime
availability may be incorrect. For example, Mariner 2.0 in FIPS mode
advertises MD5 as available, yet it fails at runtime. Which is a
feature many people rely on. Meaning creating an automatic fallback to
gocrypto MD5 implementation is not desired. And yet having coverage
feature working, would be very nice.

Copy stock gocrypto md5 implementation (cp -r src/crypto/md5
src/internal/coverage/md5), make it internal to coverage packages, and
switch internal non-security usage of "crypto/md5" to use
"internal/coverage/md5" instead. Also adjust relevant bootstraping
tools.

Remaining use cases of "crypto/md5" in teh public standard library are
considered to be security sensitive (that is "crypto/md5",
"crypto/tls", "crypto/x509", "net/smtp").

"cmd/internal/hash" is currently untouched, did not yet investigate if
it is used for security or non-security purposes.

Fixes: https://github.com/microsoft/go/issues/1253

### DIFF
--- a/patches/0015-internal-coverage-vendor-in-md5.patch
+++ b/patches/0015-internal-coverage-vendor-in-md5.patch
@@ -1,0 +1,2447 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Dimitri John Ledkov <dimitri.ledkov@surgut.co.uk>
+Date: Tue, 1 Oct 2024 18:55:51 +0100
+Subject: [PATCH] internal/coverage: vendor in md5
+
+coverage format specifies use of md5 which under systemcrypto might
+not be available at all times. Vendor in native md5 implementation and
+use it.
+
+Do not register such hash with crypto system, do not perform
+cryptotest, and update module dependencies.
+---
+ src/cmd/covdata/metamerge.go                  |   2 +-
+ src/cmd/dist/buildtool.go                     |   2 +
+ src/go/build/deps_test.go                     |  15 +-
+ src/internal/coverage/cfile/emit.go           |   2 +-
+ .../coverage/decodemeta/decodefile.go         |   2 +-
+ src/internal/coverage/encodemeta/encode.go    |   2 +-
+ .../coverage/encodemeta/encodefile.go         |   2 +-
+ src/internal/coverage/md5/example_test.go     |  42 +++
+ src/internal/coverage/md5/gen.go              | 259 +++++++++++++++
+ src/internal/coverage/md5/md5.go              | 178 +++++++++++
+ src/internal/coverage/md5/md5_test.go         | 301 ++++++++++++++++++
+ src/internal/coverage/md5/md5block.go         | 125 ++++++++
+ src/internal/coverage/md5/md5block_386.s      | 184 +++++++++++
+ src/internal/coverage/md5/md5block_amd64.s    | 197 ++++++++++++
+ src/internal/coverage/md5/md5block_arm.s      | 301 ++++++++++++++++++
+ src/internal/coverage/md5/md5block_arm64.s    | 169 ++++++++++
+ src/internal/coverage/md5/md5block_decl.go    |  12 +
+ src/internal/coverage/md5/md5block_generic.go |  13 +
+ src/internal/coverage/md5/md5block_ppc64x.s   | 212 ++++++++++++
+ src/internal/coverage/md5/md5block_s390x.s    | 177 ++++++++++
+ src/internal/coverage/pods/pods_test.go       |   2 +-
+ src/internal/pkgbits/encoder.go               |   2 +-
+ 22 files changed, 2188 insertions(+), 13 deletions(-)
+ create mode 100644 src/internal/coverage/md5/example_test.go
+ create mode 100644 src/internal/coverage/md5/gen.go
+ create mode 100644 src/internal/coverage/md5/md5.go
+ create mode 100644 src/internal/coverage/md5/md5_test.go
+ create mode 100644 src/internal/coverage/md5/md5block.go
+ create mode 100644 src/internal/coverage/md5/md5block_386.s
+ create mode 100644 src/internal/coverage/md5/md5block_amd64.s
+ create mode 100644 src/internal/coverage/md5/md5block_arm.s
+ create mode 100644 src/internal/coverage/md5/md5block_arm64.s
+ create mode 100644 src/internal/coverage/md5/md5block_decl.go
+ create mode 100644 src/internal/coverage/md5/md5block_generic.go
+ create mode 100644 src/internal/coverage/md5/md5block_ppc64x.s
+ create mode 100644 src/internal/coverage/md5/md5block_s390x.s
+
+diff --git a/src/cmd/covdata/metamerge.go b/src/cmd/covdata/metamerge.go
+index bf088b1136..3c76d61ec2 100644
+--- a/src/cmd/covdata/metamerge.go
++++ b/src/cmd/covdata/metamerge.go
+@@ -9,7 +9,6 @@ package main
+ // and "intersect" subcommands.
+ 
+ import (
+-	"crypto/md5"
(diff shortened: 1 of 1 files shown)


xnox:
```
##### Testing packages.
# internal/coverage/md5
package internal/coverage/md5 (test)
	internal/coverage/md5/md5_test.go:9:2: use of internal package crypto/internal/cryptotest not allowed
FAIL	internal/coverage/md5 [setup failed]
```

I guess i cannot just copy and paste move md5 module, so will need to either allow such usage or patch it out / fix it. Sadness.

qmuntal:
Thanks for submitting this PR. It made me think a bit more about the coverage in FIPS mode issue. I would try to avoid having to keep such a large patch file (+2000 lines) in our fork, as it means dedicating more time fixing conflicts.

I have an alternative solution: change upstream instead of changing us. MD5 might be a bit of an overkill to generate unique identifiers for code coverage internal files given that it has nothing to do with crypto. I've submitted an upstream CL switching to 128-bits FNV-1a hash, which should have the same collision resistance as MD5: https://go-review.googlesource.com/c/go/+/617360. Let's see how it evolve before taking a decision here.

xnox:
> Thanks for submitting this PR. It made me think a bit more about the coverage in FIPS mode issue. I would try to avoid having to keep such a large patch file (+2000 lines) in our fork, as it means dedicating more time fixing conflicts.
> 

I think most of it can be automated with `cp -r` at an appropriate time in the patch creation process. But yes, all the little import changes will conflict all the time.

> I have an alternative solution: change upstream instead of changing us. MD5 might be a bit of an overkill to generate unique identifiers for code coverage internal files given that it has nothing to do with crypto. I've submitted an upstream CL switching to 128-bits FNV-1a hash, which should have the same collision resistance as MD5: https://go-review.googlesource.com/c/go/+/617360. Let's see how it evolve before taking a decision here.

I like it. Reading the ./src/internal/coverage/defs.go the definition is `hash: [16]byte hash of entire meta-data payload` but it doesn't say which hash, and since the length is the same, and it looks like random numbers hence no binary format change.

Would binary instrumentation still work with FNV toolchain with this patch, when mixing in binaries compiled with old MD5 toolchain? https://go.dev/doc/build-cover

Although I'm not sure anybody has coverage turned on in release binaries, or why one would mix binaries from different compilers.

xnox:
Nonetheless updated the PR to pass all tests; and have all correct module deps declared.

dagood:
> I think most of it can be automated with `cp -r` at an appropriate time in the patch creation process. But yes, all the little import changes will conflict all the time.

FWIW, the way I would handle this is to add a test case that checks to make sure the content is identical, with a flag that updates the copy. We have a test like this for a different copy-paste here:

https://github.com/microsoft/go/blob/f3c39e6f2124a583f0ce90578616846031a4cce2/patches/0001-Add-systemcrypto-GOEXPERIMENT.patch#L133

Keeping it in its own patch file to more easily distinguish "tooled" changes vs. "thoughtful" changes would also make it easier to review and maintain in this case where the content is so large. (Similar to how we have an individual patch for rerunning go mod vendor.)

That said, I do think that fixing this upstream is more interesting if we can make that happen. 😄 (I don't personally have much of a feeling of whether or not this will work, or how it might interact with the cases you mentioned, which are interesting.)

xnox:
Also having existential thoughts on whether or not "crypto/md5" should be renamed "hash/md5".

qmuntal:
Closing this PR, as there is nothing to do here now that `go tool cover` uses FNV-11 instead of MD5: https://go-review.googlesource.com/c/go/+/617360.

xnox:
@qmuntal In src/cmd/internal/hash/hash.go they seem to introduce md5 again. Do you think that would be possible to change to FNV-11 too? Should I propose that?

qmuntal:
> In src/cmd/internal/hash/hash.go they seem to introduce md5 again. Do you think that would be possible to change to FNV-11 too? Should I propose that?

I'm not that concerned about md5 usages in the toolchain itself, as I don't expect people to use a FIPS compliant Go toolchain. Having said that, you can give it a shot. It will probably make the Go toolchain a little bit faster, which is always nice.

xnox:
> > In src/cmd/internal/hash/hash.go they seem to introduce md5 again. Do you think that would be possible to change to FNV-11 too? Should I propose that?
> 
> I'm not that concerned about md5 usages in the toolchain itself, as I don't expect people to use a FIPS compliant Go toolchain. Having said that, you can give it a shot. It will probably make the Go toolchain a little bit faster, which is always nice.

I am not sure which tool or binary is used for HTTPS communication to fetch gomodules and which cryptography is used to verify go.sums. It seems to me that all that is inside the go binary itself right? rather than a separate tool?

dagood:
> I am not sure which tool or binary is used for HTTPS communication to fetch gomodules and which cryptography is used to verify go.sums. It seems to me that all that is inside the go binary itself right? rather than a separate tool?

My understanding is yes, that's in bin/go, e.g. https://github.com/golang/go/blob/master/src/cmd/go/internal/modcmd/verify.go is a subcommand that in turn uses vendored x/mod hashing logic from https://github.com/golang/mod/blob/master/sumdb/dirhash/hash.go that uses sha256.

(I haven't traced out the execution path precisely or tried to reverse what cmd/internal/hash is used by, just vaguely aware of what's in the cmd module and thought I'd give some pointers. 😄)

PR #1348, opened 2024-10-02
Prevent gocrypto fallback for historical algorithms in FIPS mode

### LINKED ISSUE #1347: Binaries built with opensslcrypto experiment can sometimes fallback to gocrypto, for historical algorithms
Below uses MD5, but the same logic also applies to DES, 3DES, RC4.
Leaving future algorithms out of scope for this bug report (i.e. Ed25519).

- Openssl crypto on non-fips hosts usually supports MD5.
- Some may build openssl without MD5 (no-md5), though no known distributions do this yet
- There is a variety of openssl FIPS modules
- Some of them do not contain MD5 with property fips=yes (All 3.0.0+ fips modules)
- Some of them advertise MD5 as available, but then it is blocked at runtime (1.1.1 from Mainer 2.0, Old Ubuntu).
- In some of the above scenarios fallback to gocrypto MD5 is indifferent, useful, or undesired.

I would like to argue that fallback to gocrypto is undesired on any fips hosts, for historical algorithms that are being removed.

- For openssl FIPS 1.1.1 and lower, currently MD5 gets blocked at runtime because openssl module advertises MD5 as available; and toolchain/binaries are using it at runtime; which then fails at runtime.
- For openssl FIPS 3.0.0 and higher, currently MD5 is advertised as not available at all, and the same binary that previously was blocked from using MD5 - now silently fallsback to gocrypto code, potentially in violation of approved only policy of the system configured openssl.
- This lead to anomaly that “echo Hello | openssl md5” fails, and go binary that does the same also fails with old FIPS modules, but magically calculates MD5 with modern FIPS modules.
- This is particularly undesirable for very old algorithms, which has now become obsolete, historical and prohibited. And even marked as unapproved, in the existing Active 140-2 certificates.

Design proposal:

- Add an additional indicator in the Boring module to indicate if the backend is detected to be in FIPS or non-fips mode.
- In the legacy algorithms not only check for when boring is enabled, and a given algorithm is available, but also if the module is in fips mode or not
- As fips mode should not be bypassed without raising a service indicator

Previously filed as a PR at https://github.com/microsoft/go/pull/1327

xnox:
Add a new boring.FIPS() API to query if the boring backend is or is not in FIPS mode. Add bindings for openssl and CNG, but not boringcrypto.

Note that currently some openssl FIPS modules return true for SupportedHash queries, for algorithms that will be blocked at runtime. Other modules choose to instead report such algorithms as not available at all, as they have become historical.

Update boring backend logic for MD5, RC4, DES, 3DES to attempt to use boring backend when supported; but also when boring backend is in FIPS mode. This way FIPS module gets to decide how it is configured, and whether or not it will allow the operation.

This ensures that binaries that use these algorithms, correctly fail at runtime against FIPS OpenSSL v3+ modules, like they already fail at runtime against FIPS OpenSSL 1.1.1 and earlier modules.

No build/runtime behaviour changes for boringcrypto, nobackend, openssl/cng backends in non-FIPS modes.

If this change is undesired for CNG backend, let me know, and this can be rearranged to check for the OpenSSL experiment only.

Fixes: https://github.com/microsoft/go/issues/1347

Todo:
- fix boringcrypto codepath

### DIFF
--- a/patches/0015-Prevent-gocrypto-fallback-for-historical-algorithms-.patch
+++ b/patches/0015-Prevent-gocrypto-fallback-for-historical-algorithms-.patch
@@ -1,0 +1,171 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Dimitri John Ledkov <dimitri.ledkov@surgut.co.uk>
+Date: Wed, 2 Oct 2024 01:52:34 +0100
+Subject: [PATCH] Prevent gocrypto fallback for historical algorithms in FIPS
+ mode
+
+Add a new boring.FIPS() API to query if the boring backend is or is
+not in FIPS mode.
+
+Note that currently some openssl FIPS modules return true for
+SupportedHash queries, for algorithms that will be blocked at
+runtime. Other modules choose to instead report such algorithms as not
+available at all, as they have become historical.
+
+Update boring backend logic for MD5, RC4, DES, 3DES to attempt to use
+boring backend when supported; but also when boring backend is in FIPS
+mode. This way FIPS module gets to decide how it is configured, and
+whether or not it will allow the operation.
+
+This ensures that binaries that use these algorithms, correctly fail
+at runtime against FIPS OpenSSL v3+ modules, like they already fail at
+runtime against FIPS OpenSSL 1.1.1 and earlier modules.
+
+No build/runtime behaviour changes for boringcrypto, nobackend,
+openssl/cng backends in non-FIPS modes.
+---
+ src/crypto/des/cipher.go                     | 4 ++--
+ src/crypto/internal/backend/boring_linux.go  | 3 +++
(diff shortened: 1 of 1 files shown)


aravindhkudiyarasan (approved):

aravindhkudiyarasan:
@xnox   Can we make it Ready for Review ?

xnox:
I needs a rebase due to other changes, and also likely a new Godebug variables, or build tags.

It seems there are two desires to preserve ability to access unapproved (historical or future relative the vintage of a given fips module); or to block those.

dagood:
Yeah, Microsoft Go needs to be compatible. (That doesn't mean we couldn't add an "incompatible mode", but that isn't what the PR is currently, and we can't necessarily accept another mode right now because it would increase our maintenance load.) More discussion will be at https://github.com/microsoft/go/issues/1347.

PR #1350, opened 2024-10-02
Fall back to Go crypto when different RSA OAEP hashes are used

qmuntal:
[rsa.OAEPOptions.MGFHash](https://pkg.go.dev/crypto/rsa#OAEPOptions) is a niche feature that is only useful in certain use cases,. In fact, I couldn't find anyone really using it in GitHub: https://github.com/search?q=language%3AGo+MGFHash&type=code.

We were previously only supporting it in out OpenSSL backend, as CNG doesn't support specifying the MGF hash. Unfortunately, Azure Linux 3 doesn't support it neither, as both are using SymCrypt and SymCrypt doesn't support it.

There is no easy way to detect if MGF is supported by a given OpenSSL provider, so I would rather not support it at all and fall back to Go crypto instead.


qmuntal:
The diffs are a bit noise, the only real change is this line: https://github.com/microsoft/go/pull/1350/files#diff-7c141789ca0dfb33926a93dda7a2074bb76d295b171514c6624d3b173120aee5R1172

karianna (approved):

PR #1355, opened 2024-10-07
Limit support to only 2-primes RSA keys in the OpenSSL backend

qmuntal:
The built-in OpenSSL providers do support n-prime RSA keys, but the SymCrypt provider for OpenSSL only supports 2-prime RSA keys. The CNG backend already only support 2-prime RSA keys.

Only 2-prime RSA keys are FIPS compliant, other n's have compatibility and security issues. Even crypto/rsa recently deprecated [rsa.GenerateMultiPrimeKey](https://pkg.go.dev/crypto/rsa#GenerateMultiPrimeKey).

Given the above reasons, this PR updates the OpenSSL backend to only support 2-prime RSA keys, falling back to Go code if other n's are used.

### DIFF
--- a/patches/0002-Add-crypto-backend-foundation.patch
+++ b/patches/0002-Add-crypto-backend-foundation.patch
@@ -21,7 +21,7 @@ Subject: [PATCH] Add crypto backend foundation
  src/crypto/hmac/hmac_test.go                 |   2 +-
  src/crypto/internal/backend/backend_test.go  |  30 +++
  src/crypto/internal/backend/bbig/big.go      |  17 ++
- src/crypto/internal/backend/common.go        |  78 ++++++++
+ src/crypto/internal/backend/common.go        |  92 +++++++++
  src/crypto/internal/backend/isrequirefips.go |   9 +
  src/crypto/internal/backend/nobackend.go     | 193 +++++++++++++++++++
  src/crypto/internal/backend/norequirefips.go |   9 +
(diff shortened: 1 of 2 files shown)


gdams (approved):

patches/0002-Add-crypto-backend-foundation.patch
```diff
@@ -619,6 +620,19 @@ index 00000000000000..efdd080a1b7708
 +		}
 +	}
 +}
++
++func IsRSAKeySupported(primes int) bool {
++	if goexperiment.BoringCrypto {
++		return true
++	}
++	// CNG only support 2-prime RSA keys.
```
```suggestion
+	// CNG only supports 2-prime RSA keys.
```

patches/0002-Add-crypto-backend-foundation.patch
```diff
@@ -619,6 +620,19 @@ index 00000000000000..efdd080a1b7708
...
++	// The built-in OpenSSL providers do support n-prime RSA keys,
```
dagood: I don't know about this behavior firsthand, but it seems worth clarifying 3 and 1 and it sounds like this is the case:

```suggestion
+	// The built-in OpenSSL 3 providers and OpenSSL 1 do support n-prime RSA keys,
```

patches/0002-Add-crypto-backend-foundation.patch
```diff
@@ -619,6 +620,19 @@ index 00000000000000..efdd080a1b7708
...
++	// but the SymCrypt provider for OpenSSL only supports 2-prime RSA keys.
++	// Only 2-prime RSA keys are FIPS compliant, other n having compatibility
++	// and security issues. Even crypto/rsa recently deprecated rsa.GenerateMultiPrimeKey.
```
dagood: Mention the version to avoid relative "recent"?

```suggestion
+	// and security issues. Even crypto/rsa deprecated rsa.GenerateMultiPrimeKey as of Go 1.21.
```

dagood (approved):
I was going to ask about this needing a mention in the FIPS guide, but I see that n==2 is actually already covered there because it's always been the FIPS requirement. 😄

patches/0002-Add-crypto-backend-foundation.patch
dagood: (With "recent" I expected it to be in 1.24 or 1.23, but it's actually already deprecated in all in-support versions of Go.)

karianna (approved):

PR #1366, opened 2024-10-23
workflow: add automerge job to approve PRs from microsoft-golang-bot

gdams:
We're investigating switching to GitHub workflows to handle this so we can restrict the token permissions on our GitHub bots

linked to https://github.com/microsoft/go-infra/pull/183


dagood (approved):
As part of an experiment to see if this works when branch protections are in place.

PR #1382, opened 2024-10-30
Upgrade golang-fips/openssl to c51a0908

qmuntal:
Upgrading the OpenSSL backend is necessary to support Azure Linux 3 in FIPS mode. There will still be some work to to in the standard library to support AZL3.


patches/0006-Vendor-crypto-backends.patch
```diff
@@ -6854,6 +7321,147 @@ index 00000000000000..5de62f95a7a4ec
...
++}
+diff --git a/src/vendor/github.com/golang-fips/openssl/v2/zaes.go b/src/vendor/github.com/golang-fips/openssl/v2/zaes.go
+new file mode 100644
+index 00000000000000..4d945290e3bdff
+--- /dev/null
++++ b/src/vendor/github.com/golang-fips/openssl/v2/zaes.go
+@@ -0,0 +1,86 @@
++// Code generated by cmd/genaesmodes. DO NOT EDIT.
++
++//go:build !cmd_go_bootstrap
```
dagood: Got it--it's because this doesn't include cgo:

```suggestion
+//go:build !cmd_go_bootstrap && cgo
```

Nearly every other file has `import "C"` which implicitly requires cgo.

qmuntal: You saved my day! ❤️

qmuntal: https://github.com/golang-fips/openssl/pull/209

gdams (approved):

PR #1384, opened 2024-10-31
Skip SHA test that require marshaling

qmuntal:
Some OpenSSL providers (looking at you SymCrypt) don't support marshaling the internal state of the hash function. There are some tests that require that functionality to succeed. We were previously skipping those tests when using the CNG backend, which doesn't implement digest marshaling. This PR generalizes the way we detect that a given digest doesn't support marshaling so it also works for OpenSSL.

### DIFF
--- a/patches/0002-Add-crypto-backend-foundation.patch
+++ b/patches/0002-Add-crypto-backend-foundation.patch
@@ -27,7 +27,7 @@ Subject: [PATCH] Add crypto backend foundation
  src/crypto/internal/backend/norequirefips.go |   9 +
  src/crypto/internal/backend/stub.s           |  10 +
  src/crypto/md5/md5.go                        |   7 +
- src/crypto/md5/md5_test.go                   |   4 +
+ src/crypto/md5/md5_test.go                   |  14 ++
  src/crypto/purego_test.go                    |   2 +-
  src/crypto/rand/rand.go                      |   2 +-
  src/crypto/rand/rand_test.go                 |   2 +-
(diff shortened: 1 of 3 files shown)


mertakman (approved):

patches/0002-Add-crypto-backend-foundation.patch
```diff
@@ -1782,6 +1909,20 @@ index c6a2518f62ff3a..578b4d6f68504c 100644
...
  	< crypto/ed25519
+diff --git a/src/hash/marshal_test.go b/src/hash/marshal_test.go
+index 3091f7a67acede..b066f8281c5265 100644
+--- a/src/hash/marshal_test.go
++++ b/src/hash/marshal_test.go
+@@ -65,6 +65,9 @@ func TestMarshalHash(t *testing.T) {
+ 			}
+ 
+ 			h := tt.new()
++			if _, ok := h.(encoding.BinaryMarshaler); !ok {
```
dagood: Missing `if boring.Enabled {`? (Just looking at the pattern, haven't checked the context beyond the cutoff. 😄)

dagood (approved):
We might want to have a test that specifically makes sure we don't unintentionally remove marshaller support from specific OpenSSL backend scenarios... but that's probably the role of scenario testing because here we're simply doing the best we can per-platform.

patches/0002-Add-crypto-backend-foundation.patch
qmuntal: `boring` is internal to `crypto`, and this test is not under `crypto`. Worth adding the check, anyway, so I've done the usual trick to check if boring is enabled from outside `crypto`.

PR #1385, opened 2024-10-31
Fix testKeyBasics and testEverything

qmuntal:
Some crypto backends (e.g. CNG and OpenSSL with SymCrypt) fail to create RSA keys that are smaller than 512 bits. Some tests use keys lower than 512 bits, so we have to skip them. We were previously doing so for the CNG backend. This PR conditionally skips the failing tests when using any backend.

We lose a bit of coverage, as the OpenSSL backend do support small keys when using the built-in providers. There is no easy way to know which key sizes are supported by a given OpenSSL provider, and our implementation doesn't depend on the key size, so I think is makes sense to just skip the test rather than trying to be too clever.

⚠️ Don't merge until #1384 has been merged. ⚠️


karianna:
@qmuntal - merge conflicts

qmuntal:
> @qmuntal - merge conflicts

Fixed.

karianna (approved):

PR #1386, opened 2024-10-31
Support SymCrypt in TLS 1.3 handshakes

qmuntal:
TLS 1.3 handshakes use the `eencoding.BinaryMarshaler` interface to clone the hash. The SymCrypt provider for OpenSSL doesn't support that interface. Apply the same workaround as we were already doing for CNG.

⚠️ Don't merge until https://github.com/microsoft/go/pull/1385 has been merged. ⚠️

PR #1387, opened 2024-10-31
Support SymCrypt in crypto/rsa

qmuntal:
The SymCrypt provider for OpenSSL only supports digest function provided by SymCrypt itself. Unfortunately, it can happen that `SupportsHash` return true for a digest not provided by SymCrypt, but by another provider. The RSA operations will fail if that happens, so we should detect this case and fallback to Go crypto.

This PR refactors the RSA code to only use the OpenSSL backend for digest functions that we know that are also support by SymCrypt.

It also contains some refactors to make it clearer what which RSA operations are supported by each backend.

⚠️ Don't merge until https://github.com/microsoft/go/pull/1386 has been merged. ⚠️

### DIFF
--- a/patches/0004-Add-OpenSSL-crypto-backend.patch
+++ b/patches/0004-Add-OpenSSL-crypto-backend.patch
@@ -14,7 +14,7 @@ Subject: [PATCH] Add OpenSSL crypto backend
  src/crypto/ecdsa/notboring.go                 |   2 +-
  src/crypto/internal/backend/bbig/big.go       |   2 +-
  .../internal/backend/bbig/big_openssl.go      |  12 +
- src/crypto/internal/backend/openssl_linux.go  | 323 ++++++++++++++++++
+ src/crypto/internal/backend/openssl_linux.go  | 358 ++++++++++++++++++
  src/crypto/internal/boring/fipstls/stub.s     |   2 +-
  src/crypto/internal/boring/fipstls/tls.go     |   2 +-
  src/crypto/rsa/boring.go                      |   2 +-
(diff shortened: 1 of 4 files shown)


patches/0004-Add-OpenSSL-crypto-backend.patch
```diff
@@ -518,6 +518,41 @@ index 00000000000000..69af0ffe2fcf80
...
++func supportsRSAHash(h crypto.Hash) bool {
++	if !SupportsHash(h) {
++		return false
++	}
++	// The SymCrypt provider for OpenSSL only supports digest function provided by SymCrypt itself.
++	// It can happen that SupportsHash return true for a digest not provided by SymCrypt, but by another provider.
++	// The RSA operations will fail if that happens.
++	// To avoid this case we only support digest functions also supported by SymCrypt, which happens to be
++	// the most common ones.
++	return h == crypto.SHA1 || h == crypto.SHA256 || h == crypto.SHA384 || h == crypto.SHA512
```
dagood: This seems like it could be significant for someone using non-SymCrypt. (Ubuntu Pro?) FIPS guide update?

(Also, even though we don't have automation for https://github.com/microsoft/go/tree/microsoft/main/docs yet, we could start a doc for 1.24.)

karianna:
@qmuntal merged the previous one, same rebase required, thanks!

patches/0004-Add-OpenSSL-crypto-backend.patch
qmuntal: I'll revert this particular change. The SymCrypt team agreed on fixing it on their side.

qmuntal:
The SymCrypt team the RSA fallback issue on their side, so I've reverted the parts of this PR that tried to workaround the issue on our side. AZL3 pipelines will continue being broken until the SymCrypt fix lands.

This PR still has value without the fix, as it contains some nice refactors that will help with some new features I have in mind.

dagood (approved):

Issue #1368, opened 2024-10-23
Support macOS (darwin) amd64 and arm64 builds of Microsoft Go


dagood:
This issue tracks supporting Microsoft Go toolset builds for macOS on our scripts/infra. A tricky point here is signing. We need to be able to sign the macOS binaries so that when we do publish a macOS build, it's compliant. (Also, I believe only 2/4 of us have access to macOS machines at the moment.)

Note: it's probably better to do this without FIPS compliance support initially. It will be easier to work on that with the basic workflow in place. (https://github.com/microsoft/go/issues/1013)

dagood:
Signing macOS archives seems pretty nasty. If I'm reading @gdams change and how notarization works correctly, we need three passes:

1. Sign individual binaries from inside the tar.gz and put them back in.
   * I've started changing the bulk of our signing code from MSBuild to Go so that we can ensure the permissions are set correctly on our Windows builder that is running all this. Also so the logic isn't as hostile to maintenance.
2. Sign/notarize the tar.gz, which actually modifies the tar.gz itself.
3. Create a signature for the tar.gz.
   * Arguably not necessary? 

We currently have just one pass. We sign individual files in the zip, but we don't create a signature for the zip. For tar.gz files, we only create a signature.

We can use the necessity of having at least two passes as an opportunity to produce signatures for our zip files, simply to keep our artifacts more similar.

dagood:
Complete, but there is some followup:

Test builder: https://github.com/microsoft/go/issues/1394
Notarization, or determining that we don't need to do it: https://github.com/microsoft/go-lab/issues/147

PR #1388, opened 2024-11-05
Add macOS buildandpack with signing

dagood:
* Add macOS builds.
  * Based on https://github.com/microsoft/go/pull/1279
* Rework signing to support macOS hardening and notarization.
  * Support multiple signing passes, needed for macOS artifacts in particular.
  * No more complex MSBuild logic. Now written in Go.
  * Create .sig files for Windows. We can do this now that we have multiple passes.
  * Add parameter to sign existing build for faster dev iteration.
* For https://github.com/microsoft/go/issues/1368

Notarization isn't working yet, pending https://portal.microsofticm.com/imp/v5/incidents/details/560612351/summary. For now, this PR skips that step.

Test end-to-end build: https://dev.azure.com/dnceng/internal/_build/results?buildId=2575871&view=results
Test that signs an existing build: https://dev.azure.com/dnceng/internal/_build/results?buildId=2575146&view=results


eng/_util/cmd/sign/archive.go
```diff
@@ -0,0 +1,463 @@
...
+						return err
+					}
+					_, err = io.Copy(fWriter, fReader)
+					return firstError(err, fReader.Close())
+				}); err != nil {
+					return fail(err)
+				}
+				results = append(results, info)
+			}
+		}
```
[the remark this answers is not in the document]
dagood: The AI suggestion is no good, but this is indeed something to fix.

eng/_util/cmd/sign/archiveutil.go
```diff
@@ -0,0 +1,142 @@
...
+	"archive/zip"
+	"compress/gzip"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+)
+
+func eachZipEntry(r *zip.ReadCloser, f func(*zip.File) error) error {
```
qmuntal, line 19: The "each*" functions can be rewritten into iterators. They calling site would look a bit nicer and less magic.

eng/_util/cmd/sign/archiveutil.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	f, err := os.Create(path)
+	if err != nil {
+		return err
+	}
+	_, err = io.Copy(f, r)
+	return firstError(err, f.Close())
+}
+
+func firstError(errs ...error) error {
```
qmuntal: Since Go 1.22 this pattern is generically implemented in the standard library by [cmp.Or](https://pkg.go.dev/cmp#Or). I think it can replace `firstError`.

eng/_util/cmd/sign/sign.go
```diff
@@ -0,0 +1,287 @@
...
+		if err := ctx.Err(); err != nil {
+			return nil, err
+		}
+		// Ignore checksum files: we always generate new ones.
+		if strings.HasSuffix(f, ".sha256") {
+			continue
+		}
+
+		filename := filepath.Base(f)
+		if existingF, ok := archiveFilenames[filepath.Base(f)]; ok {
```
qmuntal: This does a case sensitive check, but some OSes are case insensitive (e.g. Windows). We should at least convert the file name to lower case before checking and adding file names to the map.

gdams (approved):

qmuntal (changes requested):
Done a first pass review. Overall looks great.

eng/_util/cmd/sign/archiveutil.go
dagood: I'm curious what you think about handling errors in iterators. My first impression is that `for v, err := range eachZipEntry(...)` is odd, but it's the solution I'd go with. It does look like this is popping up in the wild.

---

Other factor: our bootstrap is still 1.22, and range func is new to 1.23. (I did initially dabble with an iterator but this stopped me early on.)

Also, if we decide to move this code to go-infra, we'd have to upgrade that module from 1.22 to 1.23. (But we could just wait until 1.24 releases and do it then--if we even have a reason to. I don't anticipate e.g. moving signing into the release pipelines.)

---

Personally, I think the callbacks are just fine. A downside is that you can't interrupt them (without returning an error value) but this code doesn't need to do that. This is certainly down to my background and whims, though. 😄

eng/_util/cmd/sign/archiveutil.go
dagood: Neat! I glanced at `slices` for something like that, missed that one. Odd name but it's an exact fit.

eng/_util/cmd/sign/archiveutil.go
qmuntal: You convinced me. `each*` is fine :)

qmuntal (approved):

PR #1417, opened 2024-11-20
Add making changes section to DeveloperGuide

mertakman:
This part of document covers:
- Preparing `go/src` folder without falling much duplication with previous sections,
- Making changes on `go/src` and small information about our approach
- Generating new patch files and squashing changes to existing patch files
- Rolling back changes in `go/src` after generating patch files

### DIFF
--- a/eng/doc/DeveloperGuide.md
+++ b/eng/doc/DeveloperGuide.md
@@ -173,12 +173,11 @@ Creating new patch files is not always necessary when there are existing patch f
 
 Before starting work, please check the go/patches folder for any existing patch files related to the files you're working on. This helps maintain a clean repository by avoiding redundant patch files.
 
-### Rolling Back Changes in `go/src`
+### Submitting Changes
 
-We do not retain changes in the submodule after generating patch files, as this approach is similar to maintaining a fork. The submodule is pinned to a specific commit hash, and we always start from that point, applying patch files on top of it. Therefore, once the work is complete and patch files are generated, it is essential to clean up the working directory. 
-
-Additionally, any new commits will cause `git go-patch extract` to generate new patch files. To return the `go/src` directory to its initial state after completing your work, run the following command:
-
+When working with the `go/src` submodule, changes should not be committed directly. Instead, we use patch files to manage modifications. Since the submodule is pinned to a specific commit hash, we always start from that state and apply patch files on top. This approach avoids the need to maintain a fork and ensures consistency.
+After generating the required patch files and completing your work, it is crucial to clean up the submodule to prevent any changes from being committed. To restore the submodule to its original state, execute the following command:
 ```bash
 git submodule update --init --recursive --checkout
 ```
+This ensures the submodule remains clean and aligned with its designated commit state. Once done, you can proceed to commit the patch files to your pull request.
--- a/eng/doc/DeveloperGuide.md
+++ b/eng/doc/DeveloperGuide.md
@@ -140,25 +140,6 @@ You can use your build of `go` in VS Code by following these steps:
 
 ## Making Changes to `go/src`
 
-### Preparing `go/src` folder
-As mentioned above, after downloading the submodule using the command:
-
-```bash
-git submodule update --init --recursive
-```
-
-and applying all changes from patch files with:
-
-```bash
-git go-patch apply
-```
-
-fully prepared go/src directory will be created. From this point forward, all tasks related to vendoring, updating the go.mod file, and running tests should exclusively utilize the Go toolchain provided by the go submodule.
-
-It is crucial to note that the Go version and source code within the submodule may differ from your local setup. As a result, tests that pass in your local environment may fail in the submodule environment, and tests that pass in the submodule may fail locally. Always validate your changes within the submodule context to ensure compatibility and consistency.
-
-### Making Changes
-
 Once the `go/src` folder is prepared, any modifications made to this directory will be tracked by the Git history of the submodule. You can view these changes by running:
 
 ```bash
--- a/eng/doc/DeveloperGuide.md
+++ b/eng/doc/DeveloperGuide.md
@@ -138,7 +138,7 @@ You can use your build of `go` in VS Code by following these steps:
 1. Open the command palette.
 1. Search for `Developer: Reload Window` and select it.
 
-## Making Changes to `go/src`
+## Making changes to `go/src`
 
 Once the `go/src` folder is prepared, any modifications made to this directory will be tracked by the Git history of the submodule. You can view these changes by running:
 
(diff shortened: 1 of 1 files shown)


eng/doc/DeveloperGuide.md
```diff
@@ -138,6 +138,66 @@ You can use your build of `go` in VS Code by following these steps:
...
+
+and applying all changes from patch files with:
+
+```bash
+git go-patch apply
+```
+
+fully prepared go/src directory will be created. From this point forward, all tasks related to vendoring, updating the go.mod file, and running tests should exclusively utilize the Go toolchain provided by the go submodule.
+
+It is crucial to note that the Go version and source code within the submodule may differ from your local setup. As a result, tests that pass in your local environment may fail in the submodule environment, and tests that pass in the submodule may fail locally. Always validate your changes within the submodule context to ensure compatibility and consistency.
```
qmuntal: I don't think this section adds much value, as it mostly contain duplicated information from a previos section. If you think there is something missing in the previous section, please update it.

eng/doc/DeveloperGuide.md
```diff
@@ -138,6 +138,66 @@ You can use your build of `go` in VS Code by following these steps:
...
+
+The `go-patch` command will generate a patch file under the `go/patches` directory. The patch file will be prefixed with a serial number (one greater than the number of existing patch files), followed by a dash-separated commit message.
+
+### Squashing Changes to Existing Patch Files
+
+Creating new patch files is not always necessary when there are existing patch files with similar purposes for the same files. In such cases, you can squash new commits on top of the existing ones to update their contents. The go-patch extract command will detect the differences in these commits and regenerate the patch files with the updated contents.
+
+Before starting work, please check the go/patches folder for any existing patch files related to the files you're working on. This helps maintain a clean repository by avoiding redundant patch files.
+
+### Rolling Back Changes in `go/src`
```
qmuntal: The concept explain in this section is important, but I don't think it deserved it's won section. What if you add a new paragraph to the `Making Changes` section saying something like this:

> Before starting any new work in `go/src`, make sure that the Go submodule has all the patches are applied and that it doesn't contain previous changes. You can do that by running the following command:

```
git go-patch apply -f
```

qmuntal (changes requested):

eng/doc/DeveloperGuide.md
mertakman: This paragraph is not related to whether patches have already been applied. It states that before starting to work, one must check if a new patch file needs to be created or if the changes can be added to an existing patch file.

qmuntal: Srry, my comment was referring to the `### Rolling Back Changes in go/src` section, I put the comment in the wrong line 😓 .

qmuntal (changes requested):

eng/doc/DeveloperGuide.md
mertakman: I think it makes its own sense, since the storyline in here goes as;
Beginning
Process
Final.
So final cleanup can require more additional steps in future. and IMO keeping it in this way is making much more sense

eng/doc/DeveloperGuide.md
```diff
@@ -138,6 +138,47 @@ You can use your build of `go` in VS Code by following these steps:
...
+
+We do not retain changes in the submodule after generating patch files, as this approach is similar to maintaining a fork. The submodule is pinned to a specific commit hash, and we always start from that point, applying patch files on top of it. Therefore, once the work is complete and patch files are generated, it is essential to clean up the working directory. 
+
+Additionally, any new commits will cause `git go-patch extract` to generate new patch files. To return the `go/src` directory to its initial state after completing your work, run the following command:
+
+```bash
+git submodule update --init --recursive --checkout
```
qmuntal, line 181: Better run the following command, which puts the submodule into a clean state ready to start doing new changes.

```suggestion
git go-patch apply -f
```

qmuntal (changes requested):

eng/doc/DeveloperGuide.md
```diff
@@ -138,6 +138,47 @@ You can use your build of `go` in VS Code by following these steps:
 1. Open the command palette.
 1. Search for `Developer: Reload Window` and select it.
 
-## Making changes to go/src
+## Making Changes to `go/src`
```
dagood: We don't need to follow the Microsoft style guide, but it does say to use sentence case here rather than title case, and this doc and other docs in the repo already do. https://learn.microsoft.com/en-us/style-guide/scannable-content/headings#formatting-headings

eng/doc/DeveloperGuide.md
```diff
@@ -138,6 +138,47 @@ You can use your build of `go` in VS Code by following these steps:
...
+
+### Squashing Changes to Existing Patch Files
+
+Creating new patch files is not always necessary when there are existing patch files with similar purposes for the same files. In such cases, you can squash new commits on top of the existing ones to update their contents. The go-patch extract command will detect the differences in these commits and regenerate the patch files with the updated contents.
+
+Before starting work, please check the go/patches folder for any existing patch files related to the files you're working on. This helps maintain a clean repository by avoiding redundant patch files.
+
+### Rolling Back Changes in `go/src`
+
+We do not retain changes in the submodule after generating patch files, as this approach is similar to maintaining a fork. The submodule is pinned to a specific commit hash, and we always start from that point, applying patch files on top of it. Therefore, once the work is complete and patch files are generated, it is essential to clean up the working directory.
```
dagood: I don't think the meaning of "retain" is clear here... the changes do stay in the submodule after running `git go-patch extract`, so it sounds like something other than that is meant? I'm not following what the connection is to maintaining a fork, and why it's important to clean up.

dagood (commented):
I think the doc needs to get into `git go-patch rebase`, and I have some editing thoughts, but can save those for future PRs. Main comment is the one about cleanup.

eng/doc/DeveloperGuide.md
```diff
@@ -138,6 +138,47 @@ You can use your build of `go` in VS Code by following these steps:
...
+
+The `go-patch` command will generate a patch file under the `go/patches` directory. The patch file will be prefixed with a serial number (one greater than the number of existing patch files), followed by a dash-separated commit message.
+
+### Squashing Changes to Existing Patch Files
+
+Creating new patch files is not always necessary when there are existing patch files with similar purposes for the same files. In such cases, you can squash new commits on top of the existing ones to update their contents. The go-patch extract command will detect the differences in these commits and regenerate the patch files with the updated contents.
+
+Before starting work, please check the go/patches folder for any existing patch files related to the files you're working on. This helps maintain a clean repository by avoiding redundant patch files.
+
+### Rolling Back Changes in `go/src`
```
qmuntal: As we discussed offline, rephrase this section so that the intent is clearer.

qmuntal (changes requested):

qmuntal (approved):
Approving, although I think that the last section still needs some rewording (which can happen in a follow up PR).

Issue #1108, opened 2024-01-11
Support crypto/dsa


qmuntal:
[crypto/dsa](https://pkg.go.dev/crypto/dsa) is part of the standard library, but still not supported by CNG nor OpenSSL.

qmuntal:
CNG PR submitted: https://github.com/microsoft/go-crypto-winnative/pull/50

qmuntal:
OpenSSL WIP PR submitted: https://github.com/golang-fips/openssl/pull/135

PR #1421, opened 2024-11-26
Add GenerateDSAParameters and SupportsDSA functions 

mertakman:


qmuntal (approved):

patches/0002-Add-crypto-backend-foundation.patch
```diff
@@ -854,6 +886,14 @@ index 00000000000000..08600a2c833ac7
 +func VerifyEd25519(pub *PublicKeyEd25519, message, sig []byte) error {
 +	panic("cryptobackend: not available")
 +}
++
++func SupportsDSA(l, n int) bool {
++	return false
```
dagood: Why `return false` rather than `panic("cryptobackend: not available")`? (This is the only `return` in the file.)

patches/0003-Add-BoringSSL-crypto-backend.patch
```diff
@@ -259,3 +259,11 @@ index 00000000000000..7c5fbeea717618
 +func VerifyEd25519(pub *PublicKeyEd25519, message, sig []byte) error {
 +	panic("cryptobackend: not available")
 +}
++
++func SupportsDSA(l, n int) bool {
++	return false
```
dagood, line 264: (On the other hand, this particular return does make sense, because it's tied to a backend that might work.)

patches/0005-Add-CNG-crypto-backend.patch
```diff
@@ -467,6 +467,18 @@ index 00000000000000..3d3d13709de5ac
 +func VerifyEd25519(pub *PublicKeyEd25519, message, sig []byte) error {
 +	panic("cryptobackend: not available")
 +}
++
++func SupportsDSA(l, n int) bool {
++	return n != 224
```
dagood: It seems unusual to me to support all sizes *except* for one... add a comment? Or add a `cng.SupportsDSA` function to the CNG backend repo and call that to let it decide and document this instead?

qmuntal: Agree that this deserve a comment. Having `cng.SupportsDSA` might be an overkill for now.

patches/0002-Add-crypto-backend-foundation.patch
mertakman: fixed! Thanks for pointing this!

dagood (approved):

qmuntal:
For https://github.com/microsoft/go/issues/1108.

Issue #1416, opened 2024-11-20
Reconcile our patches with upstream FIPS changes


qmuntal:
Upstream Go is making many changes to the crypto packages in order to natively support FIPS. Some of these changes conflicts with our patches.

This is an umbrella issue to keep track of the reconciliation work. Note that we still won't migrate to the new upstream FIPS module APIs. The plan is to keep reusing the Boring infrastructure at least for Go 1.24.

PR #1383, opened 2024-10-30
Update submodule to latest `master` in `microsoft/main`

ghost:
Hi! I'm a bot, and this is an automatically generated upstream sync PR. 🔃

After submitting the PR, I will attempt to enable auto-merge in the "merge commit" configuration.

For more information, visit [sync documentation in microsoft/go-infra](https://github.com/microsoft/go-infra/tree/main/docs/automation/sync.md).

PR #1439, opened 2024-12-12
Minimize TLS PRF upstream diffs

qmuntal:
Upstream did an importat refactor in the `crypto/tls/prf.go` file that produces a lot of conflicts in our patches. The main issue is that usptream PRF functions don't return an error, while the ones implemented on our backend do. Passing the error up means changing many signatures and call sites.

This PR reduces these future conflicts by panicking instead of returning an error. We then keep the previous behavior by recovering from the panic (if there is one) and lowering the panic into an error.

For https://github.com/microsoft/go/issues/1416.
For https://github.com/microsoft/go/pull/1383.

### DIFF
--- a/patches/0002-Add-crypto-backend-foundation.patch
+++ b/patches/0002-Add-crypto-backend-foundation.patch
@@ -51,12 +51,11 @@ Subject: [PATCH] Add crypto backend foundation
  src/crypto/sha512/sha512_test.go             |  20 +-
  src/crypto/tls/boring_test.go                |   5 +
  src/crypto/tls/cipher_suites.go              |   2 +-
- src/crypto/tls/handshake_client.go           |  25 ++-
- src/crypto/tls/handshake_server.go           |  25 ++-
+ src/crypto/tls/handshake_client.go           |  10 +-
+ src/crypto/tls/handshake_server.go           |  10 +-
  src/crypto/tls/handshake_server_tls13.go     |  10 +
- src/crypto/tls/key_schedule.go               |  23 ++-
- src/crypto/tls/prf.go                        |  77 ++++---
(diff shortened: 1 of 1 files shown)


gdams (approved):
Great to see a shrinking patch! 🎉

patches/0002-Add-crypto-backend-foundation.patch
```diff
@@ -1762,204 +1679,57 @@ index a7fa3370e66c82..3470372b567eba 100644
...
++
  // prf10 implements the TLS 1.0 pseudo-random function, as defined in RFC 2246, Section 5.
--func prf10(result, secret, label, seed []byte) {
-+func prf10(result, secret, label, seed []byte) error {
-+	if boring.Enabled && boring.SupportsTLS1PRF() && boring.SupportsHash(crypto.MD5SHA1) {
+ func prf10(result, secret, label, seed []byte) {
++	if boring.Enabled && boring.SupportsTLS1PRF() {
 +		if err := boring.TLS1PRF(result, secret, label, seed, nil); err != nil {
-+			return fmt.Errorf("crypto/tls: prf10: %v", err)
++			panic(boringPRFError{fmt.Errorf("crypto/tls: prf10: %v", err)})
```
dagood: Could wrap this in a func that assembles the struct and panics (kind of symmetric to `recoverFromBoringPRFError`), but this side is pretty simple compared to the recover, so I'm indifferent. 😄

dagood (approved):

patches/0002-Add-crypto-backend-foundation.patch
qmuntal: A good suggestion is a good suggestion. Will do 👍.

PR #1441, opened 2024-12-12
Remove `crypto/internal/boring/fipstls` dependency from `crypto/internal/backend`

qmuntal:
(Review ignoring whitespaces)

Upstream refactored the fipstls logic to support the new built-in FIPS module. This is causing many conflicts in the upstream sync PR.

This PR moved the logic that detects whether FIPS mode is required or not (e.g. setting the `GOFIPS` env var) to its own package that only depends on `syscall`. This way we can easily import it elsewhere, for example in `crypto/internal/boring/fipstls` and make `crypto/internal/backend` not depend on the former. Doing so will greatly simplify reconciling the conflicts in the upstream sync PR.

The refactor shouldn't change any behavior.

For https://github.com/microsoft/go/issues/1416.
For https://github.com/microsoft/go/pull/1383.

### DIFF
--- a/patches/0002-Add-crypto-backend-foundation.patch
+++ b/patches/0002-Add-crypto-backend-foundation.patch
@@ -26,13 +26,15 @@ Subject: [PATCH] Add crypto backend foundation
  src/crypto/hmac/hmac_test.go                 |   2 +-
  src/crypto/internal/backend/backend_test.go  |  30 +++
  src/crypto/internal/backend/bbig/big.go      |  17 ++
- src/crypto/internal/backend/common.go        |  92 ++++++++
- src/crypto/internal/backend/isrequirefips.go |   9 +
+ src/crypto/internal/backend/common.go        |  68 ++++++
+ .../backend/fips140/fips140.go               |  55 +++++
+ .../backend/fips140/isrequirefips.go         |   9 +
+ .../backend/fips140/norequirefips.go         |   9 +
+ .../backend/fips140/nosystemcrypto.go        |  11 +
  src/crypto/internal/backend/nobackend.go     | 223 +++++++++++++++++++
- src/crypto/internal/backend/norequirefips.go |   9 +
  src/crypto/internal/backend/stub.s           |  10 +
  src/crypto/internal/boring/fipstls/stub.s    |   2 +-
- src/crypto/internal/boring/fipstls/tls.go    |   2 +-
+ src/crypto/internal/boring/fipstls/tls.go    |   7 +-
  src/crypto/md5/md5.go                        |   7 +
  src/crypto/md5/md5_test.go                   |  14 ++
  src/crypto/purego_test.go                    |   2 +-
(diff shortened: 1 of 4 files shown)


patches/0002-Add-crypto-backend-foundation.patch
```diff
@@ -672,6 +657,116 @@ index 00000000000000..bc595e91024f11
...
++
++// Enabled reports whether FIPS 140 mode is enabled by using GOFIPS=1, GOLANG_FIPS=1,
++// the 'requirefips' build tag, or any other platform-specific mechanism.
++func Enabled() bool {
++	return enabled
++}
++
++var enabled bool
++
++// Disabled reports whether FIPS 140 mode is disabled by using GOFIPS=0 or GOLANG_FIPS=1.
```
```suggestion
+// Disabled reports whether FIPS 140 mode is disabled by using GOFIPS=0 or GOLANG_FIPS=0.
```

patches/0005-Add-CNG-crypto-backend.patch
```diff
@@ -225,11 +223,6 @@ index 00000000000000..39c3c7043ab720
 +			panic("cngcrypto: not in FIPS mode")
 +		}
 +	}
-+	if enabled, _ := cng.FIPS(); enabled {
-+		// FIPS mode is enabled,
-+		// so force FIPS mode for crypto/tls and crypto/x509.
-+		fipstls.Force()
```
dagood, line 133: There doesn't seem to be any mechanism for CNG's FIPS mode to force FIPS TLS anymore. (Tested on my Windows machine, and non-FIPS algorithms do show up where they don't with 1.23.)

patches/0002-Add-crypto-backend-foundation.patch
```diff
@@ -672,6 +657,116 @@ index 00000000000000..bc595e91024f11
...
++++ b/src/crypto/internal/backend/fips140/isrequirefips.go
+@@ -0,0 +1,10 @@
++// Copyright 2024 The Go Authors. All rights reserved.
++// Use of this source code is governed by a BSD-style
++// license that can be found in the LICENSE file.
++
++//go:build requirefips
++
++package backend
++package fips140
```
dagood, line 1032: Double packages.

patches/0004-Add-OpenSSL-crypto-backend.patch
```diff
@@ -191,12 +194,72 @@ index 00000000000000..e6695dd66b1d02
...
 +var Dec = bbig.Dec
+diff --git a/src/crypto/internal/backend/fips140/nosystemfips.go b/src/crypto/internal/backend/fips140/nosystemfips.go
+index 9e12c6d5d2239c..06b08be3ce5106 100644
+--- a/src/crypto/internal/backend/fips140/nosystemfips.go
++++ b/src/crypto/internal/backend/fips140/nosystemfips.go
+@@ -2,6 +2,8 @@
+ // Use of this source code is governed by a BSD-style
+ // license that can be found in the LICENSE file.
+ 
++//go:build !goexperiment.opensslcrypto
```
dagood: Could this be `!goexperiment.systemcrypto`?

(Even if a backend doesn't implement `systemFIPSMode`, it'd be less prone to conflicts to have each one add its own file that also simply returns false.)

dagood (changes requested):
I like the factoring, seems like a few missing pieces though.

patches/0004-Add-OpenSSL-crypto-backend.patch
qmuntal: Agree with both. Will do 👍.

patches/0005-Add-CNG-crypto-backend.patch
qmuntal: Good catch. I'll manually call `BCryptGetFipsAlgorithmMode` in CNG's `systemFIPSMode` implementation so that we get back the previous behavior.

qmuntal:
> I like the factoring, seems like a few missing pieces though.

All missing pieces have been added.

patches/0005-Add-CNG-crypto-backend.patch
```diff
@@ -532,6 +527,42 @@ index bc595e91024f11..7766d674f5cfaf 100644
...
++// and BCryptGetFipsAlgorithmMode supports them all.
++var (
++	bcrypt = syscall.MustLoadDLL(sysdll.Add("bcrypt.dll"))
++
++	bcryptGetFipsAlgorithmMode = bcrypt.MustFindProc("BCryptGetFipsAlgorithmMode")
++)
++
++func systemFIPSMode() bool {
++	var enabled uint32
++	bcryptGetFipsAlgorithmMode.Call(uintptr(unsafe.Pointer(&enabled)))
```
dagood: Handle an error here? (In case the call puts nonsense into `enabled` in some error case?)

dagood (approved):
Nice! Works for me.

gdams (approved):

PR #1449, opened 2024-12-17
Implement crypto/hkdf with the OpenSSL/CNG backends

### LINKED ISSUE #1414: Implement crypto/hkdf with the OpenSSL/CNG backends
Some internal customers want to use HKDF in compliance with the Microsoft crypto policy. This means using OpenSSL/CNG under the hood.

Go 1.24 will most likely move `golang.org/x/crypto/hkdf` into the standard library package `crypto/hkdf` (see the [proposal](https://github.com/golang/go/issues/61477)). We should patch it so it uses OpenSSL/CNG under the hood.

Note that both backends already support HKDF, as we were using it on other parts of the standard library. There is some work to do to better support some small changes in the `crypto/hkdf` API.

gdams:
fixes: https://github.com/microsoft/go/issues/1414


qmuntal (approved):
LGTM!

dagood (approved):

karianna:
@gdams merge conflict

PR #1450, opened 2024-12-17
Implement crypto/pbkdf2 with the OpenSSL/CNG backends

### LINKED ISSUE #1415: Implement crypto/pbkdf2 with the OpenSSL/CNG backends
Some internal customers want to use PBKDF2 in compliance with the Microsoft crypto policy. This means using OpenSSL/CNG under the hood.

Go 1.24 will most likely move `golang.org/x/crypto/pbkdf2` into the standard library package `crypto/pbkdf2` (see [the proposal](https://github.com/golang/go/issues/69488)). We should patch it so it uses OpenSSL/CNG under the hood. Note that those backends already support that algorithm.

gdams:
fixes: https://github.com/microsoft/go/issues/1415


qmuntal (approved):
LGTM

dagood (approved):

PR #1453, opened 2024-12-18
initial implementation of macOS crypto backend

### LINKED ISSUE #1013: Consider FIPS support on macOS / darwin
As of writing, we've been asked *about* macOS FIPS support, but we haven't been asked to provide it. I'm opening this issue to keep track of any interest.

My reading of the internal Microsoft crypto policy is that using OpenSSL on macOS would be considered noncompliant. Instead, the platform-provided CommonCrypto, Security Transforms, and Secure Transport APIs must be used. If I'm wrong, or for anyone looking into this from outside Microsoft, maybe this isn't an issue. In that case, https://github.com/golang-fips/openssl/pull/98 is an effort to make the OpenSSL backend work on macOS (and Windows), although we haven't determined for certain that it will work in microsoft/go releases.

I haven't looked into any of those macOS-provided libraries to see what kind of effort it would take to make a backend for it.

gdams:
closes: https://github.com/microsoft/go/issues/1013


dagood:
Filed https://github.com/actions/runner-images/issues/11331 about the GOROOT workaround on macOS.

dagood:
Now, test failures.

patches/0007-Add-Darwin-crypto-backend.patch
```diff
@@ -0,0 +1,601 @@
...
++func IsPKCS1v15HashSupported(hash crypto.Hash) bool {
++	if goexperiment.DarwinCrypto {
++		switch hash {
++		case crypto.SHA1, crypto.SHA224, crypto.SHA256, crypto.SHA384, crypto.SHA512, 0:
++			return true
++		}
++		return false
++	}
++	return true
++}
```
qmuntal: Move this functions to the backend files (as we do for other Supports* helpers). Also, rename the functions to follow the  `Supports*` pattern instead of the `Is*Supported` pattern.

qmuntal (changes requested):

patches/0007-Add-Darwin-crypto-backend.patch
gdams: updated PTAL

eng/pipeline/stages/go-builder-matrix-stages.yml
```diff
@@ -81,6 +81,13 @@ stages:
             - { os: linux, arch: arm64, config: buildandpack }
         - ${{ if parameters.innerloop }}:
           - { os: darwin, arch: amd64, config: devscript }
+          - { os: darwin, arch: amd64, config: test }
+          - { experiment: darwincrypto, os: darwin, arch: amd64, config: test }
+          - { experiment: darwincrypto, os: darwin, arch: amd64, config: test, fips: true }
+          # - { os: darwin, arch: arm64, config: devscript }
```
gdams, line 87: noting that I'll enable this in a follow-up PR once a new version of GoTool@0 has been released that contains https://github.com/microsoft/azure-pipelines-tasks/pull/20783

qmuntal (approved):
LGTM!

mertakman (approved):

dagood:
Turns out the GOROOT issue is AzDO-specific, in GoTool@0: https://github.com/microsoft/azure-pipelines-tasks/issues/20796

PR #1462, opened 2025-01-07
Use standardized hash.CloneHash interface

gdams:
Upstream will most likely define a standardized hash clone interface: https://github.com/golang/go/issues/69521#issuecomment-2518198245

```go
type CloneHash interface {
    hash.Hash
    Clone() hash.Hash
}
```

We should follow suite so that our hash types are compatible with upstream. The only difference with our current approach is that the new `Clone` method doesn't return an error, which makes sense given that other hash methods are not allowed to return an error. It is also unlikely that CNG will return an error when duplicating a hash unless there is a programmer error or the there is no memory available in the system, in which case it makes sense to panic anyway.

### DIFF
--- a/patches/0002-Add-crypto-backend-foundation.patch
+++ b/patches/0002-Add-crypto-backend-foundation.patch
@@ -39,7 +39,7 @@ Subject: [PATCH] Add crypto backend foundation
  src/crypto/internal/cryptotest/allocations.go |   2 +-
  .../internal/cryptotest/implementations.go    |   2 +-
  src/crypto/md5/md5.go                         |   7 +
- src/crypto/md5/md5_test.go                    |  14 ++
+ src/crypto/md5/md5_test.go                    |  21 ++
  src/crypto/pbkdf2/pbkdf2.go                   |   4 +
  src/crypto/pbkdf2/pbkdf2_test.go              |   2 +-
  src/crypto/purego_test.go                     |   2 +-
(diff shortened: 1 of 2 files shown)


patches/0002-Add-crypto-backend-foundation.patch
```diff
@@ -2182,23 +2182,22 @@ index 76fff6974e7403..218d40171d2567 100644
...
  	marshaler, ok := in.(binaryMarshaler)
  	if !ok {
 +		if boring.Enabled {
 +			// CNG and OpenSSL with SymCrypt hash functions do not implement the
 +			// encoding.BinaryMarshaler interface, but they do implement the Clone method.
-+			if cloner, ok := in.(interface{ Clone() (hash.Hash, error) }); ok {
-+				if out, err := cloner.Clone(); err == nil {
-+					return out
-+				}
++			out := in.(interface{ Clone() hash.Hash }).Clone()
```
dagood: Would it be a good idea to keep `ok` and panic more descriptively if this fails?

qmuntal: In fact I'm not 100% sure that only non-boring hash objects can get to this point, that's why the previous behavior was falling back to `MarshalBinary` if the object didn't implement the `Clone` method. I would keep the same behavior for now.

qmuntal (changes requested):

qmuntal (approved):

mertakman (approved):

Issue #1377, opened 2024-10-28
Add migration guide


qmuntal:
We should have a migration guide to help 1P customers migrate from upstream Go to Microsoft Go.

This guide should at least contain the following information:

- How to acquire Microsoft Go
- Microsoft Go divergences with upstream Go
- Links to any existing useful documentation

gdams:
I'll take a first pass

dagood:
An angle to consider here is various forms of migration from "where they are", e.g. minimal, distroless, or scratch Azure Linux 3 containers. Making sure we say that OpenSSL must be installed *and* SymCrypt/SCOSSL packages, for example. (And perhaps troubleshooting notes: what `panic`s occur for various situations.)

dagood:
Needs to include a description/expansion of this:

> Cgo introduces a dependency on the build system's version of glibc. This may make the program incompatible with a different Linux distribution if it has a lower version of glibc. It's recommended to build and deploy using the same OS, but if deployment compatibility is a requirement, a common solution is to build on the oldest possible OS, or manually target an old version of glibc.

gdams:
Done

PR #1471, opened 2025-01-09
Add `Cross-Platform Cryptography` document

qmuntal:
This document is a port of [Cross-Platform Cryptography in .NET](https://learn.microsoft.com/en-us/dotnet/standard/security/cross-platform-cryptography).

This guide aims to replace our [FIPS User Guide](https://github.com/microsoft/go/blob/c09f0a94c696e2bf7d0c0d1080d2927c0fdaea25/eng/doc/fips/UserGuide.md) by being more user-friendly and easier to maintain. It is also a nice visual summary of what each backend supports and what needs more work.

Preview: https://github.com/microsoft/go/blob/dev/qmuntal/cpdoc/eng/doc/CrossPlatformCryptography.md

For #1377.

### DIFF
--- a/eng/doc/CrossPlatformCryptography.md
+++ b/eng/doc/CrossPlatformCryptography.md
@@ -66,7 +66,9 @@ This section includes the following packages:
 | HMAC                      | ✔️ <sup>3</sup>   | ✔️ <sup>3</sup>    |
 
 <sup>1</sup>Available starting in Microsoft Go 1.24.
+
 <sup>2</sup>Requires OpenSSL 1.1.1 or later.
+
 <sup>3</sup>The supported hash algorithms are the same as the ones supported as standalone hash functions.
 
 ## Symmetric encryption
@@ -144,6 +146,7 @@ This section includes the following packages:
 | PKCS1v15 Signature (SHA-3)        | ❌                   | ❌                 |
 
 <sup>1</sup>The supported hash algorithms are the same as the ones supported as standalone hash functions.
+
 <sup>2</sup>Available starting in Microsoft Go 1.24.
 
 #### RSA key sizes
--- a/eng/doc/CrossPlatformCryptography.md
+++ b/eng/doc/CrossPlatformCryptography.md
@@ -30,11 +30,11 @@ SCOSSL is expected to be used with the default built-in provider enabled as a fa
 
 The following legend describes the symbols used in the tables to indicate the level of support for each cryptographic algorithm:
 
-| Symbol | Meaning                                                                                                            |
-|--------|--------------------------------------------------------------------------------------------------------------------|
-| ✔️     | Supported, possibly with minor limitations that don't require user action when using the latest Go and OS versions |
-| ⚠️     | Supported with limitations that require user action                                                                |
-| ❌     | Not supported                                                                                                      |
+| Symbol | Meaning                                                                                                                      |
+|--------|------------------------------------------------------------------------------------------------------------------------------|
+| ✔️     | Supported, possibly with minor limitations that don't require special configuration when using the latest Go and OS versions |
+| ⚠️     | Supported with limitations that require special configuration action                                                         |
+| ❌     | Not supported                                                                                                                |
 
 When an algorithm is not supported or the limitations are exceeded, Microsoft Go will fall back to the Go implementation.
 
@@ -159,7 +159,9 @@ This section includes the following packages:
 
 Multi-prime RSA keys are not supported.
 
-The RSA key size is subject to the limitations of the underlying cryptographic library. For example, on Windows and when using SCOSSL, the key size should be multiple of 8.
+The RSA key size is subject to the limitations of the underlying cryptographic library.
+For example, on some Windows and SCOSSL configurations, the key size should be multiple of 8.
+Please refer to the documentation of the underlying cryptographic library for the specific limitations.
 
 #### PSS salt length
 
--- a/eng/doc/CrossPlatformCryptography.md
+++ b/eng/doc/CrossPlatformCryptography.md
@@ -113,6 +113,9 @@ This section includes the following packages:
 This section includes the following subsections:
 
 * [RSA](#rsa)
+* [ECDSA](#ecdsa)
+* [ECDH](#ecdh)
+* [DSA](#dsa)
 
 ### RSA
 
@@ -131,27 +134,99 @@ This section includes the following packages:
 | PSS (SHA-2)                       | ✔️ <sup>1</sup>      | ✔️ <sup>1</sup>    |
 | PSS (SHA-3)                       | ❌                   | ❌                 |
 | PKCS1v15 Signature (Unhashed)     | ✔️                   | ✔️                 |
-| PKCS1v15 Signature (RIPMED160)    | ❌                   | ✔️                 |
-| PKCS1v15 Signature (MD4)          | ❌                   | ✔️                 |
+| PKCS1v15 Signature (RIPMED160)    | ❌ <sup>2</sup>      | ✔️ <sup>2</sup>    |
+| PKCS1v15 Signature (MD4)          | ❌ <sup>2</sup>      | ✔️ <sup>2</sup>    |
 | PKCS1v15 Signature (MD5)          | ✔️                   | ✔️                 |
-| PKCS1v15 Signature (MD5-SHA1)     | ✔️                   | ✔️                 |
+| PKCS1v15 Signature (MD5-SHA1)     | ✔️ <sup>2</sup>      | ✔️ <sup>2</sup>    |
 | PKCS1v15 Signature (SHA-1)        | ✔️                   | ✔️                 |
 | PKCS1v15 Signature (SHA-2)        | ✔️ <sup>1</sup>      | ✔️ <sup>1</sup>    |
 | PKCS1v15 Signature (SHA-3)        | ❌                   | ❌                 |
 
 <sup>1</sup>The supported hash algorithms are the same as the ones supported as standalone hash functions.
+<sup>2</sup>Available starting in Microsoft Go 1.24.
 
 #### RSA key sizes
 
 [rsa.GenerateKey](https://pkg.go.dev/crypto/rsa#GenerateKey) only supports the following key sizes (in bits): 2048, 3072, 4096.
 
-Multiprime RSA keys are not supported.
+Multi-prime RSA keys are not supported.
+
+The RSA key size is subject to the limitations of the underlying cryptographic library. For example, on Windows and when using SCOSSL, the key size should be multiple of 8.
 
 #### PSS salt length
 
 On Windows, when verifying a PSS signature, [rsa.PSSSaltLengthAuto](https://pkg.go.dev/crypto/rsa#pkg-constants) is not supported.
 
-### Random number generation
+#### Random number generation
 
-For those padding modes that require random numbers, only the [rand.Reader](https://pkg.go.dev/crypto/rand#Reader) is supported.
+For those operations that require random numbers, only the [rand.Reader](https://pkg.go.dev/crypto/rand#Reader) is supported.
 
+### ECDSA
+
+This section includes the following packages:
+
+* [crypto/ecdsa](https://pkg.go.dev/crypto/ecdsa)
+* [crypto/elliptic](https://pkg.go.dev/crypto/elliptic)
+
+| Elliptic Curve            | Windows     | Linux        |
+|---------------------------|-------------|--------------|
+| NIST P-224 (secp224r1)    | ✔️          | ✔️          |
+| NIST P-256 (secp256r1)    | ✔️          | ✔️          |
+| NIST P-384 (secp384r1)    | ✔️          | ✔️          |
+| NIST P-521 (secp521r1)    | ✔️          | ✔️          |
+
+#### Random number generation
+
+For those operations that require random numbers, only the [rand.Reader](https://pkg.go.dev/crypto/rand#Reader) is supported.
+
+### ECDH
+
+This section includes the following packages:
+
+* [crypto/ecdh](https://pkg.go.dev/crypto/ecdsa)
+
+| Elliptic Curve            | Windows     | Linux        |
+|---------------------------|-------------|--------------|
+| NIST P-224 (secp224r1)    | ✔️          | ✔️          |
+| NIST P-256 (secp256r1)    | ✔️          | ✔️          |
+| NIST P-384 (secp384r1)    | ✔️          | ✔️          |
+| NIST P-521 (secp521r1)    | ✔️          | ✔️          |
+| X25519 (curve25519)       | ❌          | ❌          |
+
+#### Random number generation
+
+For those operations that require random numbers, only the [rand.Reader](https://pkg.go.dev/crypto/rand#Reader) is supported.
+
+### DSA
+
+| Parameters    | Windows     | Linux        |
+|---------------|-------------|--------------|
+| L1024N160     | ✔️          | ✔️          |
+| L2048N224     | ❌          | ✔️          |
+| L2048N256     | ✔️          | ✔️          |
+| L3072N256     | ✔️          | ✔️          |
+
+## KDF
+
+This section includes the following packages:
+
+* [crypto/hkdf](https://pkg.go.dev/crypto/hkdf)
+* [crypto/pbkdf2](https://pkg.go.dev/crypto/pbkdf2)
+
+| Functions     | Windows          | Linux             |
+|---------------|------------------|-------------------|
+| PBKDF2        | ✔️ <sup>1</sup>  | ✔️ <sup>1</sup>  |
+| HKDF          | ✔️ <sup>1</sup>  | ✔️ <sup>1</sup>  |
+
+<sup>1</sup>The supported hash algorithms are the same as the ones supported as standalone hash functions.
+
+## ML-KEM
+
+This section includes the following packages:
+
+* [crypto/mlkem](https://pkg.go.dev/crypto/mlkem)
+
+| Parameters    | Windows     | Linux        |
+|---------------|-------------|--------------|
+|768            | ❌          | ❌          |
+|1024           | ❌          | ❌          |
--- a/eng/doc/CrossPlatformCryptography.md
+++ b/eng/doc/CrossPlatformCryptography.md
@@ -14,6 +14,26 @@ This article identifies the features that are supported on each platform.
 
 This article assumes you have a working familiarity with cryptography in Go.
 
+## Platform support
+
+Microsoft Go supports the following platforms:
+
+### Windows
+
+On Windows, Microsoft Go uses the [Cryptography API: Next Generation](https://learn.microsoft.com/en-us/windows/win32/seccng/cng-portal) library, CNG from now on, for cryptographic operations.
+CNG is available since Windows Vista and Windows Server 2008 and it doesn't require any additional installation nor configuration.
+
+### Linux
+
+On Linux, Microsoft Go uses the [OpenSSL crypto library](https://docs.openssl.org/3.0/man7/crypto/) library, OpenSSL from now on, for cryptographic operations.
+OpenSSL is normally available on Linux distributions, but it may not be installed by default.
+If it is not installed, you can install it using the package manager of your distribution.
+
+OpenSSL 3 implements all the cryptographic algorithms using [Providers](https://docs.openssl.org/3.0/man7/crypto/#providers).
+Microsoft Go officially supports the built-in providers and the [SymCrypt provider](https://github.com/microsoft/SymCrypt-OpenSSL), SCOSSL from now on.
+The minimum SCOSSL version required is v1.6.1.
+The following tables assume that the SCOSSL provider is used together with the built-in providers.
+
 ## Hash and Message Authentication Algorithms
 
 This section includes the following packages:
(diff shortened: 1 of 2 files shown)


gdams (approved):

eng/doc/CrossPlatformCryptography.md
```diff
@@ -0,0 +1,336 @@
+# Cross-Platform Cryptography in Microsoft Go
+
+Cryptographic operations in Microsoft Go are done by operating system (OS) libraries. This dependency has advantages:
```
gdams:
```suggestion
Cryptographic operations in Microsoft Go are delegated to the operating system (OS) libraries. This dependency has advantages:
```

eng/doc/CrossPlatformCryptography.md
```diff
@@ -0,0 +1,336 @@
...
+Microsoft Go supports the following platforms:
+
+### Windows
+
+On Windows, Microsoft Go uses the [Cryptography API: Next Generation](https://learn.microsoft.com/en-us/windows/win32/seccng/cng-portal) library, CNG from now on, for cryptographic operations.
+CNG is available since Windows Vista and Windows Server 2008 and it doesn't require any additional installation nor configuration.
+
+### Linux
+
+On Linux, Microsoft Go uses the [OpenSSL crypto library](https://docs.openssl.org/3.0/man7/crypto/) library, OpenSSL from now on, for cryptographic operations.
```
gdams:
```suggestion
On Linux, Microsoft Go uses the [OpenSSL crypto library](https://docs.openssl.org/3.0/man7/crypto/), OpenSSL from now on, for cryptographic operations.
```

eng/doc/CrossPlatformCryptography.md
```diff
@@ -0,0 +1,336 @@
+# Cross-Platform Cryptography in Microsoft Go
+
+Cryptographic operations in Microsoft Go are delegated to the operating system (OS) libraries. This dependency has advantages:
```
dagood: Not sure about the blanket statement here... the section goes on to clarify, but it could come off as contradiction (or just wrong, if the reader skips ahead).

```suggestion
Cryptographic operations in Microsoft Go are delegated to the operating system (OS) libraries in some conditions. This dependency has advantages:
```

eng/doc/CrossPlatformCryptography.md
```diff
@@ -0,0 +1,336 @@
...
+
+* Go apps benefit from OS reliability. Keeping cryptography libraries safe from vulnerabilities is a high priority for OS vendors. To do that, they provide updates that system administrators should be applying.
```
dagood: Some people will read between the lines and see this as implying "OS reliability vs. vuln > Go crypto reliability vs. vuln", which AFAIK we don't intend.

I think the point is that an otherwise-unpatched Go app might be safe vs. some very specific crypto vulnerabilities. A lot of Go folks use containers and tend to be used to rebuilding/republishing every time there's *any* vuln, and system administrators wouldn't meddle with live containers, so I don't think this is even considered a benefit for how they currently deploy. IMO this needs to be laid out more clearly, especially pointing out the group (/deployment strategy) who actually gets an advantage from this approach.

eng/doc/CrossPlatformCryptography.md
```diff
@@ -0,0 +1,336 @@
...
+* Go apps have access to FIPS-validated algorithms if the OS libraries are FIPS-validated.
+
+> [!NOTE]
+> Starting with Go 1.24, Go will also be FIPS 140-3 compliant, see https://github.com/golang/go/issues/69536.
+> If the only reason you are using Microsoft Go is for FIPS 140-3 compliance, you should consider using Microsoft Go 1.24 or later.
```
dagood: This note seems a bit odd... if you are looking for FIPS 140-3 compliance, my intuition would be that you *must* use 1.24 or later?

My brain's autocorrect points me in another direction being the actual intent--to consider using upstream:

> If the only reason you are using Microsoft Go is for FIPS 140-3 compliance (not e.g. to conform to Microsoft policy), consider using upstream Go 1.24 or later.

Maybe there's just some missing context to connect how Microsoft Go, upstream Go, and "spec"-like Go 1.24 all relate to FIPS 140-3.

eng/doc/CrossPlatformCryptography.md
```diff
@@ -0,0 +1,336 @@
...
+
+Go apps will fall back to native Go implementations if the OS libraries doesn't support the algorithm.
```
dagood:
```suggestion
Go apps will fall back to native Go implementations if the OS libraries don't support the algorithm.
```

eng/doc/CrossPlatformCryptography.md
```diff
@@ -0,0 +1,336 @@
...
+
+This article assumes you have a working familiarity with cryptography in Go.
+
+## Platform support
+
+Microsoft Go supports the following platforms:
+
+### Windows
+
+On Windows, Microsoft Go uses the [Cryptography API: Next Generation](https://learn.microsoft.com/en-us/windows/win32/seccng/cng-portal) library, CNG from now on, for cryptographic operations.
```
dagood: Simplify term introduction:

```suggestion
On Windows, Microsoft Go uses the [CNG library (Cryptography API: Next Generation)](https://learn.microsoft.com/en-us/windows/win32/seccng/cng-portal) for cryptographic operations.
```

eng/doc/CrossPlatformCryptography.md
```diff
@@ -0,0 +1,336 @@
...
+CNG is available since Windows Vista and Windows Server 2008 and it doesn't require any additional installation nor configuration.
+
+### Linux
+
+On Linux, Microsoft Go uses the [OpenSSL crypto library](https://docs.openssl.org/3.0/man7/crypto/), OpenSSL from now on, for cryptographic operations.
```
dagood: Simplify term introduction:

```suggestion
On Linux, Microsoft Go uses the [OpenSSL crypto library](https://docs.openssl.org/3.0/man7/crypto/) for cryptographic operations.
```

eng/doc/CrossPlatformCryptography.md
```diff
@@ -0,0 +1,336 @@
...
+OpenSSL is normally available on Linux distributions, but it may not be installed by default.
+If it is not installed, you can install it using the package manager of your distribution.
+
+OpenSSL 3 implements all the cryptographic algorithms using [Providers](https://docs.openssl.org/3.0/man7/crypto/#providers).
+Microsoft Go officially supports the built-in providers and the [SymCrypt provider](https://github.com/microsoft/SymCrypt-OpenSSL), SCOSSL from now on.
```
dagood: Simplify term introduction:

```suggestion
Microsoft Go officially supports the built-in providers and [SCOSSL (SymCrypt provider for OpenSSL)](https://github.com/microsoft/SymCrypt-OpenSSL).
```

eng/doc/CrossPlatformCryptography.md
```diff
@@ -0,0 +1,336 @@
...
+The minimum SCOSSL version required is v1.6.1.
+The following tables assume that the SCOSSL provider is used together with the built-in providers.
```
dagood: This might be worth a little more explanation, and/or maybe more straightforward language than "used together with", maybe the built in providers are "also enabled" or something?

eng/doc/CrossPlatformCryptography.md
```diff
@@ -0,0 +1,336 @@
...
+| AES-CFB       | ❌       | ❌              |
+| AES-OFB       | ❌       | ❌              |
+| AES-GCM       | ✔️       | ✔️              |
+| DES-CBC       | ✔️       | ⚠️<sup>1</sup>  |
+| DES-ECB       | ✔️       | ⚠️<sup>1</sup>  |
+| 3DES-ECB      | ✔️       | ✔️              |
+| 3DES-CBC      | ✔️       | ✔️              |
+| RC4           | ✔️       | ⚠️<sup>1</sup>  |
+
+<sup>1</sup>When using OpenSSL 3, requires the legacy provider to be enabled.
```
dagood, line 101: Might be worth explaining ✔️ vs. ⚠️ in a key/legend. Not sure why this deserves ⚠️ but "Available starting in Microsoft Go 1.24." is an ordinary ✔️ above.

eng/doc/CrossPlatformCryptography.md
```diff
@@ -0,0 +1,336 @@
...
+
+<sup>2</sup>Available starting in Microsoft Go 1.24.
+
+#### RSA key sizes
+
+[rsa.GenerateKey](https://pkg.go.dev/crypto/rsa#GenerateKey) only supports the following key sizes (in bits): 2048, 3072, 4096.
+
+Multi-prime RSA keys are not supported.
+
+The RSA key size is subject to the limitations of the underlying cryptographic library. For example, on Windows and when using SCOSSL, the key size should be multiple of 8.
```
dagood: Is it worth being more exact? Are there more potential situations than this one example?

```suggestion
The RSA key size is subject to the limitations of the underlying cryptographic library:

* On Windows or when using SCOSSL, the key size must be a multiple of 8 (in bits).
```

But now I wonder if this was intended to be saying something about algorithms implemented by SCOSSL vs. by the built-in OpenSSL provider...

eng/doc/CrossPlatformCryptography.md
```diff
@@ -0,0 +1,336 @@
...
+| Elliptic Curve            | Windows     | Linux        |
+|---------------------------|-------------|--------------|
+| NIST P-224 (secp224r1)    | ✔️          | ✔️          |
+| NIST P-256 (secp256r1)    | ✔️          | ✔️          |
+| NIST P-384 (secp384r1)    | ✔️          | ✔️          |
+| NIST P-521 (secp521r1)    | ✔️          | ✔️          |
+
+#### Random number generation
+
+For those operations that require random numbers, only the [rand.Reader](https://pkg.go.dev/crypto/rand#Reader) is supported.
```
dagood: This could probably be refined and more specific to Go:

```suggestion
Operations that require random numbers (rand io.Reader) only support [rand.Reader](https://pkg.go.dev/crypto/rand#Reader).
```

(A few repeats of this.)

eng/doc/CrossPlatformCryptography.md
```diff
@@ -0,0 +1,336 @@
...
+| SHAKE-256                 | ❌                 | ❌                 |
+| CSHAKE-128                | ❌                 | ❌                 |
+| CSHAKE-256                | ❌                 | ❌                 |
+| HMAC<sup>3</sup>          | ✔️                 | ✔️                 |
+
+<sup>1</sup>Available starting in Microsoft Go 1.24.
+
+<sup>2</sup>Requires OpenSSL 1.1.1 or later.
+
+<sup>3</sup>The supported hash algorithms are the same as the ones supported as standalone hash functions.
```
dagood: The structure of the start of this sentence could match "available" and "requires" better, for better flow:

```suggestion
<sup>3</sup>Supports only hash algorithms that are supported as standalone hash functions.
```

When it's repeated in other sections, a link might also be nice:

```suggestion
<sup>3</sup>Supports only hash algorithms that are [supported as standalone hash functions](#hash-and-message-authentication-algorithms).
```

eng/doc/CrossPlatformCryptography.md
```diff
@@ -0,0 +1,336 @@
...
+## ML-KEM
+
+This section includes the following packages:
+
+* [crypto/mlkem](https://pkg.go.dev/crypto/mlkem)
+
+| Parameters    | Windows     | Linux        |
+|---------------|-------------|--------------|
+|768            | ❌          | ❌          |
+|1024           | ❌          | ❌          |
```
dagood: Formatting nit:

```suggestion
| 768           | ❌          | ❌          |
| 1024          | ❌          | ❌          |
```

eng/doc/CrossPlatformCryptography.md
qmuntal: I intentionally didn't want to enumerate a possible limitations, as can vary in time and are normally not part of the public documentation of the crypto libraries.

eng/doc/CrossPlatformCryptography.md
qmuntal: Good catch, the last sentence should mention upstream Go, no Microsoft Go 😅. Will fix + add some clarification.

eng/doc/CrossPlatformCryptography.md
qmuntal: Hmm, good point. I'll remove the advantages bullets and the subsequent NOTE altogether. It is confusing and of of scope for this documents.

eng/doc/CrossPlatformCryptography.md
qmuntal: Done.

eng/doc/CrossPlatformCryptography.md
qmuntal: Done!

eng/doc/CrossPlatformCryptography.md
```diff
@@ -0,0 +1,340 @@
+# Cross-Platform Cryptography in Microsoft Go
+
+Cryptographic operations in Microsoft Go are delegated to the operating system (OS) libraries in some conditions.
+
+Go apps will fall back to native Go implementations if the OS libraries don't support the algorithm.
```
dagood: Seeing this again, it feels a bit barer than it needs to be... I think a link and merging the paragraph helps make the context clearer. (A few more tweaks for flow, too.)

```suggestion
Cryptographic operations in Microsoft Go are delegated to the operating system (OS) libraries in some conditions described.
The high level conditions and the benefits of delegating cryptographic operations are described in the [Microsoft Go FIPS README](./Readme.md).
At a fine-grained level, Go apps will fall back to the native Go implementation of an algorithm if the OS libraries don't support it.
```

eng/doc/CrossPlatformCryptography.md
```diff
@@ -20,19 +13,30 @@ Microsoft Go supports the following platforms:
...
+On Linux, Microsoft Go uses the [OpenSSL crypto library](https://docs.openssl.org/3.0/man7/crypto/) for cryptographic operations.
 OpenSSL is normally available on Linux distributions, but it may not be installed by default.
 If it is not installed, you can install it using the package manager of your distribution.
 
 OpenSSL 3 implements all the cryptographic algorithms using [Providers](https://docs.openssl.org/3.0/man7/crypto/#providers).
-Microsoft Go officially supports the built-in providers and the [SymCrypt provider](https://github.com/microsoft/SymCrypt-OpenSSL), SCOSSL from now on.
-The minimum SCOSSL version required is v1.6.1.
-The following tables assume that the SCOSSL provider is used together with the built-in providers.
+Microsoft Go officially supports the built-in providers and [SCOSSL (SymCrypt provider for OpenSSL)](https://github.com/microsoft/SymCrypt-OpenSSL) v1.6.1 or later.
+SCOSSL is expected to be used with the default built-in provider enable as a fallback (which is the case when using [Azure Linux 3](https://github.com/microsoft/AzureLinux)).
```
dagood:
```suggestion
SCOSSL is expected to be used with the default built-in provider enabled as a fallback (which is the case when using [Azure Linux 3](https://github.com/microsoft/AzureLinux)).
```

eng/doc/CrossPlatformCryptography.md
```diff
@@ -20,19 +13,30 @@ Microsoft Go supports the following platforms:
...
+
+## Table legend
+
+The following table legend is used to indicate the level of support for each cryptographic algorithm:
```
dagood: For some reason it seems odd to say a table uses a legend--they both use the same symbols, but the tables don't really "use" the legend directly, to me.

```suggestion
The following legend describes the symbols used in the tables to indicate the level of support for each cryptographic algorithm:
```

eng/doc/CrossPlatformCryptography.md
```diff
@@ -20,19 +13,30 @@ Microsoft Go supports the following platforms:
...
+
+| Symbol | Meaning                                                                                                            |
+|--------|--------------------------------------------------------------------------------------------------------------------|
+| ✔️     | Supported, possibly with minor limitations that don't require user action when using the latest Go and OS versions |
+| ⚠️     | Supported with limitations that require user action                                                                |
```
dagood: This is the only time the doc mentions "user", and I think without a title like "FIPS 140-2 User Guide", it might be misinterpreted to mean that the dev's program will require some input from their end users. (Which doesn't make sense vs. a correct interpretation, but seems worth clarifying.)

Maybe the right fix is to actually update the intro to mention that this is a user guide for cryptography in Microsoft Go, to make "user" contextualized throughout the whole doc. (So maybe this can wait until this doc fully takes the place of the user guide.)

eng/doc/CrossPlatformCryptography.md
```diff
@@ -20,19 +13,30 @@ Microsoft Go supports the following platforms:
...
+| ❌     | Not supported                                                                                                      |
+
+When an algorithm is not supported or the limitations are not met, Microsoft Go will fall back to the Go implementation.
```
dagood:
```suggestion
When an algorithm is not supported or the limitations are exceeded, Microsoft Go will fall back to the Go implementation.
```

eng/doc/CrossPlatformCryptography.md
dagood: Ah, makes sense. I think the doc probably should state that variability pretty clearly, because otherwise it might be a bit confusing why more detail isn't provided. Also, it sounds like the user is meant to get a hint that they should avoid depending on exact details here.

eng/doc/CrossPlatformCryptography.md
```diff
@@ -0,0 +1,340 @@
...
+
+The RSA key size is subject to the limitations of the underlying cryptographic library. For example, on Windows and when using SCOSSL, the key size should be multiple of 8.
+
+#### PSS salt length
+
+On Windows, when verifying a PSS signature, [rsa.PSSSaltLengthAuto](https://pkg.go.dev/crypto/rsa#pkg-constants) is not supported.
+
+#### Random number generation
+
+For those operations that require random numbers, only the [rand.Reader](https://pkg.go.dev/crypto/rand#Reader) is supported.
```
dagood: (Applying this suggestion to more places.)

```suggestion
Operations that require random numbers (rand io.Reader) only support [rand.Reader](https://pkg.go.dev/crypto/rand#Reader).
```

eng/doc/CrossPlatformCryptography.md
```diff
@@ -0,0 +1,340 @@
...
+|---------------------------|-------------|--------------|
+| NIST P-224 (secp224r1)    | ✔️          | ✔️          |
+| NIST P-256 (secp256r1)    | ✔️          | ✔️          |
+| NIST P-384 (secp384r1)    | ✔️          | ✔️          |
+| NIST P-521 (secp521r1)    | ✔️          | ✔️          |
+| X25519 (curve25519)       | ❌          | ❌          |
+
+#### Random number generation
+
+For those operations that require random numbers, only the [rand.Reader](https://pkg.go.dev/crypto/rand#Reader) is supported.
```
dagood:
```suggestion
Operations that require random numbers (rand io.Reader) only support [rand.Reader](https://pkg.go.dev/crypto/rand#Reader).
```

eng/doc/CrossPlatformCryptography.md
```diff
@@ -0,0 +1,340 @@
...
+
+| Schemes     | Windows    | Linux         |
+|-------------|------------|---------------|
+| Ed25519     | ❌         | ✔️           |
+| Ed25519ctx  | ❌         | ❌           |
+| Ed25519ph   | ❌         | ❌           |
+
+#### Random number generation
+
+For those operations that require random numbers, only the [rand.Reader](https://pkg.go.dev/crypto/rand#Reader) is supported.
```
dagood:
```suggestion
Operations that require random numbers (rand io.Reader) only support [rand.Reader](https://pkg.go.dev/crypto/rand#Reader).
```

eng/doc/CrossPlatformCryptography.md
```diff
@@ -0,0 +1,340 @@
...
+| PKCS1v15 Signature (Unhashed)          | ✔️                   | ✔️                 |
+| PKCS1v15 Signature (RIPMED160)         | ❌                   | ✔️<sup>2</sup>     |
+| PKCS1v15 Signature (MD4)               | ❌                   | ✔️<sup>2</sup>     |
+| PKCS1v15 Signature (MD5)               | ✔️                   | ✔️                 |
+| PKCS1v15 Signature (MD5-SHA1)          | ✔️<sup>2</sup>       | ✔️<sup>2</sup>     |
+| PKCS1v15 Signature (SHA-1)             | ✔️                   | ✔️                 |
+| PKCS1v15 Signature (SHA-2)<sup>1</sup> | ✔️                   | ✔️                 |
+| PKCS1v15 Signature (SHA-3)             | ❌                  | ❌                 |
+
+<sup>1</sup>The supported hash algorithms are the same as the ones supported as standalone hash functions.
```
dagood: Applying this suggestion to other places and adding in a link since it's in a different section.

```suggestion
<sup>1</sup>Supports only hash algorithms that are [supported as standalone hash functions](#hash-and-message-authentication-algorithms).
```

eng/doc/CrossPlatformCryptography.md
```diff
@@ -0,0 +1,340 @@
...
+
+* [crypto/hkdf](https://pkg.go.dev/crypto/hkdf)
+* [crypto/pbkdf2](https://pkg.go.dev/crypto/pbkdf2)
+
+| Functions     | Windows          | Linux             |
+|---------------|------------------|-------------------|
+| PBKDF2        | ✔️ <sup>1</sup>  | ✔️ <sup>1</sup>  |
+| HKDF          | ✔️ <sup>1</sup>  | ✔️ <sup>1</sup>  |
+
+<sup>1</sup>The supported hash algorithms are the same as the ones supported as standalone hash functions.
```
dagood:
```suggestion
<sup>1</sup>Supports only hash algorithms that are [supported as standalone hash functions](#hash-and-message-authentication-algorithms).
```

dagood (approved):
Nice!

eng/doc/CrossPlatformCryptography.md
qmuntal: Good point. I've replaces `user action` for `special configuration`, which is more specific.

eng/doc/CrossPlatformCryptography.md
qmuntal: I've rephrased the paragraph to be less confusing.

gdams (approved):

PR #1477, opened 2025-01-13
Centralize crypto backends GOEXPERIMENT definitions

qmuntal:
This PR moves all the crypto backend GOEXPERIMENT definitions to the first patch. This changes goes in the line of having simpler and easier to update patch files. Particularly, the first patch file already had knowledge about all the supported crypto backends, so there is not point on splitting the logic in different patch files.

For #1478.


gdams (approved):

patches/0001-Add-systemcrypto-GOEXPERIMENT.patch
```diff
@@ -3,7 +3,7 @@ From: Davis Goodin <dagood@microsoft.com>
 Date: Wed, 31 May 2023 16:54:31 -0500
 Subject: [PATCH] Add systemcrypto GOEXPERIMENT
```
dagood: Update subject/filename to "Add crypto backend GOEXPERIMENTs"? (Maybe later if conflicts are a concern.)

dagood (approved):

patches/0001-Add-systemcrypto-GOEXPERIMENT.patch
qmuntal: Good point. Done.

PR #1488, opened 2025-01-15
darwin: add macOS to CrossPlatformCryptography document

gdams:
Documents which algorithms are supported for macOS.

I used prettier to format the markdown better

### DIFF
--- a/eng/doc/CrossPlatformCryptography.md
+++ b/eng/doc/CrossPlatformCryptography.md
@@ -132,9 +132,6 @@ This section includes the following subsections:
 
 ### RSA
 
-> [!WARNING]  
-> System libraries might not fully implement all RSA features. Please see the subsections below for more information.
-
 This section includes the following packages:
 
 - [crypto/rsa](https://pkg.go.dev/crypto/rsa)
--- a/eng/doc/CrossPlatformCryptography.md
+++ b/eng/doc/CrossPlatformCryptography.md
@@ -131,6 +131,9 @@ This section includes the following subsections:
 
 ### RSA
 
+> [!WARNING]  
+> System Libraries might not fully implement all RSA features. Please see the subsections below for more information.
+
 This section includes the following packages:
 
 - [crypto/rsa](https://pkg.go.dev/crypto/rsa)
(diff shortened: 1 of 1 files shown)


qmuntal (approved):

eng/doc/CrossPlatformCryptography.md
```diff
@@ -26,12 +26,17 @@ OpenSSL 3 implements all the cryptographic algorithms using [Providers](https://
 Microsoft Go officially supports the built-in providers and [SCOSSL (SymCrypt provider for OpenSSL)](https://github.com/microsoft/SymCrypt-OpenSSL) v1.6.1 or later.
 SCOSSL is expected to be used with the default built-in provider enabled as a fallback (which is the case when using [Azure Linux 3](https://github.com/microsoft/AzureLinux)).
 
+## macOS
+
+On macOS, Microsoft Go uses [CommonCrypto](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/Common%20Crypto.3cc.html) and [CryptoKit](https://developer.apple.com/documentation/cryptokit) for cryptographic operations.
+CommonCrypto and CryptoKit is shipped with macOS and it doesn't require any additional installation nor configuration. Currently macOS 13 and above is supported.
```
dagood:
```suggestion
CommonCrypto and CryptoKit are shipped with macOS and don't require any additional installation nor configuration.
Currently macOS 13 and above is supported.
```

eng/doc/CrossPlatformCryptography.md
```diff
@@ -79,83 +84,88 @@ This section includes the following packages:
...
+- [RSA](#rsa)
+- [ECDSA](#ecdsa)
+- [ECDH](#ecdh)
+- [Ed25519](#ed25519)
+- [DSA](#dsa)
 
 ### RSA
 
+> [!WARNING]  
+> System Libraries might not fully implement all RSA features. Please see the subsections below for more information.
```
dagood: Why a warning for RSA in particular? Many of the tables have feature gaps explained below them.

Also:

```suggestion
> System libraries might not fully implement all RSA features. Please see the subsections below for more information.
```

eng/doc/CrossPlatformCryptography.md
```diff
@@ -193,150 +205,158 @@ Operations that require random numbers (rand io.Reader) only support [rand.Reade
...
+
+Below are the supported [`tls.SignatureSchemes`](https://pkg.go.dev/crypto/tls#SignatureScheme).
+
+| Name                   | Windows | Linux | macOS |
+| ---------------------- | ------- | ----- | ----- |
+| PKCS1WithSHA1          | ✔️      | ✔️    | ✔️    |
+| PKCS1WithSHA256        | ✔️      | ✔️    | ✔️    |
+| PKCS1WithSHA384        | ✔️      | ✔️    | ✔️    |
+| PKCS1WithSHA512        | ✔️      | ✔️    | ✔️    |
+| PSSWithSHA1            | ❌      | ❌    | ✔️    |
```
dagood: PSSWithSHA1 doesn't show up in the TLS package, where's it from?

eng/doc/CrossPlatformCryptography.md
```diff
@@ -79,83 +85,88 @@ This section includes the following packages:
...
+- [RSA](#rsa)
+- [ECDSA](#ecdsa)
+- [ECDH](#ecdh)
+- [Ed25519](#ed25519)
+- [DSA](#dsa)
 
 ### RSA
 
+> [!WARNING]  
+> System libraries might not fully implement all RSA features. Please see the subsections below for more information.
```
dagood: Why a warning for RSA in particular? Many of the tables have feature gaps explained below them.

(Reposting https://github.com/microsoft/go/pull/1488#discussion_r1917150934, I suppose I should have split it into two comments.)

gdams: hmm good point, maybe it's implied (or perhaps we should have this warning at the top of the document)?

dagood: It sounds like the main concern here is that it's not necessarily clear that there's more info below the table--in that case I think we should use numbered notes more consistently and bring global-ish info above the table:
* https://github.com/microsoft/go/pull/1507

dagood: More generally, I think these parts establish pretty well that features have limitations:

> At a fine-grained level, Go apps will fall back to the native Go implementation of an algorithm if the OS libraries don't support it. This article identifies the features that are supported on each platform.

> When an algorithm is not supported or the limitations are exceeded, Microsoft Go will fall back to the Go implementation.

Issue #1478, opened 2025-01-13
Make patch files easier to maintain


qmuntal:
The current way of structuring our patch files makes it difficult to maintain and extend our Go fork:

- Upgrading the OpenSSL backend module causes conflicts with the CNG patch file which are difficult to fix.
- The logic to hook the different algorithms is split between different patch files instead of having it centralized in a single patch.
- The likelihood if having conflicts when working on different PRs are high, mostly in the patch summary.

qmuntal:
Tentatively marked as done. Will reopen if we see that the new improvements are not enough.

PR #1489, opened 2025-01-15
Merge backend patch files

qmuntal:
Adding a new function or type to our backends is tedious and error prone, as one have to modify 5! patch files.

It is nice to have each backend in its own patch file, but at the end it doesn't pay off.

This PR merged them all together into a single patch file that is in charge of implementing the while `crypto/internal/backend` package.

For https://github.com/microsoft/go/issues/1478.


gdams (approved):

dagood (approved):
> It is nice to have each backend in its own patch file, but at the end it doesn't pay off.

I think at first, we (or maybe just I) envisioned that we'd make more backend-specific changes/fixes against an essentially constant set of algorithms and "interface", but it's ended up that we do a *lot* of cross-backend changes. Agreed that putting them together makes more sense.

karianna (approved):

PR #1494, opened 2025-01-16
Prefer `GODEBUG=fips140` over `GOFIPS` and don't call `openssl.SetFIPS`

qmuntal:
This PR implements https://github.com/microsoft/go-lab/blob/main/docs/adr/0012-remove-gofips.md.

### DIFF
--- a/eng/_util/cmd/run-builder/run-builder.go
+++ b/eng/_util/cmd/run-builder/run-builder.go
@@ -147,7 +147,7 @@ func main() {
 		}
 
 		if *fipsMode {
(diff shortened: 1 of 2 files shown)


gdams:
seeing this failure:

```go
---- Running command: [go/bin/go tool dist test]
2025/01/16 11:55:49 Using fallback (no-op) for enableSystemWideFIPS. It either isn't supported on this platform or isn't necessary.
panic: FIPS mode requested (environment variable GODEBUG=fips140=1) but no supported crypto backend is enabled
```

Issue #1445, opened 2024-12-16
Map `GOFIPS` to upstream `GODEBUG=fips140`


qmuntal:
Go 1.24 will ship with a mechanism to enable FIPS 140 at runtime, similar to our `GOFIPS` setting: https://github.com/golang/go/issues/70123#issuecomment-2490017587.

We should investigate how to map `GOFIPS` to `GODEBUG=fips140`.

gdams:
Blog post is out https://devblogs.microsoft.com/go/go-1-24-fips-update, closing

PR #1496, opened 2025-01-16
Don't try to override system FIPS mode and use openssl.FIPSCapable

qmuntal:
As agreed in https://github.com/microsoft/go-lab/blob/main/docs/adr/0012-remove-gofips.md, we shouldn't try to modify the OpenSSL FIPS mode.

Azure Linux 3 doesn't set `fips=yes` in the default properties, even when running in kernel FIPS mode. They omit it because the default OpenSSL provider, SCOSSL, is already FIPS compliant, and not setting it allows some non-FIPS algorithms to fall back to the built-in provider (which is also enabled by default). This means we can't use `openssl.FIPS()` (which checks if `fips=yes` is set) in AZL3. Fortunately we can use `openssl.FIPSCapable()` instead, which doesn't check  if `fips=yes` is set.

This has the nice benefit of being more compliant when running in FIPS mode on AZL3, as more non-FIPS approved algorithms (like SHA1) will be provided by the built-in OpenSSL providers instead of falling back to Go.

For #1445.


gdams (approved):

dagood (approved):
LGTM, in particular the non-symmetry between using `openssl.FIPSCapable` and `openssl.FIPS` does make sense here to me with your comments. 👍

PR #1503, opened 2025-01-17
Prefer GODEBUG=fips140 over GOFIPS

qmuntal:
This PR makes our backends to fail if `GODEBUG=fips140=1` is set but FIPS mode is not enabled, as agreed in https://github.com/microsoft/go-lab/blob/main/docs/adr/0012-remove-gofips.md.

⚠️ This partially reverts #1496: it reintroduces `openssl.SetFIPS(true)` 😢 . Removing it will make life more difficult for applications running on non-FIPS distros (like Mariner 2 when not running on FIPS mode) which have properly configured OpenSSL to be FIPS-compliant, as most times one still need to call `FIPS_mode_set(true)`. On the upside, since #1496 we use `openssl.FIPSCapable` instead of `openssl.FIPS`, which makes it less likely to end up calling `openssl.SetFIPS`. Most importantly, we no longer call that function when running on Azure Linux 3, which was my main goal with #1496.


gdams (approved):

dagood:
> This partially reverts #1496: it reintroduces `openssl.SetFIPS(true)` 😢 . Removing it will make life more difficult for applications running on non-FIPS distros (like Mariner 2 when not running on FIPS mode) which have properly configured OpenSSL to be FIPS-compliant, as most times one still need to call `FIPS_mode_set(true)`.

(I assume "Removing it will make life more difficult" should be "Removing it would have made life more difficult" or similar, because this PR adds it back in?)

Why are we supporting this situation? My reading of the ADR is that we will intentionally make this situation harder--there is no provision for this halfway point. In the PR thread, my sense was that it only makes sense in a testing scenario, and we should instead document how the user can properly enable OpenSSL FIPS "locally-ish" on the test machine.

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -1676,27 +1676,32 @@ index 00000000000000..ef5af5d956163e
...
 +package fips140
 +
-+import "syscall"
++import (
++	"internal/godebug"
++	"syscall"
++)
 +
-+// Enabled reports whether FIPS 140 mode is enabled by using GOFIPS=1, GOLANG_FIPS=1,
++var fips140GODEBUG = godebug.New("#fips140")
```
xnox, line 1694: @qmuntal I am reading this code and I am not sure if this is a typo, or if it correct. Everywhere else I see calls to `godebug.New("fips140")` what is the meaning of a leading `#`? I also don't see anywhere for `GODEBUG=#fips140=on` to be set.

dagood: `#` is consumed by `New`: https://pkg.go.dev/internal/godebug#New.

> GODEBUGs meant for use by end users must be listed in ../godebugs/table.go, which is used for generating and checking various documentation. If the name is not listed in that table, New will succeed but calling Value on the returned Setting will panic. To disable that panic for access to an undocumented setting, prefix the name with a #, as in godebug.New("#gofsystrace"). The # is a signal to New but not part of the key used in $GODEBUG.

dagood: Related: upstream changed theirs from `#fips140` to `fips140` post-1.24, in https://github.com/golang/go/commit/10cef816aa9769345016c04032090ae7f5851f5c. (I don't think still having `#` should hurt, though.)

qmuntal: Yep, the leading hash doesn't hurt, but better remove it for consistency and to avoid confusion: https://github.com/microsoft/go/pull/1652.

PR #1505, opened 2025-01-17
Build the Go toolchain using `allowcryptofallback`

qmuntal:
`allowcryptofallback` can be used to allow executing the Go toolchain with `GOFIPS=1`, e.g. `GOFIPS=1 go test .`, even when the Go toolchain itself is not built with `GOEXPERIMENT=systemcrypto`.

Reusing `allowcryptofallback` allows us to drop patch 7, which was tackling the same issue with a more intrusive and less robust approach: unsetting `GOFIPS` when the runtime was initializing and setting it again every time a child process was spawned.

This will help migrating from `GOFIPS` to `GODEBUG=fips140`.

These changes uncovered a latent bug in the Go toolchain: `GOFIPS` was not being honored, for a reason that I still not understand, the Go toolchain didn't pass the `GOFIPS` env var to child processes. Note that this bug doesn't affect setting `GOFIPS` in production environments, only when running commands like `go test`, `go run`, or `go tool dist test`.

Now that `GOFIPS` actually does something when running `go tool dist test` (which we execute in CI to run all the toolchain tests), Mariner 1 and Mariner 2 builders started to fail (see [logs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=921024&view=logs&j=8a09e029-c69f-5388-071d-91fdb54fd26d&t=a90a8cba-8ff3-57f3-801b-18e076d335f4&l=389)) due to FIPS mode not being enabled (remember that we removed the `openssl.SetFIPS(true)` call in #1496, so we no longer enable FIPS mode on demand). To fix this, I've updated the builder scripts so that it enables FIPS mode before running the tests by setting the Mariner/AZL3 specific `OPENSSL_FORCE_FIPS_MODE` env var. Note that we already do something similar on Windows, where FIPS mode is enabled by modifying the system registry.


patches/0004-Use-crypto-backends.patch
```diff
@@ -106,6 +107,30 @@ index f0e3575637c62a..9eab3b4e66e60b 100644
...
+ 	os.Setenv("CC", compilerEnvLookup("CC", defaultcc, goos, goarch))
+ 	// Now that cmd/go is in charge of the build process, enable GOEXPERIMENT.
++	//
++	// Build the Go toolchain with "GOEXPERIMENT=allowcryptofallback". This
++	// allows toolchains not built with "GOEXPERIMENT=systemcrypto" to be used
++	// when GOFIPS=1 is set. For example, when running "GOFIPS=1 go test ./..."
++	// or "GOFIPS=1 go run .".
++	// Shadow goexperiment so that the global variable is not modified.
++	goexperiment := goexperiment
++	if !strings.Contains(goexperiment, "allowcryptofallback") {
```
dagood, line 125: Hmm, wonder if this patch would also let us remove this, to simplify:

https://github.com/microsoft/go/blob/88aa2863217837cd722190b963c57241d63d4c25/eng/_util/buildutil/buildutil.go#L89-L97

dagood (approved):

patches/0004-Use-crypto-backends.patch
qmuntal: Tried but didn't work. We still need to set `allowcryptofallback` for running the tests, else we would need to skip a handful of them,

PR #1507, opened 2025-01-17
CrossPlatformCryptography.md: more consistently associate limits to the table

dagood:
There was some concern that readers wouldn't notice the notes after the table, so for each one, either bring it above the table, or give them numbered notes.


eng/doc/CrossPlatformCryptography.md
```diff
@@ -100,19 +100,19 @@ This section includes the following packages:
 
 <sup>1</sup>When using OpenSSL 3, requires the legacy provider to be enabled.
 
-### AES-GCM keys, nonces, and tags
+<sup>2</sup> AES-GCM supports specific keys, nonces, and tags:
 
-* Key Sizes
+  * Key Sizes
```
karianna: intentional indenting here?

karianna (approved):

eng/doc/CrossPlatformCryptography.md
dagood: It was, but it doesn't seem to actually do anything (and the normal bullet indent looks fine to me on GitHub), so I'll remove.

PR #1508, opened 2025-01-20
Mention GODEBUG=fips140 in FIPS docs

qmuntal:

### DIFF
--- a/eng/doc/fips/README.md
+++ b/eng/doc/fips/README.md
@@ -257,10 +257,10 @@ Prior to 1.24, CommonCrypto/CryptoKit is not used by Microsoft Go.
 
 ### Build option to require FIPS mode
 
-The `requirefips` build tag is available since Go 1.21. See [the "GOFLAGS" example in the build section](#modify-the-build-command).
-The `GOFIPS140=latest` environment variable is available since Go 1.24.
+FIPS mode preference is normally determined at runtime, but the `GOFIPS140=latest` and `requirefips` options can be used to make a program always require FIPS mode and panic if FIPS mode is not enabled:
 
-FIPS mode preference is normally determined at runtime, but both options can be used to make a program always require FIPS mode and panic if FIPS mode is not enabled.
+- The `requirefips` build tag is available since Go 1.21. See [the "GOFLAGS" example in the build section](#modify-the-build-command).
+- The `GOFIPS140=latest` environment variable is available since Go 1.24.
 
 Most programs aren't expected to use these options. Determining FIPS mode at runtime is normal for FIPS compliant applications. This allows the same binary to be deployed to run in both FIPS compliant contexts and non-FIPS contexts, and allows it to be bundled with other binaries that can also run in both contexts. However, it is useful in some cases:
 
--- a/eng/doc/fips/README.md
+++ b/eng/doc/fips/README.md
@@ -66,7 +66,7 @@ There are typically two goals that lead to this document. Creating a FIPS compli
 | `GOEXPERIMENT=systemcrypto` | Default | Compliant | Can be used to create a compliant app. FIPS mode is determined by system-wide configuration. Make sure you are familiar with your platform's system-wide FIPS switch, described in [Usage: Runtime](#usage-runtime). |
 | `GOEXPERIMENT=systemcrypto` | `GODEBUG=fips140=on` or `GOFIPS=1` | Compliant | Can be used to create a compliant app. Depending on platform, the app enables FIPS mode, ensures it is already enabled, or doesn't do any additional checks. The app panics if there is a problem. See [Usage: Runtime](#usage-runtime). |
 | `GOEXPERIMENT=systemcrypto` | `GO_OPENSSL_VERSION_OVERRIDE=1.1.1k-fips` | Compliant | Can be used to create a compliant app. If the app is built for Linux, `systemcrypto` chooses `opensslcrypto`, and the environment variable causes it to load `libcrypto.so.1.1.1k-fips` instead of using the automatic search behavior. This environment variable has no effect with `cngcrypto`. |
-| `GOEXPERIMENT=systemcrypto` and `-tags=requirefips` | Default | Compliant | Can be used to create a compliant app. The behavior is the same as `GODEBUG=fips140=on`, but no runtime configuration is necessary. See [the `requirefips` section](#build-option-to-require-fips-mode) for more information on when this "locked-in" approach may be useful rather than the flexible approach. |
+| `GOEXPERIMENT=systemcrypto` and `-tags=requirefips` | Default | Compliant | Can be used to create a compliant app. The behavior is the same as `GODEBUG=fips140=on` and `GOFIPS=1`, but no runtime configuration is necessary. See [the `requirefips` section](#build-option-to-require-fips-mode) for more information on when this "locked-in" approach may be useful rather than the flexible approach. |
 
 Other notes for common configurations:
 
@@ -203,7 +203,7 @@ Another approach that generally works for any build system is to modify the buil
 
 A program built with `systemcrypto` always uses the system-provided cryptography library for supported crypto APIs. This is the case for `opensslcrypto` (always using OpenSSL), `cngcrypto` (always using CNG) and `darwincrypto` (always using CommonCrypto/CryptoKit). If the platform's crypto library can't be found or loaded, the Go program panics during initialization.
 
-The following sections describe how to enable FIPS mode and the effect of the `GODEBUG=fips140` setting on each supported platform.
+The following sections describe how to enable FIPS mode and the effect of the `GODEBUG=fips140` and `GOFIPS=1` settings on each supported platform.
 
 > [!NOTE]
 > Since Go 1.24, setting `GOFIPS=1` is equivalent to setting `GODEBUG=fips140=on`. The latter is the recommended way to enable FIPS mode. Support for the `GOFIPS` environment variable will be removed in Go 1.25.
@@ -216,6 +216,8 @@ The following sections describe how to enable FIPS mode and the effect of the `G
 To set FIPS mode on Linux, use one of the following options. The first match in this list wins:
 
 - Explicitly enable it by setting the environment variable `GODEBUG=fips140=on`.
+- Explicitly enable it by setting the environment variable `GOFIPS=1`.
+- Explicitly disable it by setting the environment variable `GOFIPS=0`.
 - Implicitly enable it by booting the Linux Kernel in FIPS mode.
   - The Linux Kernel's FIPS mode sets the content of `/proc/sys/crypto/fips_enabled` to `1`. The Go runtime reads this file.
 
@@ -254,7 +256,7 @@ FIPS mode is normally determined at runtime, but the `requirefips` build tag can
 
 Most programs aren't expected to use this tag. Determining FIPS mode at runtime is normal for FIPS compliant applications. This allows the same binary to be deployed to run in both FIPS compliant contexts and non-FIPS contexts, and allows it to be bundled with other binaries that can also run in both contexts. However, in some situations, compile-time `requirefips` is desirable:
 
-- Dependence on environment variables like `GODEBUG` in any way may be undesirable.
+- Dependence on environment variables like `GODEBUG` and `GOFIPS` in any way may be undesirable.
 - The program's documentation can state it will always run in FIPS mode without any nuance about environment variables.
 - If the program is used by someone unfamiliar with the system they're configuring, the panic will help catch mistakes before they become a problem.
 
@@ -295,7 +297,7 @@ This table shows an example of the fragile behavior that results from using `all
 | `GOOS=linux GOEXPERIMENT=systemcrypto,allowcryptofallback` | Compliant | *Not recommended,* but can be used to create a compliant app, as `allowcryptofallback` has no effect in this situation. |
 | `GOOS=linux CGO_ENABLED=0 GOEXPERIMENT=systemcrypto,allowcryptofallback` | Not compliant | Crypto usage is not FIPS compliant. `systemcrypto` on `linux` picks the OpenSSL backend. The backend requires cgo, so `CGO_ENABLED=0` would normally result in a build error. However, `allowcryptofallback` causes the Go standard library crypto to be used and ignores the error. |
 
-A scenario we expect is that a dev attempts to rebuild an open source Go app with an OpenSSL backend to start working towards FIPS compliance. A Dockerfile or other build script provided by the open source project may set `CGO_ENABLED=0` in a non-obvious way. With *silent crypto backend fallback*, the dev needs to notice that the OpenSSL backend isn't being used in some situations (e.g. `GODEBUG=fips140=on` causes failure) and figure out why. If they don't notice, they may deliver an app that uses Go crypto without realizing it. The compatibility check makes it so this issue blocks the build and can't be missed.
+A scenario we expect is that a dev attempts to rebuild an open source Go app with an OpenSSL backend to start working towards FIPS compliance. A Dockerfile or other build script provided by the open source project may set `CGO_ENABLED=0` in a non-obvious way. With *silent crypto backend fallback*, the dev needs to notice that the OpenSSL backend isn't being used in some situations (e.g. `GODEBUG=fips140=on` and `GOFIPS=1` causes failure) and figure out why. If they don't notice, they may deliver an app that uses Go crypto without realizing it. The compatibility check makes it so this issue blocks the build and can't be missed.
 
 > [!NOTE]
 > In rare cases, it may be more practical to use `allowcryptofallback` than to remove the `GOEXPERIMENT`. For example, a generic build script that supports many platforms, some of which don't support crypto backends, may find it practical to use `GOEXPERIMENT=systemcrypto,allowcryptofallback` despite the risk of unclear or accidental fallback to Go crypto.
--- a/eng/doc/fips/README.md
+++ b/eng/doc/fips/README.md
@@ -39,9 +39,11 @@ The Microsoft Go fork provides several ways to configure the crypto backend and 
   - [`GOEXPERIMENT=<backend>crypto` environment variable](#usage-build)
   - [`goexperiment.<backend>crypto` build tag](#usage-build)
   - [`requirefips` build tag](#build-option-to-require-fips-mode)
+  - [`GOFIPS140=latest` environment variable](#build-option-to-require-fips-mode)
   - [`GOEXPERIMENT` `allowcryptofallback`](#build-option-to-use-go-crypto-if-the-backend-compatibility-check-fails)
   - [`import _ "crypto/tls/fipsonly"` source change](#tls-with-fips-compliant-settings)
 - Runtime configuration:
+  - [`GOFIPS` environment variable](#usage-runtime)
   - [`GODEBUG=fips140` setting](#usage-runtime)
   - (OpenSSL backend) [`GO_OPENSSL_VERSION_OVERRIDE` environment variable](#runtime-openssl-version-override)
   - (OpenSSL backend) [`/proc/sys/crypto/fips_enabled` file containing `1`](#linux-fips-mode-openssl)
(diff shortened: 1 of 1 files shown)


eng/doc/fips/README.md
```diff
@@ -209,20 +212,22 @@ The following sections describe how to enable FIPS mode and the effect of the `G
...
+- Explicitly enable it by setting the environment variable `GODEBUG=fips140=on`.
 - Implicitly enable it by booting the Linux Kernel in FIPS mode.
   - The Linux Kernel's FIPS mode sets the content of `/proc/sys/crypto/fips_enabled` to `1`. The Go runtime reads this file.
 
-If the Go runtime detects a FIPS preference, it configures OpenSSL during program initialization. This includes disabling FIPS mode if `GOFIPS=0` even if OpenSSL is configured to be in FIPS mode by default. If configuration fails, program initialization panics.
+If FIPS mode is required, it configures OpenSSL during program initialization. If configuration fails, program initialization panics.
 
 If no preference is detected, the Go runtime doesn't set the OpenSSL FIPS mode, and the standard OpenSSL configuration is left unchanged. For more information about the standard OpenSSL FIPS behavior, see https://www.openssl.org/docs/fips.html.
 
+> [!NOTE]
```
karianna: Should that be a WARN instead?

karianna (approved):

eng/doc/fips/README.md
qmuntal: Good point. Will do.

gdams (approved):

eng/doc/fips/README.md
```diff
@@ -52,18 +52,18 @@ The Microsoft Go fork provides several ways to configure the crypto backend and
 There are typically two goals that lead to this document. Creating a FIPS compliant app is one. The other is to comply with internal Microsoft crypto policies that have been set for Go. This table summarizes common configurations and how suitable each one is for these goals.
 
 > [!NOTE]
-> This section assumes the use of Microsoft Go 1.21 or later.
+> This section assumes the use of Microsoft Go 1.24 or later.
 >
+> 1.24 introduces `GODEBUG=fips140` as a preferred way to enable FIPS mode. See also [the Go 1.24 changelog](#go-124-feb-2025).
 > 1.21 introduces `systemcrypto`, `requirefips`, and a build-time compatibility check for the selected crypto backend. The Usage sections go into more detail about the differences between 1.19/1.20 and 1.21 in context. See also [the Go 1.21 changelog](#go-121-aug-2023).
```
dagood, line 61: Runs together and makes it harder to notice version-specific info when browsing:

```suggestion
>
> 1.21 introduces `systemcrypto`, `requirefips`, and a build-time compatibility check for the selected crypto backend. The Usage sections go into more detail about the differences between 1.19/1.20 and 1.21 in context. See also [the Go 1.21 changelog](#go-121-aug-2023).
```

eng/doc/fips/README.md
```diff
@@ -42,7 +42,7 @@ The Microsoft Go fork provides several ways to configure the crypto backend and
   - [`GOEXPERIMENT` `allowcryptofallback`](#build-option-to-use-go-crypto-if-the-backend-compatibility-check-fails)
   - [`import _ "crypto/tls/fipsonly"` source change](#tls-with-fips-compliant-settings)
 - Runtime configuration:
-  - [`GOFIPS` environment variable](#usage-runtime)
+  - [`GODEBUG=fips140` setting](#usage-runtime)
```
dagood, line 47: I think GOFIPS should still be mentioned here until it's no longer in a supported version of Microsoft Go. (At least, this list was intended to include everything.)

eng/doc/fips/README.md
```diff
@@ -42,7 +42,7 @@ The Microsoft Go fork provides several ways to configure the crypto backend and
   - [`GOEXPERIMENT` `allowcryptofallback`](#build-option-to-use-go-crypto-if-the-backend-compatibility-check-fails)
```
dagood, line 43: Line above this one (too far from a diff to comment directly) mentions `requirefips` and should maybe also mention `GOFIPS140=latest`.

eng/doc/fips/README.md
```diff
@@ -200,7 +200,10 @@ Another approach that generally works for any build system is to modify the buil
 
 A program built with `systemcrypto` always uses the system-provided cryptography library for supported crypto APIs. This is the case for `opensslcrypto` (always using OpenSSL), `cngcrypto` (always using CNG) and `darwincrypto` (always using CommonCrypto/CryptoKit). If the platform's crypto library can't be found or loaded, the Go program panics during initialization.
 
-The following sections describe how to enable FIPS mode and the effect of the `GOFIPS` environment variable on each supported platform.
+The following sections describe how to enable FIPS mode and the effect of the `GODEBUG=fips140` setting on each supported platform.
+
+> [!NOTE]
+> Since Go 1.24, setting `GOFIPS=1` is equivalent to setting `GODEBUG=fips140=on`. The latter is the recommended way to enable FIPS mode. Support for the `GOFIPS` environment variable will be removed in Go 1.25.
```
dagood, line 209: I think this doesn't make it clear what `GOFIPS=1` means in Go 1.23.

dagood (changes requested):
I don't think this holds up to the claim that the doc describes all in-support versions of Go. Some comments about that, but I haven't gotten to the later parts of the doc yet.

eng/doc/fips/README.md
qmuntal: I'm afraid I don't get what's missing 😅

dagood: The PR deletes a lot of info about GOFIPS throughout:

```diff
- The following sections describe how to enable FIPS mode and the effect of the `GOFIPS` environment variable on each supported platform.

To set FIPS mode on Linux, use one of the following options. The first match in this list wins:

- - Explicitly enable it by setting the environment variable `GOFIPS=1`.
- - Explicitly disable it by setting the environment variable `GOFIPS=0`.
+ - Explicitly enable it by setting the environment variable `GODEBUG=fips140=on`.

- This includes disabling FIPS mode if `GOFIPS=0` even if OpenSSL is configured to be in FIPS mode by default. If configuration fails, program initialization panics.

- If the Go runtime detects `GOFIPS=1` and FIPS policy is not enabled [...]
```

And `GOFIPS` is replaced by `GODEBUG=fips140` in many places.

When 1.24 is released, this doc will claim to describe both 1.24 and 1.23. Until 1.24 is released, it needs to describe 1.22 and 1.23. I think the removals and replacements make it so the 1.22 and 1.23 behavior isn't clear IMO.

This note seemed like it might have been trying to bridge the gap, but I don't think it does. (But maybe that wasn't the intent, anyway.)

qmuntal: Thanks for the clarification. What you say makes all the sense. I've reintroduced the `GOFIPS` references.

eng/doc/fips/README.md
```diff
@@ -52,18 +54,19 @@ The Microsoft Go fork provides several ways to configure the crypto backend and
 There are typically two goals that lead to this document. Creating a FIPS compliant app is one. The other is to comply with internal Microsoft crypto policies that have been set for Go. This table summarizes common configurations and how suitable each one is for these goals.
 
 > [!NOTE]
-> This section assumes the use of Microsoft Go 1.21 or later.
+> This section assumes the use of Microsoft Go 1.24 or later.
+>
+> 1.24 introduces `GODEBUG=fips140` as a preferred way to enable FIPS mode. See also [the Go 1.24 changelog](#go-124-feb-2025).
```
dagood: I think including `=on` is good to match "to enable FIPS mode" here. Also, I think including the `=on` pretty frequently is good for readers who aren't already familiar with `GODEBUG` and might think that setting `GODEBUG=fips140` will do something on its own.

```suggestion
> 1.24 introduces `GODEBUG=fips140=on` as a preferred way to enable FIPS mode. See also [the Go 1.24 changelog](#go-124-feb-2025).
```

eng/doc/fips/README.md
```diff
@@ -209,31 +215,30 @@ The following sections describe how to enable FIPS mode and the effect of the `G
...
 
-If the Go runtime detects a FIPS preference, it configures OpenSSL during program initialization. This includes disabling FIPS mode if `GOFIPS=0` even if OpenSSL is configured to be in FIPS mode by default. If configuration fails, program initialization panics.
+If the Go runtime detects `GODEBUG=fips140=on` and OpenSSL is not using a FIPS-compliant engine or provider, the program will panic during program initialization. This may be useful to detect and refuse to run with incorrectly configured OpenSSL installations. Otherwise, `GODEBUG=fips140` has no effect.
 
-If no preference is detected, the Go runtime doesn't set the OpenSSL FIPS mode, and the standard OpenSSL configuration is left unchanged. For more information about the standard OpenSSL FIPS behavior, see https://www.openssl.org/docs/fips.html.
+For more information about the standard OpenSSL FIPS behavior, see https://www.openssl.org/docs/fips.html.
+
+> [!WARNING]
+> Previous to Go 1.24, setting `GOFIPS` would make the Go runtime to try to modify the configured FIPS mode. This is no longer supported.
```
dagood: Previous to -> Prior to, present tense (1.23 can still be ran) (matches other "prior to"s in doc), and I think keeping some more of the explicit description in the note makes sense:

```suggestion
> Prior to Go 1.24, setting `GOFIPS` makes the Go runtime attempt to modify the configured FIPS mode.
> This includes disabling FIPS mode if `GOFIPS=0` even if OpenSSL is configured to be in FIPS mode by default.
>
> Since Go 1.24, the Go runtime no longer makes any attempt to modify OpenSSL FIPS mode.
```

eng/doc/fips/README.md
```diff
@@ -209,31 +215,30 @@ The following sections describe how to enable FIPS mode and the effect of the `G
 
 To set FIPS mode on Linux, use one of the following options. The first match in this list wins:
 
+- Explicitly enable it by setting the environment variable `GODEBUG=fips140=on`.
 - Explicitly enable it by setting the environment variable `GOFIPS=1`.
 - Explicitly disable it by setting the environment variable `GOFIPS=0`.
 - Implicitly enable it by booting the Linux Kernel in FIPS mode.
   - The Linux Kernel's FIPS mode sets the content of `/proc/sys/crypto/fips_enabled` to `1`. The Go runtime reads this file.
```
dagood, line 222: Maybe I'm just not connecting a few dots, but (related to https://github.com/microsoft/go/pull/1513#issuecomment-2613282156), how do these work?

`GODEBUG=fips140=on OPENSSL_FORCE_FIPS_MODE=1 go test .`
Seems described by this list: turned on `fips140`, and configuring OpenSSL makes it so that when the Go runtime checks whether FIPS mode is enabled, it gets a `true`.

`OPENSSL_FORCE_FIPS_MODE=1 go test .`
Based on this doc, I'm not sure what this will do. From the ADR, I'd expect OpenSSL's mode to be detected and for the runtime to internally set `GODEBUG=fips140=on` (essentially).

(Or, more broadly, if I use `OPENSSL_CONF`.)

eng/doc/fips/README.md
```diff
@@ -209,31 +215,30 @@ The following sections describe how to enable FIPS mode and the effect of the `G
...
 
 To enable FIPS mode on Windows, [enable the Windows FIPS policy](https://docs.microsoft.com/en-us/windows/security/threat-protection/fips-140-validation#step-3-enable-the-fips-security-policy).
 
-If the Go runtime detects `GOFIPS=1` and FIPS policy is not enabled, the program will panic during program initialization. This may be useful to detect and refuse to run on incorrectly configured Windows systems. Otherwise, `GOFIPS` has no effect.
-
-> [!NOTE]
-> Unlike `opensslcrypto`, a Windows program built with `cngcrypto` doesn't include the ability to enable/disable FIPS mode. The change must be made by configuring the OS, not the Go program.
->
-> This is because Windows FIPS mode is not a per-process setting, and changing it may require elevated permissions. We expect that adding a feature that attempts to change the Windows policy would have unintended consequences.
+If the Go runtime detects `GODEBUG=fips140=on` and FIPS policy is not enabled, the program will panic during program initialization. This may be useful to detect and refuse to run on incorrectly configured Windows systems. Otherwise, `GODEBUG=fips140` has no effect.
```
dagood: Mention `GOFIPS`? I think the note earlier might technically cover this, but CNG would be the only backend config section not to explicitly mention `GOFIPS` if we don't include it here.

eng/doc/fips/README.md
```diff
@@ -245,9 +250,13 @@ The `requirefips` feature is available since Go 1.21.
 
 FIPS mode is normally determined at runtime, but the `requirefips` build tag can be used to make a program always require FIPS mode and panic if FIPS mode can't be enabled for any reason.
 
+> [!NOTE]
+> Since Go 1.24, one can achieve the same effect by setting `GOFIPS140=latest` when building the program.
+> This is the recommended way to build a program that always requires FIPS mode, as it uses the the same mechanism as upstream Go.
```
dagood: Should `requirefips` be described by the note rather than `GOFIPS140=latest`? This would emphasize GOFIPS140 more, and be more consistent with the other prior-to-1.24 notes.

eng/doc/fips/README.md
```diff
@@ -209,31 +215,30 @@ The following sections describe how to enable FIPS mode and the effect of the `G
...
->
-> This is because Windows FIPS mode is not a per-process setting, and changing it may require elevated permissions. We expect that adding a feature that attempts to change the Windows policy would have unintended consequences.
+If the Go runtime detects `GODEBUG=fips140=on` and FIPS policy is not enabled, the program will panic during program initialization. This may be useful to detect and refuse to run on incorrectly configured Windows systems. Otherwise, `GODEBUG=fips140` has no effect.
 
 For testing purposes, Windows FIPS policy can be enabled via the registry key `HKLM\SYSTEM\CurrentControlSet\Control\Lsa\FipsAlgorithmPolicy`, dword value `Enabled` set to `1`.
 
 ### macOS FIPS mode (CommonCrypto/CryptoKit)
 
-CommonCrypo/CrytoKit is FIPS compliant by default. This means that regardless of which mode you set `GOFIPS` to, the cryptographic functions will always be FIPS-enabled.
+CommonCrypo/CrytoKit is FIPS compliant by default. This means that regardless of which mode you set `GODEBUG=fips140` or `GOFIPS` to, the cryptographic functions will always be FIPS-enabled.
```
```suggestion
CommonCrypto/CryptoKit is FIPS compliant by default. This means that regardless of which mode you set `GODEBUG=fips140` or `GOFIPS` to, the cryptographic functions will always be FIPS-enabled.
```

dagood: It seems like this means this backend will always put the Go runtime into FIPS-enabled mode, considering the ADR. I'm not sure if the implications are interesting enough to be worth mentioning, though.

eng/doc/fips/README.md
```diff
@@ -209,31 +215,30 @@ The following sections describe how to enable FIPS mode and the effect of the `G
...
 To set FIPS mode on Linux, use one of the following options. The first match in this list wins:
 
+- Explicitly enable it by setting the environment variable `GODEBUG=fips140=on`.
 - Explicitly enable it by setting the environment variable `GOFIPS=1`.
 - Explicitly disable it by setting the environment variable `GOFIPS=0`.
 - Implicitly enable it by booting the Linux Kernel in FIPS mode.
   - The Linux Kernel's FIPS mode sets the content of `/proc/sys/crypto/fips_enabled` to `1`. The Go runtime reads this file.
 
-If the Go runtime detects a FIPS preference, it configures OpenSSL during program initialization. This includes disabling FIPS mode if `GOFIPS=0` even if OpenSSL is configured to be in FIPS mode by default. If configuration fails, program initialization panics.
+If the Go runtime detects `GODEBUG=fips140=on` and OpenSSL is not using a FIPS-compliant engine or provider, the program will panic during program initialization. This may be useful to detect and refuse to run with incorrectly configured OpenSSL installations. Otherwise, `GODEBUG=fips140` has no effect.
```
dagood: I think the "preference" language makes it work better while describing both `1` and `0` and `GOFIPS`:

```suggestion
If the Go runtime detects a preference to enable FIPS and OpenSSL is not using a FIPS-compliant engine or provider, the program will panic during program initialization. This may be useful to detect and refuse to run with incorrectly configured OpenSSL installations.

If the Go runtime detects a preference to disable FIPS and OpenSSL is using a FIPS-compliant engine or provider, the program will panic during program initialization.

Otherwise, FIPS preference has no effect.
```

Maybe "enable" and "disable" and "set" aren't even the right words to use anymore in 1.24, because it seems like it's all about enforcement rather than really changing anything.

eng/doc/fips/README.md
```diff
@@ -209,31 +215,30 @@ The following sections describe how to enable FIPS mode and the effect of the `G
 
 To set FIPS mode on Linux, use one of the following options. The first match in this list wins:
```
dagood: This isn't really setting the mode anymore, and it might make sense to match the previous suggestion to keep calling it a preference:

```suggestion
To set FIPS mode preference on Linux, use one of the following options. The first match in this list wins:
```

eng/doc/fips/README.md
qmuntal: Done!

eng/doc/fips/README.md
qmuntal: I've reworded all this section to make it less `requirefips`-dependent while adding some more info.

eng/doc/fips/README.md
qmuntal: Hmm, when writing the ADR I was thinking on a Mariner / Azure Linux  booted in FIPS mode or Windows with the FIPS policy enabled, and that's what I end up implementing. But you are right, having the FIPS mode enabled at the OpenSSL config level should also set `GODEBUG=fips140=on`. Will fix that, thanks!

qmuntal: https://github.com/microsoft/go/pull/1519

eng/doc/fips/README.md
qmuntal: macOS doesn't have the concept of FIPS mode. It being FIPS compliant is an internal detail that is not surfaced in any API. This means that we don't know if the user intent is to really be in FIPS mode or not. We shouldn't set the Go runtime into FIPS-enabled mode (which means `fips140.Enabled` return true and `crypto/tls` only accepts FIPS compliant settings) unless we have a clear signal, and that signal is `GODEBUG=fips140` or `GOFIPS` on macOS, nothing else.

dagood: I see, that makes sense, but certainly isn't what I'd expect based on the ADR, and I think this text is misleading... `CommonCrypo/CrytoKit is FIPS compliant by default`, and `the cryptographic functions will always be FIPS-enabled`, but `crypto/tls` will accept non-FIPS-compliant settings by default? I don't think the doc establishes the different meanings of FIPS modes/compliance well enough to try to make detailed distinctions like this without more direct explanation.

dagood: Details like this make me wonder again if it really would be worth splitting this doc up into <1.24 and >=1.24. 1.24 has enough nuance that I worry that trying to describe both is simply not useful for the reader. (It could be as simple as duplicating the doc as the starting point and still maintain it in `microsoft/main`. Keeping it in main would make sure it's still easy to make clarifications that apply to both pre- and post-1.24 without dealing with backports and multiple PRs.)

dagood: FWIW, the impression I get from this (and maybe the way to document it?) is that the lack of user intent is a problem that causes a potential compatibility issue, and because we prefer compatibility, we choose not to enable Go runtime FIPS mode in this case.

qmuntal: > I see, that makes sense, but certainly isn't what I'd expect based on the ADR

Let's discuss this in the next team meeting. It might deserve a new ADR or amending the one we already have.

> I think this text is misleading...

Agree, I've rewritten the macOS section to make it less confusing.

>  I don't think the doc establishes the different meanings of FIPS modes/compliance well enough to try to make detailed distinctions like this without more direct explanation.

Good point. I'll create a follow up PR clearly stating what Go FIPS mode means, and how it might differ from system FIPS mode.

eng/doc/fips/README.md
```diff
@@ -213,56 +213,62 @@ The following sections describe how to enable FIPS mode and the effect of the `G
...
 
-If the Go runtime detects `GODEBUG=fips140=on` and FIPS policy is not enabled, the program will panic during program initialization. This may be useful to detect and refuse to run on incorrectly configured Windows systems. Otherwise, `GODEBUG=fips140` has no effect.
+If the Go runtime detects `GOFIPS=1` or `GODEBUG=fips140=on` and FIPS policy is not enabled, the program will panic during program initialization. This may be useful to detect and refuse to run on incorrectly configured Windows systems. Otherwise, `GODEBUG=fips140` has no effect.
 
 For testing purposes, Windows FIPS policy can be enabled via the registry key `HKLM\SYSTEM\CurrentControlSet\Control\Lsa\FipsAlgorithmPolicy`, dword value `Enabled` set to `1`.
 
 ### macOS FIPS mode (CommonCrypto/CryptoKit)
 
-CommonCrypo/CrytoKit is FIPS compliant by default. This means that regardless of which mode you set `GODEBUG=fips140` or `GOFIPS` to, the cryptographic functions will always be FIPS-enabled.
+macOS cryptographic primitives are FIPS compliant by default, there is no need for system-wide nor process-wide configuration.
```
```suggestion
macOS cryptographic primitives are FIPS compliant by default, so there is no need for system-wide nor process-wide configuration.
```

eng/doc/fips/README.md
```diff
@@ -200,60 +203,72 @@ Another approach that generally works for any build system is to modify the buil
...
 ## Usage: Extra configuration options
 
 ### Build option to require FIPS mode
 
-The `requirefips` feature is available since Go 1.21.
+The `requirefips` build tag is available since Go 1.21. See [the "GOFLAGS" example in the build section](#modify-the-build-command).
+The `GOFIPS140=latest` environment variable is available since Go 1.24.
 
-FIPS mode is normally determined at runtime, but the `requirefips` build tag can be used to make a program always require FIPS mode and panic if FIPS mode can't be enabled for any reason.
+FIPS mode preference is normally determined at runtime, but both options can be used to make a program always require FIPS mode and panic if FIPS mode is not enabled.
```
dagood: The ordering and bunching together of 1.21 and 1.24 sentences seems like it can be improved:

```
FIPS mode preference is normally determined at runtime, but the `GOFIPS140=latest` and `requirefips` options can be used to make a program always require FIPS mode and panic if FIPS mode is not enabled:

- The `requirefips` build tag is available since Go 1.21. See [the "GOFLAGS" example in the build section](#modify-the-build-command).
- The `GOFIPS140=latest` environment variable is available since Go 1.24.
```

(GitHub doesn't let me put this up as a suggestion because it spans deleted lines. 😄)

---

FIPS mode preference is normally determined at runtime, but the `GOFIPS140=latest` and `requirefips` options can be used to make a program always require FIPS mode and panic if FIPS mode is not enabled:

- The `requirefips` build tag is available since Go 1.21. See [the "GOFLAGS" example in the build section](#modify-the-build-command).
- The `GOFIPS140=latest` environment variable is available since Go 1.24.

eng/doc/fips/README.md
```diff
@@ -200,60 +203,72 @@ Another approach that generally works for any build system is to modify the buil
...
 
-Most programs aren't expected to use this tag. Determining FIPS mode at runtime is normal for FIPS compliant applications. This allows the same binary to be deployed to run in both FIPS compliant contexts and non-FIPS contexts, and allows it to be bundled with other binaries that can also run in both contexts. However, in some situations, compile-time `requirefips` is desirable:
+Most programs aren't expected to use this tag. Determining FIPS mode at runtime is normal for FIPS compliant applications. This allows the same binary to be deployed to run in both FIPS compliant contexts and non-FIPS contexts, and allows it to be bundled with other binaries that can also run in both contexts. However, it is useful in some cases:
```
```suggestion
Most programs aren't expected to use these options. Determining FIPS mode at runtime is normal for FIPS compliant applications. This allows the same binary to be deployed to run in both FIPS compliant contexts and non-FIPS contexts, and allows it to be bundled with other binaries that can also run in both contexts. However, it is useful in some cases:
```

dagood (approved):

dagood:
I filed a couple followup issues, also gathered my thoughts on the topics: 🙂 

* https://github.com/microsoft/go/issues/1534
* https://github.com/microsoft/go/issues/1535

PR #1513, opened 2025-01-24
Remove `openssl.SetFIPS(true)` call

qmuntal:
As agreed in https://github.com/microsoft/go-lab/blob/main/docs/adr/0012-remove-gofips.md, we shouldn't try to modify the OpenSSL FIPS mode.

This PR removes the `openssl.SetFIPS(true)` call and update our build scripts to enable FIPS mode system-wide.

Our CI Mariner 2 image is not FIPS-enabled by default, so we need to force FIPS mode by setting `OPENSSL_FORCE_FIPS_MODE`. That flag should be passes to the `TestScript` child processes as they only inherit a filtered set of environment variables, which includes `GODEBUG`.

Note that since we switched from `GOFIPS` to `GODEBUG=fips140`, our test FIPS test coverage has increased, as `GOFIPS` was not being passed to `TestScript` child processes, making them not aware of the required FIPS mode.

Also, this is unlikely that users need to update their code to also pass `OPENSSL_FORCE_FIPS_MODE` to child processes that don't inherit all environment variables. Mainly because they should be running a FIPS-enabled Mariner image on production. If they don't, possible for testing purposes, then child processes won't inherit the `GODEBUG` env var neither.

For #1445.

### DIFF
--- a/eng/_util/cmd/run-builder/systemfips_linux.go
+++ b/eng/_util/cmd/run-builder/systemfips_linux.go
@@ -1,0 +1,73 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//go:build linux
+
+package main
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+)
+
+// enableSystemWideFIPS fallback is a no-op because the current platform either doesn't support or
+// doesn't require system-wide FIPS to be enabled to run tests.
+func enableSystemWideFIPS() (restore func(), err error) {
+	cmd := exec.Command("openssl", "version", "-a")
+	log.Printf("---- Running command: %v\n", cmd.Args)
+	out, err := cmd.CombinedOutput()
+	sout := string(out)
+	if err != nil {
+		return nil, fmt.Errorf("failed to check openssl version: %v, %v", err, string(out))
+	}
+	log.Print(sout)
+
+	lines := strings.Split(sout, "\n")
+	if !strings.Contains(sout, "OpenSSL 1.") {
+		// Only OpenSSL 1 needs special handling for FIPS mode,
+		// at least on the platforms we test on.
+		log.Println("Using fallback (no-op) for enableSystemWideFIPS. It either isn't supported on this platform or isn't necessary.")
+		//return nil, nil
+	}
+
+	// Search for the OPENSSLDIR path in the output.
+	var ossldir string
+	for _, line := range lines {
+		var found bool
+		if ossldir, found = strings.CutPrefix(line, "OPENSSLDIR: "); found {
+			break
+		}
+	}
+	if ossldir == "" {
+		return nil, fmt.Errorf("failed to find OPENSSLDIR in openssl version output")
+	}
+	ossldir = strings.Trim(ossldir, `"`)
+
+	// Append the FIPS configuration to the openssl.cnf file.
+	// OpenSSL will merge duplicated sections, so we don't need
+	// to check if the section already exists.
+	opensslcnf := filepath.Join(ossldir, "openssl.cnf")
+	prevContent, err := os.ReadFile(opensslcnf)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read openssl.cnf file: %v", err)
+	}
+	err = os.WriteFile(opensslcnf, append(prevContent, []byte("\n\n[evp_sect]\nfips_mode = yes\n")...), 0644)
+	if err != nil {
+		return nil, fmt.Errorf("failed to write to openssl.cnf file: %v", err)
+	}
+
+	log.Println("Enabled FIPS mode.")
+
+	return func() {
+		err := os.WriteFile(opensslcnf, prevContent, 0644)
+		if err != nil {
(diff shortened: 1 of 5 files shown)


qmuntal:
Looks like Mariner 2 hasn't ported forward-ported the code to enable FIPS mode from the config file. OpenSSL only officially supports FIPS mode in v1.0.2, so the relevant code was removed in OpenSSL 1.1. Will have to find another way.

patches/0004-Use-crypto-backends.patch
```diff
@@ -198,6 +199,18 @@ index ed0fbf3d53d75b..8111b143a1295b 100644
...
  	tg.parallel()
+diff --git a/src/cmd/go/script_test.go b/src/cmd/go/script_test.go
+index 390a36723787f4..105bac4e1c627b 100644
+--- a/src/cmd/go/script_test.go
++++ b/src/cmd/go/script_test.go
+@@ -254,6 +254,7 @@ func scriptEnv(srv *vcstest.Server, srvCertFile string) ([]string, error) {
+ 		"HGRCPATH=",
+ 		"GOTOOLCHAIN=auto",
+ 		"newline=\n",
++		"OPENSSL_FORCE_FIPS_MODE=" + os.Getenv("OPENSSL_FORCE_FIPS_MODE"), // useful for testing on Mariner 2.
```
dagood: I feel like this would fit better in `extraEnvKeys`. (Unless there's a reason that it must be set to empty string even if the env var is unassigned?)

eng/_util/cmd/run-builder/systemfips_linux.go
```diff
@@ -0,0 +1,31 @@
...
+// Licensed under the MIT License.
+
+package main
+
+import (
+	"log"
+	"os"
+)
+
+// enableSystemWideFIPS enables Mariner and Azure Linux 3 process-wide FIPS mode.
```
dagood: I think in context what's here is plenty, but I figure a bit more explicit detail doesn't hurt:

```suggestion
// enableSystemWideFIPS enables Mariner and Azure Linux 3 process-wide FIPS mode
// for any process that inherits the current process' environment variables.
```

eng/_util/cmd/run-builder/systemfips_linux.go
```diff
@@ -0,0 +1,31 @@
...
+func enableSystemWideFIPS() (restore func(), err error) {
+	// FIPS mode is enabled if OPENSSL_FORCE_FIPS_MODE is set, regardless of the value.
+	_, ok := os.LookupEnv("OPENSSL_FORCE_FIPS_MODE")
+	if ok {
+		log.Println("FIPS mode already enabled.")
```
dagood: Might be confusing if you only see the log because there are a few things "FIPS mode" could mean.

```suggestion
		log.Println("Mariner and Azure Linux 3 forced FIPS mode (OPENSSL_FORCE_FIPS_MODE) already enabled.")
```

eng/_util/cmd/run-builder/systemfips_linux.go
```diff
@@ -0,0 +1,31 @@
...
+		return nil, nil
+	}
+
+	env("OPENSSL_FORCE_FIPS_MODE", "1")
+	log.Println("Enabled Mariner and Azure Linux 3 FIPS mode.")
```
dagood: `env` will have just logged a message, but it doesn't hurt to make this a bit more explicit too.

```suggestion
	log.Println("Enabled Mariner and Azure Linux 3 FIPS mode (OPENSSL_FORCE_FIPS_MODE).")
```

dagood:
> Also, this is unlikely that users need to update their code to also pass `OPENSSL_FORCE_FIPS_MODE` to child processes that don't inherit all environment variables. Mainly because they should be running a FIPS-enabled Mariner image on production. If they don't, possible for testing purposes, then child processes won't inherit the `GODEBUG` env var neither.

To make sure I have this right: for a user to run a test with forced FIPS mode, this will work:

```
GODEBUG=fips140=on OPENSSL_FORCE_FIPS_MODE=1 go test .
```

For the user to have an issue with passthrough, they'd have to be using yet another custom test runner that passes through `GODEBUG` but not `OPENSSL_FORCE_FIPS_MODE`.

patches/0004-Use-crypto-backends.patch
qmuntal: Good catch, will do!

gdams (approved):

qmuntal:
> > Also, this is unlikely that users need to update their code to also pass `OPENSSL_FORCE_FIPS_MODE` to child processes that don't inherit all environment variables. Mainly because they should be running a FIPS-enabled Mariner image on production. If they don't, possible for testing purposes, then child processes won't inherit the `GODEBUG` env var neither.
> 
> To make sure I have this right: for a user to run a test with forced FIPS mode, this will work:
> 
> ```
> GODEBUG=fips140=on OPENSSL_FORCE_FIPS_MODE=1 go test .
> ```
> 
> For the user to have an issue with passthrough, they'd have to be using yet another custom test runner that passes through `GODEBUG` but not `OPENSSL_FORCE_FIPS_MODE`.

Yep, that's accurate.

PR #1519, opened 2025-01-31
Take into account the OpenSSL FIPS mode when setting GODEBUG=fips140

qmuntal:
As agreed in https://github.com/microsoft/go-lab/blob/main/docs/adr/0012-remove-gofips.md, we should set `GODEBUG=fips140=on` if OpenSSL is in FIPS mode. We are currently only taking into account the kernel FIPS mode (aka `/proc/sys/crypto/fips_enabled`).

This PR adds the missing check. I had to factor the OpenSSL init logic into its own package, `crypto/internal/backend/internal/opensslsetup`, so that it can be used in `crypto/internal/backend/fips140` without creating a dependency cycle.

I've also added some much needed tests to the OpenSSL library selection logic and improved the `TestDependencies` script to reduce the chance of upstream conflicts.


karianna (approved):

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -1794,21 +1800,37 @@ index 00000000000000..83691d7dd42d51
...
 --- /dev/null
 +++ b/src/crypto/internal/backend/fips140/openssl.go
-@@ -0,0 +1,41 @@
+@@ -0,0 +1,57 @@
 +// Copyright 2024 The Go Authors. All rights reserved.
 +// Use of this source code is governed by a BSD-style
 +// license that can be found in the LICENSE file.
 +
-+//go:build goexperiment.opensslcrypto
++//go:build goexperiment.opensslcrypto && cgo
```
dagood, line 1813: Hmm, I'm guessing that the `&& cgo` is needed to make the stdlib build with `allowcryptofallback`? (Otherwise, this scenario doesn't matter because the user will hit the `backenderr` first.)

It feels to me like splitting this file into `openssl_cgo.go` and `openssl_nocgo.go` rather than reusing `nosystemcrypto.go` would make the situation a bit simpler (or at least more straightforward), and it makes a clear place to put a doc comment.

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -1839,6 +1861,195 @@ index 00000000000000..2d126bcfc053de
...
+--- /dev/null
++++ b/src/crypto/internal/backend/internal/opensslsetup/nocgo.go
+@@ -0,0 +1,9 @@
++// Copyright 2025 The Go Authors. All rights reserved.
++// Use of this source code is governed by a BSD-style
++// license that can be found in the LICENSE file.
++
++//go:build !cgo
++
++// Do nothing if cgo is not enabled.
```
dagood: What is this file for? If it's just a placeholder to make sure the package can be pointed to, maybe it could have no build tag at all? Otherwise I kinda wonder why this is `!cgo` rather than `!(goexperiment.opensslcrypto && cgo)`.

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -1839,6 +1861,195 @@ index 00000000000000..2d126bcfc053de
...
++
++	"github.com/golang-fips/openssl/v2"
++)
++
++// knownVersions is a list of supported and well-known libcrypto.so suffixes in decreasing version order.
++// FreeBSD library version numbering does not directly align to the version of OpenSSL.
++// Its preferred search order is 11 -> 111.
++// Some distributions use 1.0.0 and others (such as Debian) 1.0.2 to refer to the same OpenSSL 1.0.2 version.
++// Fedora derived distros use different naming for the version 1.0.x.
++var knownVersions = [...]string{"3", "1.1", "11", "111", "1.0.2", "1.0.0", "10"}
```
dagood, line 1915: Still exists as a duplicate in openssl_linux.go, looks like it just needs cleanup. (Also `lcryptoPrefix`.)

dagood (approved):

patches/0003-Implement-crypto-internal-backend.patch
qmuntal: >  If it's just a placeholder to make sure the package can be pointed to, maybe it could have no build tag at all?

Will do that

> Otherwise I kinda wonder why this is !cgo rather than !(goexperiment.opensslcrypto && cgo).

This package is only imported if `goexperiment.opensslcrypto` is set, so the only variable is `cgo`.

patches/0003-Implement-crypto-internal-backend.patch
qmuntal: > Hmm, I'm guessing that the && cgo is needed to make the stdlib build with allowcryptofallback?

Yep.

> It feels to me like splitting this file into openssl_cgo.go and openssl_nocgo.go

Can do that. Thanks!

Issue #1535, opened 2025-02-06
FIPS docs: describe how to enable OpenSSL 1 and 3 FIPS mode for testing without GOFIPS=1


dagood:
In production, using `GOFIPS=1` to *enforce* FIPS mode may be useful. This is preserved in 1.24+, and 1.24 also makes `GODEBUG=fips140=on` do this enforcement.

In production, using `GOFIPS=1` to *change* FIPS mode should never be useful. If the system isn't in FIPS mode, putting a Go app and that process' OpenSSL engine into FIPS mode doesn't help with compliance, because the rest of the system already isn't compliant. (To be precise: we haven't identified any production scenario where this is useful, and we doubt one exists.)

In testing on Linux, `GOFIPS=1` to *change* FIPS mode can be useful. A dev probably doesn't want to run their entire machine in FIPS mode all the time for compatibility and potential performance reasons, so setting `GOFIPS=1` can be used to test how FIPS behavior affects their app.

(That is Linux-specific: on Windows, you can't change FIPS mode per-process, and it requires admin. macOS doesn't have a systemwide FIPS mode.)

Since 1.24+ removes the ability for `GOFIPS=1` to *change* FIPS mode, we should describe (or refer people to something that shows) how to enable FIPS mode per-process on Linux for testing. This might be system-specific or version-specific. (AzL `OPENSSL_FORCE_FIPS_MODE`, OpenSSL 3 `OPENSSL_CONF`.)

dagood:
* Addressed by #1556

Issue #1534, opened 2025-02-06
FIPS docs: clarify system FIPS mode vs. Go Runtime FIPS mode


dagood:
(Followup item from initial 1.24 FIPS docs PRs.)

There's more nuance in 1.24 about FIPS modes. This can affect things like TLS fipsonly.

On the existing platforms (Linux, Windows), behavior matches what I'd expect based on earlier releases. However, the preview macOS support makes distinctions between platform FIPS, crypto library FIPS, and Go runtime FIPS more important to the user, because macOS doesn't seem to have a widely accepted "system FIPS mode" that we can refer to at runtime.

dagood:
* Addressed by #1556

PR #1556, opened 2025-02-12
Additional FIPS readme updates for 1.24

dagood:
First, a separate commit to use "FIPS 140" rather than "FIPS 140-2" in most places. I don't think we have much to say about FIPS 140-2 in particular vs. FIPS 140-3. (I kept it separate to easily back it out if it doesn't seem right.)

Then, I read through to make updates. While here, I removed most clarifications for totally out of support versions and outdated info to make room for new clarifications. I also specifically looked at addressing:

* https://github.com/microsoft/go/issues/1534
* https://github.com/microsoft/go/issues/1535

Preview: https://github.com/microsoft/go/blob/dev/dagood/fips-docs-1.24/eng/doc/fips/README.md

### DIFF
--- a/eng/doc/fips/README.md
+++ b/eng/doc/fips/README.md
@@ -7,44 +7,54 @@ This directory contains documentation about FIPS and the FIPS implementation in 
 
 ## Background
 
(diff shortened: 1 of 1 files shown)


karianna (approved):

gdams (approved):

qmuntal (approved):

dagood:
Thanks all. 😄

PR #1561, opened 2025-02-13
Don't install custom pwsh on Linux

qmuntal:
Our Linux 1ES images already contain `pwsh`. There is no need to manually install it.


dagood (approved):

dagood:
I initially wanted to try running a pipeline to test, but the pool/vm images on these pipelines are exactly the same as https://github.com/microsoft/go/pull/1561 (which I tried earlier and it succeeded), so this is fine. 👍

PR #1562, opened 2025-02-13
Upgrade Go bootstrap version to go1.24.0-1

qmuntal:
Bump bootstrap version to go1.24.0-1. This allows using MS Go when bootstrapping on macOS.

### DIFF
--- a/eng/utilities.ps1
+++ b/eng/utilities.ps1
@@ -24,7 +24,7 @@ function Download-Stage0() {
   # pre-installed (or the right version pre-installed). This CI script installs a consistent version
   # of Go to handle this. This also makes it easier to locally repro issues in CI that involve a
   # specific version of Go. The downloaded copy of Go is called the "stage 0" version.
-  $stage0_go_version = 'go1.23.3-1'
+  $stage0_go_version = 'go1.24.0-1'
 
   # Source the install script so that we can use the PATH it assigns.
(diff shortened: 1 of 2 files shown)


eng/utilities.ps1
```diff
@@ -24,11 +24,11 @@ function Download-Stage0() {
   # pre-installed (or the right version pre-installed). This CI script installs a consistent version
   # of Go to handle this. This also makes it easier to locally repro issues in CI that involve a
   # specific version of Go. The downloaded copy of Go is called the "stage 0" version.
-  $stage0_go_version = 'go1.23.3-1'
+  $stage0_go_version = 'go1.24.0-1'
 
   # Source the install script so that we can use the PATH it assigns.
   $installScriptPath = Join-Path $PSScriptRoot "_util" "go-install.ps1"
-  . $installScriptPath -Version $stage0_go_version
+  . $installScriptPath -Version $stage0_go_version -AzurePipelinePath
```
dagood: A better fix is to remove:

https://github.com/microsoft/go/blob/b360b6d9e2385d10b8a000bdc4b71757049001ba/eng/pipeline/stages/run-stage.yml#L97-L100

(Either way, should remove these lines.)

(`-AzurePipelinePath` makes things uglier when running on a local machine.)

qmuntal: I like your suggestion, will do that.

dagood (approved):

PR #1612, opened 2025-04-02
Disable GOTOOLCHAIN

### LINKED ISSUE #1213: Disable GOTOOLCHAIN
We currently set the default `GOTOOLCHAIN` value to `local` ([here](https://github.com/microsoft/go/blob/ee7879695fe568396021b1a0ac7a0b9d75326505/patches/0008-Update-default-go.env.patch#L24)) in order to avoid unintentionally switch to the upstream toolchain.

This has been proven to not be enough, see #1208. We should disable the automatic toolchain upgrade functionality until we support upgrading to our own toolchains.

qmuntal:
We were previously setting the `GOTOOLCHAIN` value to `local`. This PR goes further and panics if the user manually changes that variable. 

Fixed https://github.com/microsoft/go/issues/1213.


gdams (dismissed):

patches/0005-Disable-GOTOOLCHAIN-support.patch
```diff
@@ -0,0 +1,63 @@
...
+--- a/go.env
++++ b/go.env
+@@ -7,6 +7,8 @@
+ GOPROXY=https://proxy.golang.org,direct
+ GOSUMDB=sum.golang.org
+ 
+-# Automatically download newer toolchains as directed by go.mod files.
++# Use the locally installed Go toolchain, never downloading a different one.
++# Upstream uses `GOTOOLCHAIN=auto` instead, but `auto` can download and switch
++# to a non-Microsoft Go toolchain, and we want to avoid that.
```
dagood: Good catch, can still untangle it a bit more IMO:

```suggestion
+# to a Go toolchain not built by Microsoft, and we want to avoid that.
```

patches/0005-Disable-GOTOOLCHAIN-support.patch
```diff
@@ -0,0 +1,63 @@
...
+ 	val := os.Getenv(key)
+ 	if val != "" {
++		if key == "GOTOOLCHAIN" && val != "local" {
++			// Don't allow GOTOOLCHAIN to be set to anything but "local".
++			// That would cause the go command to potentially use a different
++			// toolchain than the the Microsoft build of Go.
++			// Allow it while tunning the TestScript test from the cmd/go
++			// package, else many tests will fail.
++			if v := os.Getenv("TESTGO_MS_TESTSCRIPT"); v != "1" {
++				panic(`invalid GOTOOLCHAIN "` + val + `", only "local" is allowed`)
```
dagood: Does a `panic` here end up showing the user a call stack? It might be nicer to print a message and exit with an error code instead, if so, to make it as readable as possible.

OTOH, a trace lets a savvy user easily find the bypass var and the above comment containing our rationale. 😄

patches/0005-Disable-GOTOOLCHAIN-support.patch
```diff
@@ -0,0 +1,63 @@
...
+ 	}
+ 	val := os.Getenv(key)
+ 	if val != "" {
++		if key == "GOTOOLCHAIN" && val != "local" {
++			// Don't allow GOTOOLCHAIN to be set to anything but "local".
++			// That would cause the go command to potentially use a different
++			// toolchain than the the Microsoft build of Go.
++			// Allow it while tunning the TestScript test from the cmd/go
++			// package, else many tests will fail.
++			if v := os.Getenv("TESTGO_MS_TESTSCRIPT"); v != "1" {
```
dagood: Re: discussion in the sync about making this look reasonable to use as a general bypass, my first thought is:
`MS_GOTOOLCHAIN_ALLOW_NON_LOCAL`.

(I thought about terms like "dynamic selection" and "safety", but simple seems like it might be best here.)

patches/0005-Disable-GOTOOLCHAIN-support.patch
```diff
@@ -0,0 +1,63 @@
...
+@@ -401,6 +401,16 @@ func Getenv(key string) string {
+ 	}
+ 	val := os.Getenv(key)
+ 	if val != "" {
++		if key == "GOTOOLCHAIN" && val != "local" {
++			// Don't allow GOTOOLCHAIN to be set to anything but "local".
++			// That would cause the go command to potentially use a different
++			// toolchain than the the Microsoft build of Go.
++			// Allow it while tunning the TestScript test from the cmd/go
++			// package, else many tests will fail.
```
dagood: Typo, but also in light of making it a general bypass:

```suggestion
+			//
+			// We allow bypassing this safety feature. We need to while running
+			// the TestScript test from the cmd/go package, else many tests will fail.
+			// It's also possible for existing workflows to intentionally depend on this behavior.
```

patches/0005-Disable-GOTOOLCHAIN-support.patch
```diff
@@ -0,0 +1,63 @@
...
+--- a/src/cmd/go/internal/cfg/cfg.go
++++ b/src/cmd/go/internal/cfg/cfg.go
+@@ -401,6 +401,16 @@ func Getenv(key string) string {
+ 	}
+ 	val := os.Getenv(key)
+ 	if val != "" {
++		if key == "GOTOOLCHAIN" && val != "local" {
++			// Don't allow GOTOOLCHAIN to be set to anything but "local".
++			// That would cause the go command to potentially use a different
++			// toolchain than the the Microsoft build of Go.
```
dagood: I think the unexpectedness of it is a major motivation and deserves mention.

```suggestion
+			// That could cause the go command to use a different toolchain
+			// than the Microsoft build of Go without warning. This can be
+			// difficult to diagnose and may silently cause the user to
+			// unintentionally build a program that violates Microsoft's
+			// internal policies for Go.
```

patches/0005-Disable-GOTOOLCHAIN-support.patch
```diff
@@ -0,0 +1,63 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: qmuntal <qmuntaldiaz@microsoft.com>
+Date: Thu, 8 Jun 2023 14:17:13 +0200
+Subject: [PATCH] Disable GOTOOLCHAIN support
+
+The GOTOOLCHAIN feature can potentially make a Go invocation switch
+to a non-Microsoft toolchain. To aovid it, change the GOTOOLCHAIN
```
dagood:
```suggestion
to a non-Microsoft toolchain. To avoid it, change the GOTOOLCHAIN
```

patches/0005-Disable-GOTOOLCHAIN-support.patch
qmuntal: Good point. I've added some more context and used `os.Exit`.

dagood:
Unmarked resolved on a few comments that don't seem to have any assoc changes.

qmuntal:
> Unmarked resolved on a few comments that don't seem to have any assoc changes.

Hmm, I forgot to push the changes... Done now 😄

dagood (approved):

PR #1631, opened 2025-04-14
remove @microsoft-golang-review-bot from CODEOWNERS

gdams:
We've switched to GitHub apps so this is no longer needed.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,7 +2,8 @@
 
 * @microsoft/golang-compiler
 
-# Don't assign a code owner to automatically updated files to allow GitHub apps through the auto-merge flow.
+# Don't assign a code owner to automatically updated files to allow GitHub apps to use the
+# auto-merge flow without human intervention.
 go/
 MICROSOFT_REVISION
 VERSION
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,7 +1,7 @@
 # Require review from golang-compiler team for changes in any file. This keeps us in the loop on
 # auto-merge PRs. The review bot is also an owner so that it can still trigger auto-merge for sync
 # PRs on its own. We may remove this rule once auto-merges are routine.
-* @microsoft/golang-compiler @microsoft-golang-review-bot
+* @microsoft/golang-compiler @review-bot-for-go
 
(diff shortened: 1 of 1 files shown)


dagood (changes requested):
We haven't switched quite yet, we still need to make the review count as code owner. https://github.com/microsoft/go-infra/pull/251#issuecomment-2802137664

.github/CODEOWNERS
```diff
@@ -1,7 +1,7 @@
 # Require review from golang-compiler team for changes in any file. This keeps us in the loop on
 # auto-merge PRs. The review bot is also an owner so that it can still trigger auto-merge for sync
 # PRs on its own. We may remove this rule once auto-merges are routine.
-* @microsoft/golang-compiler @microsoft-golang-review-bot
+* @microsoft/golang-compiler @review-bot-for-go
```
dagood: Mentioned in IM, but here for context: this doesn't seem like it'll work: https://github.com/orgs/community/discussions/23064. There is a workaround: explicitly list all the files that the app will touch and assign no code owner.

dagood: Ah, actually the file viewer directly says it won't work:

image

.github/CODEOWNERS
```diff
@@ -1,8 +1,8 @@
...
-# PRs on its own. We may remove this rule once auto-merges are routine.
-* @microsoft/golang-compiler @microsoft-golang-review-bot
+# Require review from golang-compiler team for changes in all files.
 
-# Automatically request review from golang-compiler team for changes in the Microsoft-specific
-# files. This takes precedence over earlier rules in the file.
-/eng/ @microsoft/golang-compiler
+* @microsoft/golang-compiler
+
+# Don't assign a code owner to automatically updated files to allow GitHub apps through the auto-merge flow.
```
dagood: A little more specific:

```suggestion
# Don't assign a code owner to automatically updated files to allow GitHub apps to use the
# auto-merge flow without human intervention.
```

dagood (approved):

dagood (approved):
LGTM, *but* with the way it's written it must be merged at the same time as https://github.com/microsoft/go-infra/pull/251, which [wasn't] ready [as of writing].

PR #1634, opened 2025-04-15
Remove deprecated registry and Mariner1

### LINKED ISSUE #1371: Replace `golangpublicimages.azurecr.io` images
Our CI pipelines are currently puling some images from `golangpublicimages.azurecr.io/go-infra-images/prereqs`, which is deprecated and shouldn't be used. 

We should replace it with `mcr.microsoft.com/microsoft-go/infra-images`, whose images are generated from https://github.com/microsoft/go-infra-images.

mertakman:
This PR removes long time unsupported linux distribution and removes deprecated docker registry.
`mcr.microsoft.com/microsoft-go/infra-images` does not support Mariner 1, and most other services stopped supporting it long ago and it was single remaining dependency of deprecated docker registry. In this PR we removing support for it on Microsoft version of Go, so we can completely drop `golangpublicimages.azurecr.io/go-infra-images/prereqs` dependency of this repository.
Fixes #1371


eng/pipeline/stages/run-stage.yml
```diff
@@ -75,13 +75,9 @@ stages:
                 container: mariner2
               ${{ elseif eq(parameters.builder.distro, 'azurelinux3') }}:
                 container: azurelinux3
-              ${{ else }}:
-                container: mariner1
```
dagood, line 79: Removing the `else` condition means that the jobs that would have hit this condition will run with no container. I believe the host machines are a recent Ubuntu. (This might be why `linux-amd64 test [opensslcrypto] FIPS` is failing.)

I think we should actually make mariner 2 the default, and we don't need `mariner2`-specific builders anymore.

https://github.com/microsoft/go/blob/f10f06228f690238ddbf07515edd06c1e8e0d20f/eng/pipeline/stages/go-builder-matrix-stages.yml is important here because it feeds into these conditions.

dagood (approved):

Issue #1626, opened 2025-04-09
TLSv1.3 causes panic on certified FIPS 140-3 linux instances


michel-laterman:
Using an ubuntu certified AMI (which has a [FIPS 140-3 certification](https://ubuntu.com/blog/fips-140-3-for-ubuntu-22-04lts)), and the latest microsoft/go (v1.24.2 at the time of reporting), operations that require TLSv1.3 cause a panic.

This can be tested with the following program:
```go
package main

import (
        "crypto/tls"
        "log"
        "net/http"
)

func main() {
        client := http.Client{
                Transport: &http.Transport{
                        TLSClientConfig: &tls.Config{
                                MinVersion: tls.VersionTLS13,
                        },
                },
        }

        resp, err := client.Get("https://www.google.com")
        if err != nil {
                panic(err)
        }
        log.Default().Printf("Request status: %d\n", resp.StatusCode)
}
```

When executed with `GOEXPERIMENT=systemcrypto CGO_ENABLED=1 go run main.go`, the following panic occurs:

```
panic: EVP_KDF_derive
    openssl error(s):
    error:1C800069:Provider routines::invalid key length
        ../providers/implementations/kdfs/hkdf.c:163

goroutine 5 [running]:
crypto/tls/internal/tls13.ExpandLabel[...](0xc0000258c8, {0xc0000b4620, 0x20, 0x20}, {0x72c473, 0x2}, {0x0, 0x0, 0x0}, 0xc)
    /usr/local/go/src/crypto/tls/internal/tls13/tls13.go:41 +0x27f
crypto/tls.(*cipherSuiteTLS13).trafficKey(0x9b7540, {0xc0000b4620, 0x20, 0x20})
    /usr/local/go/src/crypto/tls/key_schedule.go:29 +0x10d
crypto/tls.(*halfConn).setTrafficSecret(0xc000190220, 0x9b7540, 0xc0000b45e0?, {0xc0000b4620?, 0x20?, 0x72e255?})
    /usr/local/go/src/crypto/tls/conn.go:234 +0x67
crypto/tls.(*clientHandshakeStateTLS13).establishHandshakeKeys(0xc000025c48)
    /usr/local/go/src/crypto/tls/handshake_client_tls13.go:525 +0x374
crypto/tls.(*clientHandshakeStateTLS13).handshake(0xc000025c48)
    /usr/local/go/src/crypto/tls/handshake_client_tls13.go:134 +0x73e
[...]

PR #1654, opened 2025-04-24
Add TLS13-KDF support for openssl backend

nicholasberlin:
Relevant issue: https://github.com/microsoft/go/issues/1626

Add TLS13-KDF support
- Update golang-fips to [c494c21](https://github.com/golang-fips/openssl/commit/c494c216979153a2fc07afab57c0858f055ccdc0)
- Use new functions in openssl backend
- Add stub functions for other backends
- Minor documentation fix

### DIFF
--- a/patches/0004-Use-crypto-backends.patch
+++ b/patches/0004-Use-crypto-backends.patch
@@ -20,8 +20,8 @@ Subject: [PATCH] Use crypto backends
  src/crypto/cipher/ctr_aes_test.go             |   2 +-
  src/crypto/cipher/gcm_test.go                 |   2 +-
  src/crypto/des/cipher.go                      |   7 +
- src/crypto/dsa/boring.go                      | 113 +++++++++++
- src/crypto/dsa/dsa.go                         |  88 +++++++++
+ src/crypto/dsa/boring.go                      | 113 ++++++++++
+ src/crypto/dsa/dsa.go                         |  88 ++++++++
  src/crypto/dsa/notboring.go                   |  16 ++
  src/crypto/ecdh/ecdh.go                       |   2 +-
  src/crypto/ecdh/ecdh_test.go                  |   4 +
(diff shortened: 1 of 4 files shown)


patches/0004-Use-crypto-backends.patch
```diff
@@ -2452,6 +2457,19 @@ index 00000000000000..573896b9c1e6a8
 +// the underlying functions because the TLS 1.3 KDF does not have a standard of
 +// its own.
 +
++func checkFIPS140Only[Hash hash.Hash](h func() Hash, key []byte) error {
```
nicholasberlin: this function is copy-pasted from `./src/crypto/hkdf/hkdf.go`

any refactoring suggestions? export the function?

qmuntal: `tls13.ExpandLabel` is an internal API that's not reachable by the user, but only by `crypto/tls`. That package only uses FIPS-compliant parameters when calling `tls13.ExpandLabel`, so there is nothing to check here. You can remove this function.

nicholasberlin: thanks, force pushed an update: https://github.com/microsoft/go/pull/1654/commits/c76587805a1d197c3eb8a667229c868c0f4f9196

nicholasberlin:
@microsoft-github-policy-service agree company="Elastic"

qmuntal (changes requested):
Almost good, thanks!

nicholasberlin:
Ok, managed to "rebase".

qmuntal (approved):
LGTHM, thanks!

dagood (approved):

dagood:
In case that merge commit looks odd to anyone else--rest assured, it's a merge commit. I think I just got bit by a GitHub UI bug that applied the squash-style message to a merge commit. 😕

Issue #1658, opened 2025-04-29
1.24 backport: TLSv1.3 causes panic on certified FIPS 140-3 linux instances


dagood:
* https://github.com/microsoft/go/issues/1626
* https://github.com/golang-fips/openssl/pull/272
* https://github.com/microsoft/go/pull/1654

xnox:
Likely also needs https://github.com/golang-fips/openssl/pull/260

As that is what is causing the runtime errors most of the time.

nicholasberlin:
golang/openssl 1.24 backport PR: https://github.com/golang-fips/openssl/pull/276

nicholasberlin:
1.24 release branch PR: https://github.com/microsoft/go/pull/1662

dagood:
Thanks for the contributions! The `ms_tls13kdf` feature is now released in 1.24.3-1.

PR #1662, opened 2025-05-05
Backport TLS13-KDF to 1.24 release branch

nicholasberlin:
Issue: https://github.com/microsoft/go/issues/1658
 - Generally speaking, this back-ports https://github.com/microsoft/go/pull/1654 from main to the 1.24 release branch
---
 - Update the golang-fips/openssl vendor to https://github.com/golang-fips/openssl/commit/6020143833228ed05e0f2a570f1afb53d97427ad (which is the result of this [PR](https://github.com/golang-fips/openssl/pull/276) )
 - Adds a new build tag to gate the use of TLS13-KDF. This was accomplished by adding: `internal/backend/tls13kdf_default.go` and `internal/backend/tls13kdf_ms_kdftls13.go` in `patches/0003-Implement-crypto-internal-backend.patch`. The files are conditionally compiled via ~`//go:build goexperiment.opensslcrypto && !goexperiment.ms_tls13kdf && linux && cgo`~ `//go:build !ms_tls13kdf` and ~`//go:build goexperiment.opensslcrypto && goexperiment.ms_tls13kdf && linux && cgo`~ `//go:build ms_tls13kdf`, respectively. So, without adding the new build flag, `ms_tls13kdf`, `TLS13-KDF` will not be used.

### DIFF
--- a/patches/0003-Implement-crypto-internal-backend.patch
+++ b/patches/0003-Implement-crypto-internal-backend.patch
@@ -13,10 +13,10 @@ Subject: [PATCH] Implement crypto/internal/backend
  src/crypto/internal/backend/bbig/big_cng.go   |  12 +
  .../internal/backend/bbig/big_darwin.go       |  12 +
  .../internal/backend/bbig/big_openssl.go      |  12 +
- src/crypto/internal/backend/boring_linux.go   | 279 ++++++++++++++
- src/crypto/internal/backend/cng_windows.go    | 336 ++++++++++++++++
+ src/crypto/internal/backend/boring_linux.go   | 287 ++++++++++++++
+ src/crypto/internal/backend/cng_windows.go    | 344 ++++++++++++++++
  src/crypto/internal/backend/common.go         |  59 +++
- src/crypto/internal/backend/darwin_darwin.go  | 359 ++++++++++++++++++
+ src/crypto/internal/backend/darwin_darwin.go  | 367 ++++++++++++++++++
  src/crypto/internal/backend/fips140/boring.go |  11 +
  src/crypto/internal/backend/fips140/cng.go    |  33 ++
  src/crypto/internal/backend/fips140/darwin.go |  11 +
@@ -29,9 +29,11 @@ Subject: [PATCH] Implement crypto/internal/backend
  .../internal/opensslsetup/opensslsetup.go     |  70 ++++
  .../opensslsetup/opensslsetup_test.go         |  92 +++++
  .../backend/internal/opensslsetup/stub.go     |   8 +
- src/crypto/internal/backend/nobackend.go      | 240 ++++++++++++
- src/crypto/internal/backend/openssl_linux.go  | 331 ++++++++++++++++
+ src/crypto/internal/backend/nobackend.go      | 246 ++++++++++++
+ src/crypto/internal/backend/openssl_linux.go  | 339 ++++++++++++++++
  src/crypto/internal/backend/stub.s            |  10 +
+ .../internal/backend/tls13kdf_default.go      |   9 +
+ .../internal/backend/tls13kdf_ms_kdftls13.go  |   9 +
  src/go/build/deps_test.go                     |  27 +-
  .../exp_allowcryptofallback_off.go            |   9 +
  .../exp_allowcryptofallback_on.go             |   9 +
@@ -49,7 +51,7 @@ Subject: [PATCH] Implement crypto/internal/backend
  ...ckenderr_gen_requirefips_nosystemcrypto.go |  17 +
  .../backenderr_gen_systemcrypto_nobackend.go  |  16 +
  src/runtime/runtime_boring.go                 |   5 +
- 45 files changed, 2690 insertions(+), 3 deletions(-)
+ 47 files changed, 2746 insertions(+), 3 deletions(-)
  create mode 100644 src/crypto/internal/backend/backend_test.go
  create mode 100644 src/crypto/internal/backend/backendgen.go
  create mode 100644 src/crypto/internal/backend/backendgen_test.go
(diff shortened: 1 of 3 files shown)


qmuntal:
I like the approach taken in this PR to filter TLS 1.3 KDF on the backend package. After reviewing this PR, I now think that using a goexperiment is an overkill, better use a good old build tag to gate the new KDF.

nicholasberlin:
Should I change the string `goexperiment.ms_tls13kdf`?
Get rid of `goexperiment`?

dagood:
> After reviewing this PR, I now think that using a goexperiment is an overkill, better use a good old build tag to gate the new KDF.

My two cents... there's an odd tradeoff here for a one-major-version-only build tag:

* With a tag, you can upgrade from 1.24 to 1.25 without removing the tag and the build will still work. The tag won't do anything to the project at that point.
* If someone sets `GOEXPERIMENT=sytemcrypto,ms_tls13kdf` and they aren't using the right version of Go, it'd show `go: unknown GOEXPERIMENT ms_tls13kdf`, and that can help catch infra problems.

Another option would be a `godebug` option. Quite different, and more flexible in a lot of ways. I don't think they're intended by upstream to exist in one major version and be removed in the next, but I think the upgrade behavior would resemble GOEXPERIMENT and not be overly strange. (But maybe there's actually good reason to do this as an on-by-default `godebug` in 1.25, if we anticipate any problems?)

That said, I don't think it's unreasonable to start with a tag. (Simplest.)

> Should I change the string `goexperiment.ms_tls13kdf`?
> Get rid of `goexperiment`?

Yeah, to avoid the freshly untrue implication that the GOEXPERIMENT system is involved.

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -2671,6 +2709,45 @@ index 00000000000000..5e4b436554d44d
...
+index 00000000000000..8f13a59c4fc6c1
+--- /dev/null
++++ b/src/crypto/internal/backend/tls13kdf_ms_kdftls13.go
+@@ -0,0 +1,16 @@
++
++// Copyright 2023 The Go Authors. All rights reserved.
++// Use of this source code is governed by a BSD-style
++// license that can be found in the LICENSE file.
++
++//go:build goexperiment.opensslcrypto && goexperiment.ms_tls13kdf && linux && cgo
```
dagood: I feel it would be better to make this independent. Tying `ms_tls13kdf` to the return value of `SupportsTLS13KDF` means "splitting" `openssl_linux.go` into more files, making the backend system more complicated to maintain. I don't think `tls13kdf_default.go` and `tls13kdf_ms_kdftls13.go` need to care about `goexperiment.opensslcrypto`, `linux`, or `cgo`. Adding another check to `if boring.Enabled && boring.SupportsTLS13KDF() {` seems like it would also convey the intent more clearly.

(It's not as if setting this tag actually affects whether this algorithm is *supported*, it only means that the algorithm should be used if it *is* supported.)

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -2671,6 +2709,45 @@ index 00000000000000..5e4b436554d44d
...
++func SupportsTLS13KDF() bool {
++	return false
++}
+diff --git a/src/crypto/internal/backend/tls13kdf_ms_kdftls13.go b/src/crypto/internal/backend/tls13kdf_ms_kdftls13.go
+new file mode 100644
+index 00000000000000..8f13a59c4fc6c1
+--- /dev/null
++++ b/src/crypto/internal/backend/tls13kdf_ms_kdftls13.go
+@@ -0,0 +1,16 @@
++
```
dagood: Stray newline?

patches/0003-Implement-crypto-internal-backend.patch
nicholasberlin: Ok, reworked this. These files are now only controlled by `//go:build ms_tls13kdf`. And, they only set a new global boolean, `TLS13KDFEnabled`. `crypto/tls/internal/tls13/tls13.go` was updated to check the new boolean, before making a runtime support check.

This:
```
if boring.Enabled && boring.SupportsTLS13KDF() {
```
was changed to this:
```
if boring.Enabled && boring.TLS13KDFEnabled && boring.SupportsTLS13KDF() {
```

patches/0003-Implement-crypto-internal-backend.patch
nicholasberlin: 👍, gone now.

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -2671,6 +2713,36 @@ index 00000000000000..5e4b436554d44d
...
+@@ -0,0 +1,9 @@
++// Copyright 2023 The Go Authors. All rights reserved.
++// Use of this source code is governed by a BSD-style
++// license that can be found in the LICENSE file.
++
++//go:build ms_tls13kdf
++
++package backend
++
++const TLS13KDFEnabled = true
```
dagood, line 2745: Apologies if you preferred the original @qmuntal. I just saw in your comment that you might have. 😄

My semi-outside perspective is that this is more "use the TLS13KDF backend function for TLS 1.3", and there could in theory be other uses for the backend function, so it doesn't necessarily make sense to include it in `SupportsTLS13KDF` and a more direct reference is reasonable. But maybe that's not a reasonable way to look at it given the OpenSSL API's name matching the scenario exactly.

A middle ground would be to change OpenSSL's `SupportsTLS13KDF` to `return TLS13KDFEnabled && openssl.SupportsTLS13KDF()`. This would preserve the independent files and their maintainability, but incorporate the tag's const into `SupportsTLS13KDF` rather than handling it in `src/crypto/tls/internal/tls13/tls13.go`.

I'd be fine with either one.

dagood (approved):

qmuntal (approved):

nicholasberlin:
Thanks for the help!!

Issue #1648, opened 2025-04-22
crypto.NewGCMWithRandomNonce error for go >= 1.24 in FIPS mode


simitt:
## Problem Description
Calling [`crypto.NewGCMWithRandomNonce`](https://github.com/golang/go/blob/336626bac4c62b617127d41dccae17eed0350b0f/src/crypto/cipher/gcm.go#L93C6-L93C27) in FIPS mode returns an error.

When building a go program with the Microsoft build of Go and `GOEXPERIMENT=systemcrypto` and running it with `GOFIPS=1`, the following code
```
cipher, err := aes.NewCipher(key)
if err != nil {
  return  err
}
aesgcm, err := cipher.NewGCMWithRandomNonce(cipher)
if err != nil {
  return  err
}
```
leads to this error:
```
cipher: NewGCMWithRandomNonce requires aes.Block
```

The error seems to come from a concrete [type check](https://github.com/golang/go/blob/d584d2b3dda9ab88a51a3e9cffe8ecd140479cef/src/crypto/cipher/gcm.go#L94). When using `microsoft/go` in FIPS mode, the `cipher` is not of this concrete type. 

## Expectations
With the introduction of `crypto.NewGCMWithRandomNonce` as part of the go std-lib `crypto` package, `microsoft/go 1.24` should also support this functionality. 

## Related Issues
Similar issue to what was reported for usage with `boringcrypto` in https://github.com/golang/go/issues/72016.

qmuntal:
Thanks for reporting @simitt. We should at least fall back to native Go if the crypto backend doesn't implement a given function, so this issue looks like an oversights from our end. Will investigate a bit more and report back.

simitt:
Thanks for looking into it @qmuntal . 
This is as far as I investigated yet: With the patch in https://github.com/microsoft/go/blob/microsoft/release-branch.go1.24/patches/0004-Use-crypto-backends.patch#L320-L328 the `internal/backend` logic is applied. If a call to `boring.NewAESCipher(key)` is made by go, it really calls into `crypto/internal/backend.NewAESCipher(key)`.
```
diff --git a/src/crypto/aes/aes.go b/src/crypto/aes/aes.go
index 5bc2d13d673e0a..b803c77be62a66 100644
--- a/src/crypto/aes/aes.go
+++ b/src/crypto/aes/aes.go
@@ -15,7 +15,7 @@ package aes
 
 import (
 	"crypto/cipher"
-	"crypto/internal/boring"
[...]

PR #1673, opened 2025-05-12
Support backend ciphers in cipher.NewGCMWithRandomNonce

mertakman:
Solves #1648
Currently type check is very strict at NewGCMWithRandomNonce, it only supports stdlib's aes.Block. This is causing a big issue because in Microsoft build of Go we are trying to use this function with our custom ciphers.  This PR aims to use less strict interface when possible, and also when FIPS 140-only mode is not enforced.

### DIFF
--- a/patches/0003-Implement-crypto-internal-backend.patch
+++ b/patches/0003-Implement-crypto-internal-backend.patch
@@ -15,7 +15,7 @@ Subject: [PATCH] Implement crypto/internal/backend
  .../internal/backend/bbig/big_openssl.go      |  12 +
  src/crypto/internal/backend/boring_linux.go   | 294 ++++++++++++++
  src/crypto/internal/backend/cng_windows.go    | 348 +++++++++++++++++
- src/crypto/internal/backend/common.go         | 150 +++++++
+ src/crypto/internal/backend/common.go         |  65 ++++
  src/crypto/internal/backend/darwin_darwin.go  | 367 ++++++++++++++++++
  src/crypto/internal/backend/fips140/boring.go |  11 +
  src/crypto/internal/backend/fips140/cng.go    |  33 ++
@@ -49,7 +49,7 @@ Subject: [PATCH] Implement crypto/internal/backend
  ...ckenderr_gen_requirefips_nosystemcrypto.go |  17 +
  .../backenderr_gen_systemcrypto_nobackend.go  |  16 +
  src/runtime/runtime_boring.go                 |   5 +
- 45 files changed, 2837 insertions(+), 3 deletions(-)
+ 45 files changed, 2752 insertions(+), 3 deletions(-)
  create mode 100644 src/crypto/internal/backend/backend_test.go
  create mode 100644 src/crypto/internal/backend/backendgen.go
  create mode 100644 src/crypto/internal/backend/backendgen_test.go
@@ -554,7 +554,7 @@ index 00000000000000..e6695dd66b1d02
 +var Dec = bbig.Dec
 diff --git a/src/crypto/internal/backend/boring_linux.go b/src/crypto/internal/backend/boring_linux.go
 new file mode 100644
-index 00000000000000..5befae98deed4d
+index 00000000000000..dba1ac10532eef
 --- /dev/null
 +++ b/src/crypto/internal/backend/boring_linux.go
 @@ -0,0 +1,294 @@
@@ -620,7 +620,7 @@ index 00000000000000..5befae98deed4d
 +	if err != nil {
 +		return nil, err
 +	}
-+	return wrapGCMCapableBlock(cipher), nil
++	return cipher, nil
 +}
 +
 +func NewGCMTLS(c cipher.Block) (cipher.AEAD, error)   { return boring.NewGCMTLS(c) }
@@ -854,7 +854,7 @@ index 00000000000000..5befae98deed4d
 +}
 diff --git a/src/crypto/internal/backend/cng_windows.go b/src/crypto/internal/backend/cng_windows.go
 new file mode 100644
-index 00000000000000..21345775fe387c
+index 00000000000000..bb001dca9b7a0b
 --- /dev/null
 +++ b/src/crypto/internal/backend/cng_windows.go
 @@ -0,0 +1,348 @@
@@ -935,7 +935,7 @@ index 00000000000000..21345775fe387c
 +	if err != nil {
 +		return nil, err
 +	}
-+	return wrapGCMCapableBlock(cipher), nil
++	return cipher, nil
 +}
 +
 +func NewGCMTLS(c cipher.Block) (cipher.AEAD, error) {
@@ -1208,10 +1208,10 @@ index 00000000000000..21345775fe387c
 +}
 diff --git a/src/crypto/internal/backend/common.go b/src/crypto/internal/backend/common.go
 new file mode 100644
-index 00000000000000..6a85c9af5c5dd7
+index 00000000000000..a6786c5ad8e33d
 --- /dev/null
 +++ b/src/crypto/internal/backend/common.go
-@@ -0,0 +1,150 @@
+@@ -0,0 +1,65 @@
 +// Copyright 2022 The Go Authors. All rights reserved.
 +// Use of this source code is governed by a BSD-style
 +// license that can be found in the LICENSE file.
@@ -1219,12 +1219,10 @@ index 00000000000000..6a85c9af5c5dd7
 +package backend
 +
 +import (
-+	"crypto/cipher"
 +	"crypto/internal/backend/fips140"
 +	"crypto/internal/boring/sig"
 +	"internal/goexperiment"
 +	"runtime"
-+	"unsafe"
 +)
 +
 +func init() {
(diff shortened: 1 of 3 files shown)


patches/0004-Use-crypto-backends.patch
```diff
@@ -379,19 +380,54 @@ index 33942467784ad3..0282ffa9fa23c8 100644
...
++++ b/src/crypto/cipher/gcm.go
+@@ -93,7 +93,10 @@ func newGCM(cipher Block, nonceSize, tagSize int) (AEAD, error) {
+ func NewGCMWithRandomNonce(cipher Block) (AEAD, error) {
+ 	c, ok := cipher.(*aes.Block)
+ 	if !ok {
+-		return nil, errors.New("cipher: NewGCMWithRandomNonce requires aes.Block")
++		if fips140only.Enabled {
++			return nil, errors.New("crypto/cipher: use of GCM with non-AES ciphers is not allowed in FIPS 140-only mode")
++		}
++		return newGCMFallback(cipher, gcmStandardNonceSize, gcmTagSize)
```
qmuntal: `newGCMFallback` doesn't return an AEAD that implements a random nonce, so we can't use it in here. If we that, we would be causing a security vulnerability, as callers would expect the returned AEAD to use random nonces.

qmuntal (changes requested):

patches/0004-Use-crypto-backends.patch
```diff
@@ -379,19 +380,129 @@ index 33942467784ad3..0282ffa9fa23c8 100644
...
++
++// SealWithRandomNonce encrypts plaintext to out, and writes a random nonce to
++// nonce. nonce must be 12 bytes, and out must be 16 bytes longer than plaintext.
++// out and plaintext may overlap exactly or not at all. additionalData and out
++// must not overlap.
++//
++// This complies with FIPS 140-3 IG C.H Scenario 2.
++//
++// Note that this is NOT a [cipher.AEAD].Seal method.
++func SealWithRandomNonce(g AEAD, nonce, out, plaintext, additionalData []byte) {
```
qmuntal: This function should be unexported as this is a public package.

qmuntal: Add a comment saying that this function is copied from fips140/gcm

patches/0004-Use-crypto-backends.patch
```diff
@@ -379,19 +380,129 @@ index 33942467784ad3..0282ffa9fa23c8 100644
...
 -	"crypto/internal/boring"
-+	boring "crypto/internal/backend"
  	"crypto/internal/cryptotest"
  	"crypto/internal/fips140"
  	fipsaes "crypto/internal/fips140/aes"
+@@ -727,9 +726,6 @@ func testGCMAEAD(t *testing.T, newCipher func(key []byte) cipher.Block) {
+ 
+ 			// Test NewGCMWithRandomNonce.
+ 			t.Run("GCMWithRandomNonce", func(t *testing.T) {
+-				if _, ok := block.(*wrapper); ok || boring.Enabled {
```
qmuntal, line 519: Tests were not failing in your previous commit, which AFAIU it was not correct. I wouldn't merge this PR  without having a proper test that validates we are really generating a random nonce (which we should probably upstream).

qmuntal (changes requested):

spalatnik:
FYI, Golang 1.24 prevents using cipher.NewGC because it says that it is not FIPS compliant anymore

```
func NewGCM(cipher Block) (AEAD, error) {
	if fips140only.Enabled {
		return nil, errors.New("crypto/cipher: use of GCM with arbitrary IVs is not allowed in FIPS 140-only mode, use NewGCMWithRandomNonce")
	}
	return newGCM(cipher, gcmStandardNonceSize, gcmTagSize)
}
```

So `NewGCMWithRandomNonce` becomes the only viable option here. 

I'm only pointing this out because the initial commit had this - which has since been reverted - but I still want to leave this comment - 

```
+		if fips140only.Enabled {
+			return nil, errors.New("crypto/cipher: use of GCM with non-AES ciphers is not allowed in FIPS 140-only mode")
+		}
```

If that were to be put in, it would prevent any sort of AES encryption operations on this distribution in FIPS only mode using the standard libraries. Just an FYI.

mertakman:
> spalatnik

Thanks for information. Yes, I realized it just after submitting and reverted back.

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -2434,7 +2451,14 @@ index 00000000000000..ac656468e3ed32
...
 +
-+func NewAESCipher(key []byte) (cipher.Block, error)   { return openssl.NewAESCipher(key) }
++func NewAESCipher(key []byte) (cipher.Block, error) {
++	cipher, err := openssl.NewAESCipher(key)
++	if err != nil {
++		return nil, err
++	}
++	return cipher, nil
++}
++
```
qmuntal: These changes are not needed.

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -1260,6 +1271,12 @@ index 00000000000000..9436b00381aaf8
 +		}
 +	}
 +}
++
++const (
++	gcmBlockSize         = 16
++	gcmStandardNonceSize = 12
++	gcmTagSize           = 16
++)
```
qmuntal: Are you using these values?

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -924,7 +931,11 @@ index 00000000000000..6abb6ff007c99f
 +}
 +
 +func NewAESCipher(key []byte) (cipher.Block, error) {
-+	return cng.NewAESCipher(key)
++	cipher, err := cng.NewAESCipher(key)
++	if err != nil {
++		return nil, err
++	}
++	return cipher, nil
```
qmuntal: Undo these changes.

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -615,7 +615,14 @@ index 00000000000000..48d44ec1723ab6
 +
 +func NewHMAC(h func() hash.Hash, key []byte) hash.Hash { return boring.NewHMAC(h, key) }
 +
-+func NewAESCipher(key []byte) (cipher.Block, error)   { return boring.NewAESCipher(key) }
++func NewAESCipher(key []byte) (cipher.Block, error) {
++	cipher, err := boring.NewAESCipher(key)
++	if err != nil {
++		return nil, err
++	}
++	return cipher, nil
```
qmuntal: Undo these changes.

patches/0004-Use-crypto-backends.patch
```diff
@@ -379,19 +380,162 @@ index 33942467784ad3..0282ffa9fa23c8 100644
...
++		g, err := cipher.NewGCM(nonceSize, tagSize)
++		if err != nil {
++			return nil, err
++		}
++
++		if _, ok := g.(randomNonceSealer); ok {
++			return &customGCMWrapper{g}, nil
++		}
++
++		return g, nil
```
qmuntal: Are these changes needed?

patches/0004-Use-crypto-backends.patch
```diff
@@ -379,19 +380,162 @@ index 33942467784ad3..0282ffa9fa23c8 100644
...
+ 	}
+ 	if cipher.BlockSize() != gcmBlockSize {
+ 		return nil, errors.New("cipher: NewGCM requires 128-bit block cipher")
+@@ -220,6 +245,24 @@ func newGCMFallback(cipher Block, nonceSize, tagSize int) (AEAD, error) {
+ 	return &gcmFallback{cipher: cipher, nonceSize: nonceSize, tagSize: tagSize}, nil
+ }
+ 
++// customGCMWrapper wraps an AEAD implementation to expose only the AEAD interface methods.
++// This prevents extra exported methods from causing test failures in cryptotest.NoExtraMethods.
++type customGCMWrapper struct {
```
qmuntal, line 476: I think this type can be removed.

qmuntal (changes requested):

patches/0004-Use-crypto-backends.patch
mertakman: yes. They are needed for making ExtraMethods tests pass.

patches/0004-Use-crypto-backends.patch
mertakman: no. They are needed for making ExtraMethods tests pass.

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -615,7 +615,10 @@ index 00000000000000..48d44ec1723ab6
 +
 +func NewHMAC(h func() hash.Hash, key []byte) hash.Hash { return boring.NewHMAC(h, key) }
 +
-+func NewAESCipher(key []byte) (cipher.Block, error)   { return boring.NewAESCipher(key) }
++func NewAESCipher(key []byte) (cipher.Block, error) {
```
qmuntal: Keep previous formatting to reduce diffs.

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -2434,7 +2437,10 @@ index 00000000000000..ac656468e3ed32
 +
 +func NewHMAC(h func() hash.Hash, key []byte) hash.Hash { return openssl.NewHMAC(h, key) }
 +
-+func NewAESCipher(key []byte) (cipher.Block, error)   { return openssl.NewAESCipher(key) }
++func NewAESCipher(key []byte) (cipher.Block, error) {
```
qmuntal: Keep previous formatting to reduce diffs.

patches/0004-Use-crypto-backends.patch
```diff
@@ -379,19 +380,162 @@ index 33942467784ad3..0282ffa9fa23c8 100644
...
  	"encoding/hex"
+diff --git a/src/crypto/cipher/gcm.go b/src/crypto/cipher/gcm.go
+index 5580f96d55a0fb..cc759af19b116b 100644
+--- a/src/crypto/cipher/gcm.go
++++ b/src/crypto/cipher/gcm.go
+@@ -93,7 +93,14 @@ func newGCM(cipher Block, nonceSize, tagSize int) (AEAD, error) {
+ func NewGCMWithRandomNonce(cipher Block) (AEAD, error) {
+ 	c, ok := cipher.(*aes.Block)
+ 	if !ok {
+-		return nil, errors.New("cipher: NewGCMWithRandomNonce requires aes.Block")
```
qmuntal: We still need to return this error if the cipher doesn't implement `cipher` nor the gcm implements `randomNonceSealer.`

patches/0004-Use-crypto-backends.patch
```diff
@@ -379,19 +380,162 @@ index 33942467784ad3..0282ffa9fa23c8 100644
...
+@@ -157,7 +164,12 @@ func (g gcmWithRandomNonce) Seal(dst, nonce, plaintext, additionalData []byte) [
+ 		plaintext = ciphertext[:len(plaintext)]
+ 	}
+ 
+-	gcm.SealWithRandomNonce(g.GCM, nonce, ciphertext, plaintext, additionalData)
++	if gc, ok := g.aead.(*gcm.GCM); ok {
++		gcm.SealWithRandomNonce(gc, nonce, ciphertext, plaintext, additionalData)
++	} else if gc, ok := g.aead.(randomNonceSealer); ok {
++		gc.SealWithRandomNonce(ciphertext, nonce, plaintext, additionalData)
++	}
```
qmuntal, line 423: If you implement what I said in https://github.com/microsoft/go/pull/1673/files#r2108709392, then `g.aead` should always be either a `*gcm.GCM` or a `randomNonceSealer`. Therefore it is an error on our side if none of these cases are meet. I would add an else case that panics to be sure we haven't forgotten any case.

qmuntal (changes requested):

patches/0004-Use-crypto-backends.patch
```diff
@@ -379,19 +380,164 @@ index 33942467784ad3..0282ffa9fa23c8 100644
...
++		g, err := cipher.NewGCM(nonceSize, tagSize)
++		if err != nil {
++			return nil, err
++		}
++
++		if _, ok := g.(randomNonceSealer); ok {
++			return &customGCMWrapper{g}, nil
++		}
++
++		return nil, errors.New("cipher: NewGCMWithRandomNonce requires aes.Block")
```
qmuntal: `newGCMFallback` is not used in `NewGCMWithRandomNonce`, so all these changes are unnecessary.

qmuntal (changes requested):

patches/0004-Use-crypto-backends.patch
```diff
@@ -379,19 +380,166 @@ index 33942467784ad3..0282ffa9fa23c8 100644
...
+ 	}
+ 	if cipher, ok := cipher.(gcmAble); ok {
+-		return cipher.NewGCM(nonceSize, tagSize)
++		g, err := cipher.NewGCM(nonceSize, tagSize)
++		if err != nil {
++			return nil, err
++		}
++
++		if _, ok := g.(randomNonceSealer); ok {
++			return &customGCMWrapper{g}, nil
```
qmuntal: Does `customGCMWrapper` needs to be a pointer? This means we will be doing an additional allocation for every new GCM, which I would try to avoid.

patches/0004-Use-crypto-backends.patch
```diff
@@ -379,19 +380,166 @@ index 33942467784ad3..0282ffa9fa23c8 100644
...
+ 	c, ok := cipher.(*aes.Block)
+ 	if !ok {
+-		return nil, errors.New("cipher: NewGCMWithRandomNonce requires aes.Block")
++		if cipher, ok := cipher.(gcmAble); ok {
++			g, err := cipher.NewGCM(gcmStandardNonceSize, gcmTagSize)
++			if err != nil {
++				return nil, err
++			}
++
++			return gcmWithRandomNonce{g}, nil
```
qmuntal: We should only return `gcmWithRandomNonce` if g implements `randomNonceSealer`, else return the error down below.

qmuntal (changes requested):

patches/0004-Use-crypto-backends.patch
mertakman: yes. otherwise tests for additional methods fails.

qmuntal (approved):
LGTM!

patches/0004-Use-crypto-backends.patch
```diff
@@ -379,19 +380,164 @@ index 33942467784ad3..0282ffa9fa23c8 100644
...
+ 
+ func (g gcmWithRandomNonce) NonceSize() int {
+@@ -157,7 +166,14 @@ func (g gcmWithRandomNonce) Seal(dst, nonce, plaintext, additionalData []byte) [
+ 		plaintext = ciphertext[:len(plaintext)]
+ 	}
+ 
+-	gcm.SealWithRandomNonce(g.GCM, nonce, ciphertext, plaintext, additionalData)
++	if gc, ok := g.aead.(*gcm.GCM); ok {
++		gcm.SealWithRandomNonce(gc, nonce, ciphertext, plaintext, additionalData)
++	} else if gc, ok := g.aead.(randomNonceSealer); ok {
```
dagood: `gc` seems like an odd name? `SealWithRandomNonce` calls it `g`, and `gcm` would be another short name I'd expect, but not `gc`. I see that those are both already used, but I think we can do better by being more specific, maybe `aeadGCM`?

patches/0004-Use-crypto-backends.patch
```diff
@@ -379,19 +380,164 @@ index 33942467784ad3..0282ffa9fa23c8 100644
...
+-	_, err := g.GCM.Open(out[:0], nonce, ciphertext, additionalData)
++	_, err := g.aead.Open(out[:0], nonce, ciphertext, additionalData)
+ 	if err != nil {
+ 		return nil, err
+ 	}
+@@ -204,6 +220,10 @@ type gcmAble interface {
+ 	NewGCM(nonceSize, tagSize int) (AEAD, error)
+ }
+ 
++type randomNonceSealer interface {
```
dagood, line 445: Comment to explain the type, like `gcmAble` has above?

patches/0004-Use-crypto-backends.patch
```diff
@@ -379,19 +380,164 @@ index 33942467784ad3..0282ffa9fa23c8 100644
...
+ 		t.Run("NewGCMWithRandomNonce", func(t *testing.T) {
+ 			block := newCipher(make([]byte, 16))
+-			if _, ok := block.(*wrapper); ok || boring.Enabled {
++			if _, ok := block.(*wrapper); ok || goexperiment.BoringCrypto {
+ 				t.Skip("NewGCMWithRandomNonce requires an AES block cipher")
+ 			}
+ 			a, _ := cipher.NewGCMWithRandomNonce(block)
++			if a.NonceSize() != 0 {
++				t.Errorf("NewGCMWithRandomNonce returned AEAD with non-zero nonce size: %d", a.NonceSize())
++			}
```
dagood: Why is this `a.NonceSize() != 0` test happening in `NewGCMWithRandomNonce` too? `GCMWithRandomNonce` covers more key sizes, and I'm not spotting other differences.

On the other hand, it seems odd that `GCMWithRandomNonce` has the *real* coverage of `cipher.NewGCMWithRandomNonce`, rather than `NewGCMWithRandomNonce`.

For the sake of diff minimization, I think removing this (and leaving `GCMWithRandomNonce` as it is) makes sense, even though the test name doesn't line up. However, for upstreaming, think might need to be worked out better.

dagood (approved):
Basically just comment/naming nitpicks. Approving with a few suggestions.

gdams (approved):

qmuntal (approved):

PR #1684, opened 2025-05-29
Add appinsights telemetry

### LINKED ISSUE #1672: Add telemetry to the Go toolchain

qmuntal:
This PR makes all `go` executions upload telemetry data to Azure Application Insights as an opt-in feature which requires manually setting an instrumentation key. Therefore is safe to merge into main until we pass the privacy review.

Set this env variable to enable telemetry on your computed:

```bash
MS_GOTOOLCHAIN_TELEMETRY_INSTRUMENTATION_KEY=x # here you need a real instrumentation key!
MS_GOTOOLCHAIN_TELEMETRY_ALLOW_GO_DEVEL=1
MS_GOTOOLCHAIN_TELEMETRY_STATS=1 # optional, but this way you can see how much the telemetry data takes to upload
```

`MS_GOTOOLCHAIN_TELEMETRY_INSTRUMENTATION_KEY` will be removed once we have the final instrumentation key, which will be hardcoded.
`MS_GOTOOLCHAIN_TELEMETRY_ALLOW_GO_DEVEL` will probably stay, but won't be necessary for users outside our team.

I've done some local benchmarking, and the telemetry upload takes ~300ms, which is a reasonable amount of time.

Closes #1672.

Note for reviewers: you don't really need to review most of the new lines. Skip those under `src/cmd/vendor`, as they are `github.com/microsoft/go-infra/telemetry` being vendored.


patches/0010-add-appinsights-telemetry.patch
```diff
@@ -0,0 +1,1746 @@
...
+--- a/README.md
++++ b/README.md
+@@ -40,3 +40,7 @@ places to ask questions about the Go language.
+ 
+ [rf]: https://reneefrench.blogspot.com/
+ [cc4-by]: https://creativecommons.org/licenses/by/4.0/
++
++### Data Collection
++
++The software may collect information about you and your use of the software and send it to Microsoft. Microsoft may use this information to provide services and improve our products and services. You may turn off the telemetry by setting the `MS_GOTOOLCHAIN_TELEMETRY_ENABLED` environment variable to `off`. There are also some features in the software that may enable you and Microsoft to collect data from users of your applications. If you use these features, you must comply with applicable law, including providing appropriate notices to users of your applications together with a copy of Microsoft’s privacy statement. Our privacy statement is located at https://go.microsoft.com/fwlink/?LinkID=824704. You can learn more about data collection and use in the help documentation and our privacy statement. Your use of the software operates as your consent to these practices.
```
dagood: This doesn't seem to match the current behavior.

patches/0010-add-appinsights-telemetry.patch
```diff
@@ -0,0 +1,1746 @@
...
+ func NewStack(name string, depth int) dummyCounter                        { return dummyCounter{} }
+ func CountFlags(name string, flagSet flag.FlagSet)                        {}
+ func CountFlagValue(prefix string, flagSet flag.FlagSet, flagName string) {}
++
++func Close() {}
+diff --git a/src/cmd/vendor/github.com/microsoft/go-infra/telemetry/LICENSE b/src/cmd/vendor/github.com/microsoft/go-infra/telemetry/LICENSE
+new file mode 100644
+index 00000000000000..9e841e7a26e4eb
+--- /dev/null
++++ b/src/cmd/vendor/github.com/microsoft/go-infra/telemetry/LICENSE
```
dagood: Split into its own patch (like backends) for easier review on GitHub? Maybe even generalize the backend vendoring patch to instead be the "vendor everything that needs vendoring" patch?

patches/0010-add-appinsights-telemetry.patch
```diff
@@ -0,0 +1,1746 @@
...
++var appInsightsInstrumentationKey = os.Getenv("MS_GOTOOLCHAIN_TELEMETRY_INSTRUMENTATION_KEY") // TODO: hardcode to the correct value once we have a real key
++
+ var openCalled bool
+ 
+ func OpenCalled() bool { return openCalled }
+@@ -22,11 +30,29 @@ func OpenCalled() bool { return openCalled }
+ func Open() {
+ 	openCalled = true
+ 	counter.OpenDir(os.Getenv("TEST_TELEMETRY_DIR"))
++	if appInsightsInstrumentationKey != "" && os.Getenv("MS_GOTOOLCHAIN_TELEMETRY_ENABLED") != "off" {
```
dagood: A point of reference is the .NET SDK, which uses the term "OPTOUT" in `DOTNET_CLI_TELEMETRY_OPTOUT` and parses ["1" as well as "true" and "yes" insensitively](https://github.com/dotnet/sdk/blob/0934942c262f12b26007f1cd3bad148d360becf5/src/WebSdk/Publish/Tasks/EnvironmentHelper.cs#L23-L26).

I instinctively question a true-by-default bool... someone e.g. might see `MS_GOTOOLCHAIN_TELEMETRY_ENABLED=off` and remove it because it would be unnecessary if it were a "typical" off-by-default bool.

But: the world is full of true-by-default, I don't think this is really worth considering unless aligning with .NET is useful for some other reason. Thinking of this variable as matching with e.g. `CGO_ENABLED` and `GOGC` also seems reasonable to me.

dagood: I think we should accept 1 and 0, not (only) words, to avoid mistype/misread/misremember. Words should also not be case sensitive if we support them.

patches/0010-add-appinsights-telemetry.patch
qmuntal: How it differs? This is the standard data collection notice we have to add.

mertakman (approved):

patches/0010-add-appinsights-telemetry.patch
dagood: > You may turn off the telemetry by setting the `MS_GOTOOLCHAIN_TELEMETRY_ENABLED` environment variable to `off`.

This indicates (or maybe, technically, just heavily implies) an opt-out feature, but it's currently opt-in.

> There are also some features in the software that may enable you and Microsoft to collect data from users of your applications.

Are there? (Other than the fact that Go has a HTTP Client, which doesn't seem like the point. 😄) This would seem accurate to me in the microsoft/go README.md because when you build the repo, you end up with a Go that can send us telemetry, but I don't think it's true for apps built by our build of Go.

dagood: FWIW: I'm looking at it from a user's point of view. There are probably intentional generalities here that just look odd to me in "review mode". 🙂

patches/0010-add-appinsights-telemetry.patch
qmuntal: Good call. I've moved the cmd vendoring to the same patch file where we vendor the backends.

patches/0010-add-appinsights-telemetry.patch
qmuntal: Lets go for the 1 and 0 instead of words 👍 .

qmuntal: I like the `ENABLED` prefix more than the `OPTOUT`, probably because of the similarity with `CGO_ENABLED`, but don't have a strong opinion.

README.md
```diff
@@ -143,3 +143,7 @@ trademarks or logos is subject to and must follow
...
 Use of Microsoft trademarks or logos in modified versions of this project must not cause confusion or imply Microsoft sponsorship.
 Any use of third-party trademarks or logos are subject to those third-party's policies.
+
+## Data Collection
+
+The software may collect information about you and your use of the software and send it to Microsoft. Microsoft may use this information to provide services and improve our products and services. You may turn off the telemetry by setting the `MS_GOTOOLCHAIN_TELEMETRY_ENABLED` environment variable to `0`. There are also some features in the software that may enable you and Microsoft to collect data from users of your applications. If you use these features, you must comply with applicable law, including providing appropriate notices to users of your applications together with a copy of Microsoft’s privacy statement. Our privacy statement is located at https://go.microsoft.com/fwlink/?LinkID=824704. You can learn more about data collection and use in the help documentation and our privacy statement. Your use of the software operates as your consent to these practices.
```
dagood: I mentioned that "There are also some features in the software that may enable you and Microsoft to collect data from users of your applications" makes sense to me in the main readme rather than the build's readme, but I only meant that clause--I do think it makes sense (and we probably must) provide this notice in the product readme too.

Perhaps what we really need is for our automation to pack up the outer readme into the product tar.gzs... but until then, I think we might need to put it in both places manually.

dagood: Hmm, maybe wrap this to 80 columns like the rest of the readme for those `cat`ting to console?

patches/0001-Vendor-external-dependencies.patch
```diff
@@ -1,11 +1,34 @@
 From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
 From: Quim Muntal <qmuntaldiaz@microsoft.com>
 Date: Mon, 23 May 2022 12:59:36 +0000
-Subject: [PATCH] Vendor crypto backends
+Subject: [PATCH] Vendor external dependencies
 
 To reproduce changes in 'src/vendor', run 'go mod vendor' in 'src'.
+To reproduce changes in 'src/cmd/vendor', run 'go mod vendor' in 'src/cmd'.
```
dagood, line 7: While here, maybe include a reminder to `tidy`:

```suggestion
To reproduce changes in 'src/vendor', run 'go mod vendor' in 'src'.
To reproduce changes in 'src/cmd/vendor', run 'go mod vendor' in 'src/cmd'.
If 'go mod vendor' makes any changes, run 'go mod tidy'.
```

patches/0010-add-appinsights-telemetry.patch
```diff
@@ -0,0 +1,213 @@
...
++	appInsightsAddr = flag.String("appInsights", "", "run a fake App Insights server on this network address")
++	appInsightsURL  string
++)
++
++var appInsightsOnce sync.Once
++
++var appInsightsRequests atomic.Int64
++
++// StartAppInsightsServer starts a fake App Insights server running on *appInsightsAddr (like "localhost:1234")
++// and sets appInsightsURL to the MS_GOTOOLCHAIN_TELEMETRY_ENDPOINT setting to use to access the sever.
```
dagood:
```suggestion
+// and sets appInsightsURL to the MS_GOTOOLCHAIN_TELEMETRY_ENDPOINT setting to use to access the server.
```

dagood (approved):

dagood (approved):

PR #1686, opened 2025-06-02
Updated backends and vendored 

mertakman:
Recent changes requiring new versions of backends, temporarily we merged on place updates on vendor folder, but we need to fix this approach for having best practices. This PR updates all backends versions and vendors them.

### DIFF
--- a/patches/0002-Vendor-crypto-backends.patch
+++ b/patches/0002-Vendor-crypto-backends.patch
@@ -8,7 +8,7 @@ Use a 'go' that was recently built by the current branch to ensure stable result
 ---
  src/crypto/internal/backend/deps_ignore.go    |   22 +
  src/go.mod                                    |    6 +
- src/go.sum                                    |    6 +
+ src/go.sum                                    |   12 +
  src/go/build/deps_test.go                     |   35 +-
  src/go/build/vendor_test.go                   |    3 +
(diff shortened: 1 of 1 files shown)


patches/0002-Vendor-crypto-backends.patch
```diff
@@ -235,21 +239,27 @@ index 28e250668dd52c..e5aa4b223a54d4 100644
...
++	github.com/microsoft/go-crypto-winnative v0.0.0-20250527092650-fa74f148acbc
 +)
 diff --git a/src/go.sum b/src/go.sum
-index ca2e7027fad3bd..f623f220d18598 100644
+index ca2e7027fad3bd..f8a640dc9f26f4 100644
 --- a/src/go.sum
 +++ b/src/go.sum
-@@ -1,3 +1,9 @@
+@@ -1,3 +1,15 @@
 +github.com/golang-fips/openssl/v2 v2.0.4-0.20250424082917-c494c2169791 h1:5PZl6k34BcDxn4DT+f8GLx51OZBtmx22oQ01ygrbGt0=
```
qmuntal: There are stale (harmless) entries here. Run `go mod tidy` to remove them.

qmuntal (changes requested):

qmuntal (approved):

PR #1694, opened 2025-06-06
Add Clone interface for hashes in backends

### LINKED ISSUE #1681: Update the Clone method of our backend hash objects
Our backend hash objects implement a Clone method whose signature was defined before upstream standardized the `hash.Cloner` interface (which just happened yesterday): https://github.com/golang/go/issues/69521#issuecomment-2900024818. We had to do this without waiting upstream because nor CNG, nor Darwin, nor SCOSSL support serializing the hash object, and that is required to support the `crypto/tls` package.

The Cloner interface we defined is this:

```go
type Cloner interface {
    Clone() hash.Hash
}
```

The interface has been approved is this:

```go
type Cloner interface {
    Hash
    Clone() (Cloner, error)
}
```

This means that we need to update the backends so that the hash objects defined in there implement `hash.Cloner` properly. Note that the `Clone` method returns an error, but it really only allows returning `errors.ErrUnsupported` (see https://github.com/golang/go/issues/69521#issuecomment-2917367139), which don't apply to use. We should panic if the underlying crypto library returns any error.

Making this change can be considered a breaking change if someone is using our backends directly as a module, but the Clone methods are not part of the public API, one should type cast the `hash.Hash` to an interface that we haven't exported neither, so that should be fine.

The `hash.Cloner` interface is only defined in Go 1.25, but we still want to allow building our backends with older toolchains. To accomplish this, define a `HashCloner` interface that is an alias to `hash.Cloner` when building with Go 1.25 or greater. When building with Go 1.24 or lower, just make it look like upstream `hash.Cloner`.

Upstream also implemented the new `hash.Cloner` interface for the hash returned by `hmac.New`, so we need to make of hmacs also implement that interface. Take these CLs as reference: https://go-review.googlesource.com/c/go/+/675197, https://go-review.googlesource.com/c/go/+/675555.

Finally, note that CL 675197 added some new tests in `internal/testhash` to check the clone behavior. Our backends have a copy of that package, which should also be updated to accomodate the new tests.

mertakman:
Fixes #1681 
This PR introduces a Clone interface to our backend systems. This is to maintain compatibility with Go 1.25's newly introduced hash.Clone interface in the standard library.

Currently, our backend's Clone implementations don't return an error. Since the standard library's hash.Clone interface requires an error return, we now wrap the hash structures returned by our backends. This wrapper then includes a Clone method that correctly implements the hash.Clone interface, allowing us to incorporate the necessary error handling.

### DIFF
--- a/patches/0003-Implement-crypto-internal-backend.patch
+++ b/patches/0003-Implement-crypto-internal-backend.patch
@@ -7,16 +7,16 @@ Subject: [PATCH] Implement crypto/internal/backend
  .gitignore                                    |   2 +
  src/crypto/internal/backend/backend_test.go   |  30 ++
  src/crypto/internal/backend/backendgen.go     |  20 +
- .../internal/backend/backendgen_test.go       | 284 ++++++++++++
+ .../internal/backend/backendgen_test.go       | 284 +++++++++++++
  src/crypto/internal/backend/bbig/big.go       |  17 +
  .../internal/backend/bbig/big_boring.go       |  12 +
  src/crypto/internal/backend/bbig/big_cng.go   |  12 +
  .../internal/backend/bbig/big_darwin.go       |  12 +
  .../internal/backend/bbig/big_openssl.go      |  12 +
  src/crypto/internal/backend/boring_linux.go   | 287 +++++++++++++
- src/crypto/internal/backend/cng_windows.go    | 379 ++++++++++++++++
- src/crypto/internal/backend/common.go         |  59 +++
- src/crypto/internal/backend/darwin_darwin.go  | 404 ++++++++++++++++++
+ src/crypto/internal/backend/cng_windows.go    | 368 ++++++++++++++++
+ src/crypto/internal/backend/common.go         |  74 ++++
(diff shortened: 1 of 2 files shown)


patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -901,14 +901,49 @@ index 00000000000000..6abb6ff007c99f
...
++    if err != nil {
++        panic("cngcrypto: clone failed: " + err.Error())
++    }
++    return wrapHash(cloned), nil
++}
++
++func wrapHash(h hash.Hash) *hashWrapper {
++    cloner, ok := h.(interface{ Clone() (hash.Hash, error) })
++    if !ok {
++        panic("cngcrypto: hash does not implement Clone() (hash.Hash, error)")
```
mertakman: @qmuntal maybe I should remove the conversion in here, because we are panicking in clone method anyways?

qmuntal (changes requested):
PR description missing. The description should contain a link to the issue that reported this problem.

Will review the PR in more detail once CI is happy 😄.

PR #1698, opened 2025-06-13
Enable the `systemcrypto` goexperiment by default

### LINKED ISSUE #1352: Use systemcrypto by default
Extract from https://github.com/microsoft/go-lab/blob/main/docs/adr/0007-comply-microsoft-policy-by-default-with-go-toolchain.md:

We will update the toolchain to use the system crypto by default, removing the related experiments and implement the `nosystemcrypto` tag.

Additionally, we will implement the following migration plan for current GOEXPERIMENT users:

1. In Go 1.24, the default behavior will be to use the system crypto.
1. In Go 1.24, the `GOEXPERIMENT=systemcrypto`, `GOEXPERIMENT=opensslcrypto`, and `GOEXPERIMENT=cngcrypto` options will be deprecated.
   * When a deprecated `GOEXPERIMENT` value is used, `go build` will print a warning message to stderr, although the functionality will remain intact.
   * The new behavior will be announced via our blog, and each release note up to Go 1.25 will include a reminder about the upcoming changes.
1. In Go 1.25, the deprecated `GOEXPERIMENT` values will be removed, and a dedicated error message will be displayed.
1. In Go 1.26, the dedicated error message will be removed.

qmuntal:
As agreed in #1352, we should make `GOEXPERIMENT=systemcrypto` the default behavior. Users can opt-out from it using `GOEXPERIMENT=nosystemcrypto`.

Note that `systemcrypto` is only enabled by default on Windows and Linux. `darwincrypto` is still experimental, therefore this PR doesn't enable `systemcrypto` by default on `darwin`.

Closes #1352.

### DIFF
--- a/eng/doc/fips/README.md
+++ b/eng/doc/fips/README.md
@@ -139,7 +139,7 @@ See [the microsoft/go-images documentation][microsoft-go-images] for more inform
 ### Dockerfile env instruction
 
 > [!NOTE]
-> Since Go 1.25 `systemcrypto` is enabled by default on Windows and Linux, there is no need to set it anymore. See also [the Go 1.25 changelog](#go-125-aug-2025).
+> Since Go 1.25, `systemcrypto` is enabled by default on Linux and Windows. There is no need to manually enable using OpenSSL/CNG under the hood anymore. See also [the Go 1.25 changelog](#go-125-aug-2025).
 
 If you don't use the standard Go base images (e.g. your Dockerfile downloads the Microsoft build of Go manually), you can use an `env` instruction before the build instruction in your Dockerfile:
 
@@ -150,7 +150,7 @@ env GOEXPERIMENT=systemcrypto
 ### Modify the build command
 
 > [!NOTE]
-> Since Go 1.25 `systemcrypto` is enabled by default on Windows and Linux, there is no need to set it anymore. See also [the Go 1.25 changelog](#go-125-aug-2025).
+> Since Go 1.25, `systemcrypto` is enabled by default on Linux and Windows. There is no need to manually enable using OpenSSL/CNG under the hood anymore. See also [the Go 1.25 changelog](#go-125-aug-2025).
 
 Another approach that generally works for any build system is to modify the build command or build script. This section lists some helpful snippets to select a backend.
 
--- a/eng/doc/fips/README.md
+++ b/eng/doc/fips/README.md
@@ -68,7 +68,7 @@ There are typically two goals that lead to this document. Creating a FIPS compli
 > 1.24 introduces `GODEBUG=fips140=on` as the preferred way to enable FIPS mode. See also [the Go 1.24 changelog](#go-124-feb-2025).
 
 > [!NOTE]
-> Since Go 1.25 `systemcrypto` is enabled by default on Windows and Linux, there is no need to set it anymore. See also [the Go 1.25 changelog](#go-125-aug-2025).
+> Since Go 1.25, `systemcrypto` is enabled by default on Linux and Windows. There is no need to manually enable using OpenSSL/CNG under the hood anymore. See also [the Go 1.25 changelog](#go-125-aug-2025).
 
 | Build-time config | Runtime config | Internal Microsoft crypto policy | FIPS behavior |
 | --- | --- | --- | --- |
@@ -95,7 +95,7 @@ The `GOEXPERIMENT` environment variable is used at build time to select a crypto
 
 - `systemcrypto` automatically selects the suggested crypto backend for the target platform
    - Prior to Go 1.21, this experiment is not available and the backend must be selected manually
-   - Since Go 1.25, this experiment es enabled automatically on Windows and Linux. It can be disabled in the usual way: setting `GOEXPERIMENT=nosystemcrypto`
+   - Since Go 1.25, this experiment is enabled automatically on Windows and Linux. It can be disabled like any other `GOEXPERIMENT`: setting `GOEXPERIMENT=nosystemcrypto`
 - `opensslcrypto` selects OpenSSL, for Linux
 - `cngcrypto` selects CNG, for Windows
 - Since 1.24, `darwincrypto` selects CommonCrypto & CryptoKit for macOS
@@ -446,7 +446,7 @@ This list of major changes is intended for quick reference and for access to his
 
 ### Go 1.25 (Aug 2025)
 
-- Enable the `systemcrypto` goexperiment by default behavior on Windows and Linux. To disable it, set `GOEXPERIMENT=nosystemcrypto`.
+- The `systemcrypto` goexperiment is now enabled by default on Windows and Linux. To disable it, set `GOEXPERIMENT=nosystemcrypto`.
 
 - Running `go version -m` on a binary which uses a system crypto backend now shows the `microsoft_systemcrypto=1` build setting.
 
--- a/eng/doc/fips/README.md
+++ b/eng/doc/fips/README.md
@@ -68,7 +68,7 @@ There are typically two goals that lead to this document. Creating a FIPS compli
 > 1.24 introduces `GODEBUG=fips140=on` as the preferred way to enable FIPS mode. See also [the Go 1.24 changelog](#go-124-feb-2025).
 
 > [!NOTE]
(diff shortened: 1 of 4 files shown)


eng/doc/fips/README.md
```diff
@@ -67,13 +67,17 @@ There are typically two goals that lead to this document. Creating a FIPS compli
 >
 > 1.24 introduces `GODEBUG=fips140=on` as the preferred way to enable FIPS mode. See also [the Go 1.24 changelog](#go-124-feb-2025).
 
+> [!NOTE]
+> Since Go 1.25 `systemcrypto` is enabled by default on Windows and Linux, there is no need to set it anymore. See also [the Go 1.25 changelog](#go-125-aug-2025).
```
```suggestion
> Since Go 1.25, `systemcrypto` is enabled by default on Linux and Windows. There is no need to manually enable using OpenSSL/CNG under the hood anymore. See also [the Go 1.25 changelog](#go-125-aug-2025).
```

Fix missing comma and comma splice, and it also might be interesting to borrow language from the Microsoft crypto policy for a reader who got here from there. (So, rearrange platforms to match.)

eng/doc/fips/README.md
```diff
@@ -89,8 +93,9 @@ Some configurations are invalid and intentionally result in a build error or run
 
 The `GOEXPERIMENT` environment variable is used at build time to select a cryptographic library backend. This modifies the Go runtime included in the program to use the specified platform-provided cryptographic library whenever it calls a Go standard library crypto API. The `GOEXPERIMENT` values that pick a crypto backend are:
 
-- *Recommended:* `systemcrypto` automatically selects the suggested crypto backend for the target platform
-  - Prior to Go 1.21, this alias is not available and the backend must be selected manually
+- `systemcrypto` automatically selects the suggested crypto backend for the target platform
+   - Prior to Go 1.21, this experiment is not available and the backend must be selected manually
+   - Since Go 1.25, this experiment es enabled automatically on Windows and Linux. It can be disabled in the usual way: setting `GOEXPERIMENT=nosystemcrypto`
```
```suggestion
   - Since Go 1.25, this experiment is enabled automatically on Windows and Linux. It can be disabled like any other `GOEXPERIMENT`: setting `GOEXPERIMENT=nosystemcrypto`
```

Trying to answer the Q: "in the usual way *for what*?"

This could also be moved to "Usage: GOEXPERIMENTs and backend build tags", potentially.

eng/doc/fips/README.md
```diff
@@ -133,6 +138,9 @@ See [the microsoft/go-images documentation][microsoft-go-images] for more inform
 
 ### Dockerfile env instruction
 
+> [!NOTE]
+> Since Go 1.25 `systemcrypto` is enabled by default on Windows and Linux, there is no need to set it anymore. See also [the Go 1.25 changelog](#go-125-aug-2025).
```
dagood: ^

eng/doc/fips/README.md
```diff
@@ -141,6 +149,9 @@ env GOEXPERIMENT=systemcrypto
 
 ### Modify the build command
 
+> [!NOTE]
+> Since Go 1.25 `systemcrypto` is enabled by default on Windows and Linux, there is no need to set it anymore. See also [the Go 1.25 changelog](#go-125-aug-2025).
```
dagood: ^

eng/doc/fips/README.md
```diff
@@ -435,6 +446,8 @@ This list of major changes is intended for quick reference and for access to his
 
 ### Go 1.25 (Aug 2025)
 
+- Enable the `systemcrypto` goexperiment by default behavior on Windows and Linux. To disable it, set `GOEXPERIMENT=nosystemcrypto`.
```
dagood: Wording seems out of place:

```suggestion
- The `systemcrypto` goexperiment is now enabled by default on Windows and Linux. To disable it, set `GOEXPERIMENT=nosystemcrypto`.
```

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -342,8 +373,39 @@ index 3e691abe41f702..b799031d598892 100644
...
++	"slices"
+ 	"strings"
+ 	"sync"
+ 	"time"
+@@ -232,6 +233,18 @@ func ForceHost() {
+ 	// Recompute experiments: the settings determined depend on GOOS and GOARCH.
+ 	// This will also update the BuildContext's tool tags to include the new
+ 	// experiment tags.
++	//
++	// Remove systemcrypto from the RawGOEXPERIMENT and add instead
```
```suggestion
+	// Remove systemcrypto from the RawGOEXPERIMENT and add
```

"Instead" implies that it will only add `nosystemcrypto` if it actually found and removed `systemcrypto`.

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -342,8 +373,39 @@ index 3e691abe41f702..b799031d598892 100644
...
++	// nosystemcrypto so that the tool is built without systemcrypto,
++	// as the rest of the toolchain.
```
```suggestion
+	// as is the rest of the toolchain.
```

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -342,8 +373,39 @@ index 3e691abe41f702..b799031d598892 100644
...
++	goexp := strings.Split(RawGOEXPERIMENT, ",")
++	goexp = slices.DeleteFunc(goexp, func(s string) bool {
++		return s == "systemcrypto" || s == "nosystemcrypto" || s == ""
++	})
++	goexp = append(goexp, "nosystemcrypto")
++	RawGOEXPERIMENT = strings.Join(goexp, ",")
++	// Set the ms_skipfipscheck build tag, as the rest of the toolchain.
```
```suggestion
+	// Set the ms_skipfipscheck build tag, like the rest of the toolchain.
```

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -342,8 +373,39 @@ index 3e691abe41f702..b799031d598892 100644
...
+ 	// experiment tags.
++	//
++	// Remove systemcrypto from the RawGOEXPERIMENT and add instead
++	// nosystemcrypto so that the tool is built without systemcrypto,
++	// as the rest of the toolchain.
++	goexp := strings.Split(RawGOEXPERIMENT, ",")
++	goexp = slices.DeleteFunc(goexp, func(s string) bool {
++		return s == "systemcrypto" || s == "nosystemcrypto" || s == ""
++	})
++	goexp = append(goexp, "nosystemcrypto")
```
dagood: It seems like this could potentially be shared between `cmd/go/internal/cfg/cfg.go`, `cmd/dist/build.go`, and `cmd\link\internal\ld\main.go`, but... it isn't all *that* much that would be shared vs. the overhead in adding a new cmd/internal package. (Does `cmd` have those strict package dependency tests?) I'm ok with the copy.

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -356,6 +418,31 @@ index e913f9885234fd..ef331980a816a3 100644
...
+ 
+ 	// Ignore go.mod and go.work: we don't need them, and we want to be able
+ 	// to run the tool even if there's an issue with the module or workspace the
+@@ -288,6 +292,9 @@ func buildAndRunBuiltinTool(ctx context.Context, toolName, tool string, args []s
+ 	modload.RootMode = modload.NoRoot
+ 
+ 	runFunc := func(b *work.Builder, ctx context.Context, a *work.Action) error {
++		// Now that the tool is built, revert the GOEXPERIMENT
++		// environment variable to the original value.
++		os.Setenv("GOEXPERIMENT", originalGOEXPERIMENT)
```
dagood: Does this func ever get called in parallel? I assume not, but if so, seems like it would be easy to get "stuck" with a GOEXPERIMENT value and end up with flakiness.

dagood (approved):

patches/0003-Implement-crypto-internal-backend.patch
qmuntal: I don't think it can, but `runBuiltTool` accepts an environment variable slice that is then passed to the run command. We can use that to discard this potential issue.

patches/0003-Implement-crypto-internal-backend.patch
qmuntal: Even better, I've removed all this duplication by just taking the tags and experiments from the toolchain binary 😄.

Note that the build tags in particular should already be honored, but they aren't due to a limitation on how built-in tools are built on demand. I've filled an upstream bug to track this: https://github.com/golang/go/issues/74620.

gdams (approved):

PR #1701, opened 2025-06-16
Downgrade `GOEXPERIMENT=allowcryptofallback` to a build tag

qmuntal:
The `allowcryptofallback` was a mechanism to facilitate upgrading to Go 1.21. In that release we stopped falling back to go crypto by default, and the `allowcryptofallback` restored the old behavior (hurting compliance, ofc).

Go 1.21 is not supported anymore, so that use case is no longer relevant. I did a cursory search in GitHub and AzDO and found no real uses of `GOEXPERIMENT=allowcryptofallback` other than in this repo.

Given these facts, we can downgrade it to a simple build tag, which has less visibility. I've also removed most of the related mentions from our FIPS guide, as we were almost spending more time describing how to use `GOEXPERIMENT=allowcryptofallback` than explaining how to be compliant.

Note that our CI code no longer sets `allowcryptofallback` globally, but on a case by case basis. This mainly because build tags are not passed around as good as goexperiments in `go tool dist test`. Instead of making sure they are properly honored I took the approach of being explicit about which tests require `allowcryptofallback` so that we can fix them upstream.

This PR helps simplifying our code and docs in preparation for #1352.

### DIFF
--- a/eng/doc/fips/README.md
+++ b/eng/doc/fips/README.md
@@ -84,7 +84,7 @@ Some configurations are invalid and intentionally result in a build error or run
 | `GOEXPERIMENT=systemcrypto` and `-tags=requirefips` | `GOFIPS=0` | The app panics due to the conflict between build-time and runtime configuration. |
 | `-tags=requirefips` | | The build fails. A crypto backend must be specified to enable FIPS features. |
 | `GOEXPERIMENT=cngcrypto,opensslcrypto` | | The build fails. Only one crypto backend can be enabled at a time. |
-| `GOOS=linux CGO_ENABLED=0 GOEXPERIMENT=systemcrypto` | | The build fails. Cgo is required to use the OpenSSL backend. <br/> Prior to Go 1.21 or if 
+| `GOOS=linux CGO_ENABLED=0 GOEXPERIMENT=systemcrypto` | | The build fails. Cgo is required to use the OpenSSL backend. <br/> Prior to Go 1.21 the build would succeed but use standard Go crypto, making the app non-compliant. |
 
 ## Usage: Build
 
--- a/eng/doc/fips/README.md
+++ b/eng/doc/fips/README.md
@@ -50,7 +50,6 @@ These are described in the following sections in detail.
   - [`goexperiment.<backend>crypto` build tag](#usage-build)
   - [`requirefips` build tag](#build-option-to-require-fips-mode)
   - [`GOFIPS140=latest` environment variable](#build-option-to-require-fips-mode) (go1.24+)
-  - [`GOEXPERIMENT` `allowcryptofallback`](#build-option-to-use-go-crypto-if-the-backend-compatibility-check-fails)
   - [`import _ "crypto/tls/fipsonly"` source change](#tls-with-fips-compliant-settings)
 - Runtime configuration:
   - [`GOFIPS` environment variable](#usage-runtime)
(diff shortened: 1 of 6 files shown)


gdams (approved):

dagood:
> Given these facts, we can downgrade it to a simple build tag, which has less visibility. I've also removed most of the related mentions from our FIPS guide, as we were almost spending more time describing how to use `GOEXPERIMENT=allowcryptofallback` than explaining how to be compliant.

I think that while we're renaming it, we should find something to call it rather than "fallback" to further avoid confusion with the per-algo fallback from platform crypto to Go crypto.

> The `allowcryptofallback` was a mechanism to facilitate upgrading to Go 1.21.

I don't agree with this entirely--it's true that it restores an old behavior, however that behavior could be useful in certain CI scenarios and there is no reason to stop using it just because someone's upgraded past 1.21. That said, I don't think it's a feature that we must continue supporting. (Or rather: continue making it easy to use.)

dagood:
> > The `allowcryptofallback` was a mechanism to facilitate upgrading to Go 1.21.
> 
> I don't agree with this entirely--it's true that it restores an old behavior, however that behavior could be useful in certain CI scenarios and there is no reason to stop using it just because someone's upgraded past 1.21

To refine this a bit: I would say it's not only about compat with pre-1.21, but also about compat with upstream Go in existing CI.

qmuntal:
> I think that while we're renaming it, we should find something to call it rather than "fallback" to further avoid confusion with the per-algo fallback from platform crypto to Go crypto.

Good point. Let's get this merge and then come up with a better name. The PR is already becoming too big.

> To refine this a bit: I would say it's not only about compat with pre-1.21, but also about compat with upstream Go in existing CI.

Agree. I was talking more from the user perspective. AFAIK we added this feature to support migrations. We then used it for some thing more internally.

eng/doc/fips/README.md
```diff
@@ -458,6 +437,10 @@ A program running in FIPS mode can claim it is using a FIPS-certified cryptograp
 
 This list of major changes is intended for quick reference and for access to historical information about versions that are no longer supported. The behavior of all in-support versions are documented in the sections above with notes for version-specific differences where necessary.
 
+### Go 1.25 (Aug 2025)
+
+- `GOEXPERIMENT=allowcryptofallback` has been downgraded to the `allowcryptofallback` build tag. Applications shouldn't be using it anymore, as it was a mechanism to facilitate migrating to Go 1.21. The build tag is still there for rare exceptions, please contact the Go team at Microsoft if you need to use it so we can help you migrate to the new behavior.
```
```suggestion
- `GOEXPERIMENT=allowcryptofallback` has been downgraded to the `allowcryptofallback` build tag. This is an internal mechanism that is not intended for use when building a Go application. This document has always recommended against using it, so we anticipate that this change won't affect users of the Microsoft build of Go. Please [contact the maintainers of the Microsoft build of Go](https://github.com/microsoft/go/blob/microsoft/main/SUPPORT.md) if you need to use it so we can understand the scenario and help find a safer alternative.
```

dagood (commented):
> AFAIK we added this feature to support migrations. We then used it for some thing more internally.

I poked around, and funnily enough, it [started as a tag that we used to fix CI](https://github.com/microsoft/go/pull/965), and we changed it to be an experiment later to make it flow more sanely to the right places in CI. We did [consider migrations later](https://github.com/microsoft/go/pull/982#discussion_r1263783048), but [decided not to suggest it even in that case](https://github.com/microsoft/go/pull/984#discussion_r1264201144) because it somewhat defeats the point. (The error message only suggested removing `opensslcrypto`.)  When we [documented it](https://github.com/microsoft/go/pull/970), we didn't mention its potential use for migration.

Someone might have read far between the lines and seen it as a way to migrate (or perhaps a way to avoid migrating! 😄) but I wouldn't say it was our intent.

Telemetry sure will be handy in the future if we can see if people are using these kinds of things!

dagood (approved):

PR #1704, opened 2025-06-19
Move backend error files to the `crypto` package

qmuntal:
We will soon enable `systemcrypto` by default (see #1352). We currently require cgo to be enabled when building with systemcrypto on Linux and Darwin, even when applications don't use crypto at all. That made sense when systemcrypto was opt-in, as it makes the behavior more consistent.

Now that systemcrypto will be the default, that restriction puts an unnecessary burden on applications that don't use crypto. For example, they can't set `CGO_ENABLE=0` and they can't cross-compile without manually setting `CGO_ENABLED=1` (as cgo is automatically disabled when cross-compiling).

To avoid this pain we better only require cgo if crypto is really used.

Note that most of the `allowcryptofallback` build tag used in tests can be removed thanks to this change.

While here, I've changed the backend error files name prefix from `backenderr_gen_` to `zbackenderr`. Starting autogenerated files with a `z` is more idiomatic.

### DIFF
--- a/patches/0003-Implement-crypto-internal-backend.patch
+++ b/patches/0003-Implement-crypto-internal-backend.patch
@@ -5,7 +5,7 @@ Subject: [PATCH] Implement crypto/internal/backend
 
 ---
  .gitignore                                    |   2 +
- src/cmd/dist/test.go                          |   6 +
+ src/cmd/dist/test.go                          |   7 +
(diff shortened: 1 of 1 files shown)


dagood (approved):
FIPS doc needs updates to match, but can be done any time before 1.25.0.

PR #1709, opened 2025-06-26
Use `skipfipscheck` instead of `allowcryptofallback` (sometimes)

qmuntal:
`allowcryptofallback` is overriden with too many duties (and the name doesn't fit any of them). One of those duties is to allow executing the Go toolchain with `GODEBUG=fips140=on` even when the toolchain is not built with a system crypto backend or when the system doesn't support FIPS mode. This is because the user doesn't care about the toolchain being executed in FIPS mode (which is in fact not officially supported), but to run the binary built by the toolchain in FIPS mode.

This PR stops using the`allocryptofallback` build tag for this purpose, and starts using the new `skipfipscheck` build tag instead. This build tag is not documented as it is supposed to only be used by the Go toolchain (we might revisit this decision if there is a compelling 1P use case).

### DIFF
--- a/patches/0003-Implement-crypto-internal-backend.patch
+++ b/patches/0003-Implement-crypto-internal-backend.patch
@@ -127,7 +127,7 @@ index b50f3342fe8714..23a7f132788ebc 100644
  	goInstall(toolenv(), goBootstrap, append([]string{"-pgo=off"}, toolchain...)...)
  	if debug {
 diff --git a/src/cmd/dist/test.go b/src/cmd/dist/test.go
-index d335e4cfbce468..7e362c80be609a 100644
+index d335e4cfbce468..f8abfe6f62216f 100644
 --- a/src/cmd/dist/test.go
 +++ b/src/cmd/dist/test.go
 @@ -847,6 +847,11 @@ func (t *tester) registerTests() {
@@ -137,7 +137,7 @@ index d335e4cfbce468..7e362c80be609a 100644
 +		var tags []string
 +		if goos != "windows" {
 +			// Linux and macOS don't support systemcrypto with CGO_ENABLED=0.
-+			tags = append(tags, "allowcryptofallback")
++			tags = append(tags, "allowcryptofallback,skipfipscheck")
 +		}
  		t.registerTest("internal linking, -buildmode=pie",
  			&goTest{
--- a/patches/0003-Implement-crypto-internal-backend.patch
+++ b/patches/0003-Implement-crypto-internal-backend.patch
@@ -3,14 +3,19 @@ From: qmuntal <qmuntaldiaz@microsoft.com>
 Date: Wed, 15 Jan 2025 16:32:26 +0100
 Subject: [PATCH] Implement crypto/internal/backend
 
+Implement the supporting code for using the crypto backends,
+but without using them yet.
+
+Update the Go toolchain build settings so that it uses the
+desired goexperiments and build tags.
 ---
  .gitignore                                    |   2 +
- src/cmd/dist/build.go                         |   6 +
+ src/cmd/dist/build.go                         |   5 +
  src/cmd/dist/test.go                          |  17 +-
- src/cmd/go/systemcrypto_test.go               |  92 +++++
+ src/cmd/go/systemcrypto_test.go               |  97 +++++
  src/crypto/internal/backend/backend_test.go   |  30 ++
  src/crypto/internal/backend/backendgen.go     |  20 +
- .../internal/backend/backendgen_test.go       | 278 +++++++++++++
+ .../internal/backend/backendgen_test.go       | 288 ++++++++++++++
  src/crypto/internal/backend/bbig/big.go       |  17 +
  .../internal/backend/bbig/big_boring.go       |  12 +
  src/crypto/internal/backend/bbig/big_cng.go   |  12 +
@@ -48,11 +53,12 @@ Subject: [PATCH] Implement crypto/internal/backend
  src/crypto/zbackenderrnofallback_darwin.go    |  20 +
  src/crypto/zbackenderrnofallback_openssl.go   |  20 +
  .../zbackenderrrequirefips_nosystemcrypto.go  |  17 +
+ .../zbackenderrrequirefips_skipfipscheck.go   |  16 +
  .../zbackenderrsystemcrypto_nobackend.go      |  16 +
  src/go/build/deps_test.go                     |  27 +-
  src/runtime/runtime_boring.go                 |   5 +
  src/syscall/exec_linux_test.go                |   2 +-
- 48 files changed, 2829 insertions(+), 8 deletions(-)
+ 49 files changed, 2859 insertions(+), 8 deletions(-)
  create mode 100644 src/cmd/go/systemcrypto_test.go
  create mode 100644 src/crypto/internal/backend/backend_test.go
  create mode 100644 src/crypto/internal/backend/backendgen.go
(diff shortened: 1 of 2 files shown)


patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -106,68 +109,180 @@ index c6512e64a4ef39..b3b01db73b009d 100644
...
++++ b/src/cmd/dist/build.go
+@@ -1539,6 +1539,12 @@ func cmdbootstrap() {
+ 	os.Setenv("CC", compilerEnvLookup("CC", defaultcc, goos, goarch))
+ 	// Now that cmd/go is in charge of the build process, enable GOEXPERIMENT.
+ 	os.Setenv("GOEXPERIMENT", goexperiment)
++	// Build the Go toolchain with "-tags=allowcryptofallback,skipfipscheck". This
++	// allows toolchains not built with "GOEXPERIMENT=systemcrypto" to be used
++	// when GODEBUG=fips140=on is set. For example, when running
++	// "GODEBUG=fips140=on go test ./..." or "GODEBUG=fips140=on go run .".
++	// Shadow goexperiment so that the global variable is not modified.
```
dagood: The line isn't new, but I'm a bit lost about what shadowing goexperiment is referring to here...

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -106,68 +109,180 @@ index c6512e64a4ef39..b3b01db73b009d 100644
...
++	os.Setenv("GOFLAGS", "-tags=allowcryptofallback,skipfipscheck")
```
dagood: Why is this moving to 0003, vs the "right patch" 0002 in

* https://github.com/microsoft/go/pull/1700 ?

Is it because it's not an experiment anymore, and the previous PR just missed the move?

Maybe add a patch description that describes a little what should be here?

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -106,68 +109,180 @@ index c6512e64a4ef39..b3b01db73b009d 100644
...
+diff --git a/src/cmd/dist/build.go b/src/cmd/dist/build.go
+index b50f3342fe8714..23a7f132788ebc 100644
+--- a/src/cmd/dist/build.go
++++ b/src/cmd/dist/build.go
+@@ -1539,6 +1539,12 @@ func cmdbootstrap() {
+ 	os.Setenv("CC", compilerEnvLookup("CC", defaultcc, goos, goarch))
+ 	// Now that cmd/go is in charge of the build process, enable GOEXPERIMENT.
+ 	os.Setenv("GOEXPERIMENT", goexperiment)
++	// Build the Go toolchain with "-tags=allowcryptofallback,skipfipscheck". This
++	// allows toolchains not built with "GOEXPERIMENT=systemcrypto" to be used
```
dagood: Seems outdated now that systemcrypto is on by default. Being more direct keeps it simple though:

```suggestion
+	// allows toolchains not built with the systemcrypto goexperiment to be used
```

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -106,68 +109,180 @@ index c6512e64a4ef39..b3b01db73b009d 100644
...
++	// when GODEBUG=fips140=on is set. For example, when running
```
dagood, line 129: Won't the build also run into problems if `systemcrypto` is enabled, but the system is not in FIPS mode? (Even a problem if it's FIPS-capable, because we will panic rather than try to turn it on.)

(Edit: I think I misunderstood what "this" referred to... wasn't considering the `skipfipscheck` for some reason.)

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -1945,9 +2094,41 @@ index 00000000000000..60ed26591d05d8
...
+new file mode 100644
+index 00000000000000..8509e61c588df7
+--- /dev/null
++++ b/src/crypto/internal/backend/fips140/skipfipscheck_on.go
+@@ -0,0 +1,9 @@
++// Copyright 2025 The Go Authors. All rights reserved.
++// Use of this source code is governed by a BSD-style
++// license that can be found in the LICENSE file.
++
++//go:build skipfipscheck
```
dagood: Consider `ms_` prefix. I'm also wondering if it's possible to indicate through the name that it's skipping the backend-side FIPS check--but I think the `ms_` prefix is likely sufficient to point people in the right direction.

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -106,68 +109,180 @@ index c6512e64a4ef39..b3b01db73b009d 100644
...
++
++// writeFile creates a temporary file with the given content and returns its path.
++func writeFile(t *testing.T, content string) string {
++	t.Helper()
++	name := "main.go"
++	if strings.Contains(content, "testing") {
++		name = "main_test.go"
++	}
++	name = filepath.Join(t.TempDir(), name)
++	if err := os.WriteFile(name, []byte(content), 0644); err != nil {
```
```suggestion
+	if err := os.WriteFile(name, []byte(content), 0o644); err != nil {
```

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -106,68 +109,180 @@ index c6512e64a4ef39..b3b01db73b009d 100644
...
++
++	cryptoFile := writeFile(t, fileWithCrypto)
++	nonCryptoFile := writeFile(t, fileWithoutCrypto)
++
++	// Build should always succeed given that the go toolchain
++	// is built with the skipfipscheck build tag.
++	execGoTool(t, "build", env, cryptoFile)
++	execGoTool(t, "build", env, nonCryptoFile)
++
++	// Run should always succed if the target go package
```
```suggestion
+	// Run should always succeed if the target go package
```

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -106,68 +109,180 @@ index c6512e64a4ef39..b3b01db73b009d 100644
...
++	// doesn't use crypto.
++	execGoTool(t, "run", env, nonCryptoFile)
++}
```
dagood, line 296: Even if we can't test the negative because we can't control the environment well enough, IMO it's worth leaving a comment to indicate what the negative case is. (I think knowing what *would* be tested helps clarify the positive case and intended behavior.)

```suggestion
+ // Run may or may not fail if the target go package uses crypto,
+ // because while the toolchain was built with skipfipscheck, the
+ // target program was not. Failure depends on system crypto mode
+ // and presence of system-provided crypto, and it can't be tested here.
+}
```

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -1789,32 +1913,53 @@ index 00000000000000..b11251ad4c6c9c
...
-+		Message = "system FIPS mode"
++		message = "system FIPS mode"
 +		value = "1"
 +	}
 +	enabled = value == "1"
 +	if isRequireFIPS {
-+		Message = "requirefips tag set"
++		if isSkipFIPSCheck {
++			panic("the 'requirefips' build tag is enabled, but it conflicts " +
++				"with the 'skipfipscheck' build tag")
```
dagood: Consider adding a file to make this a build error?

dagood (approved):
Nice, just a few suggestions and comments about comments.

patches/0003-Implement-crypto-internal-backend.patch
qmuntal: Good catch, this line is not needed now (I'll probably reintroduce it in a follow-up PR, though).

patches/0003-Implement-crypto-internal-backend.patch
qmuntal: allowcryptofallback is not an experiment anymore, so I moved it back to the previous patch. Sorry for the back and fort 😅 .

Will add a commit description 👍.

patches/0003-Implement-crypto-internal-backend.patch
qmuntal: Makes sense!

patches/0003-Implement-crypto-internal-backend.patch
qmuntal: Like this prefix. Will do.

patches/0003-Implement-crypto-internal-backend.patch
qmuntal: Done!

PR #1712, opened 2025-06-30
Add `microsoft_systemcrypto` setting

### LINKED ISSUE #963: Add `debug.BuildSetting` entries corresponding to crypto backend usage

### LINKED ISSUE #1706: FIPS doc needs updates to match backend error files move before 1.25.0.
FIPS doc needs updates to match, but can be done any time before 1.25.0.

_Originally posted by @dagood in https://github.com/microsoft/go/pull/1704#pullrequestreview-2950765064_

qmuntal:
It is not trivial to determine if a given Go binary uses systemcrypto or not, as it depends on multiple build settings. It would be even more difficult after #1352.

We can solve this issue by extending  the build info hardcoded in a Go binary to also contain whether systemcrypto is enabled. 

Example:

```
go version -m ./main
./main: go1.25-devel_04738e85ef2 Thu Jun 12 12:49:08 2025 +0200
        path    command-line-arguments
        build   microsoft_systemcrypto=1 // <----- New setting
        build   -buildmode=exe
        build   -compiler=gc
        build   CGO_ENABLED=1
        build   CGO_CFLAGS=
        build   CGO_CPPFLAGS=
        build   CGO_CXXFLAGS=
        build   CGO_LDFLAGS=
        build   GOARCH=arm64
        build   GOOS=darwin
        build   GOARM64=v8.0
```

For #1352.
Closes #1706.
Closes #963.

### DIFF
--- a/eng/doc/fips/README.md
+++ b/eng/doc/fips/README.md
@@ -437,7 +437,9 @@ This list of major changes is intended for quick reference and for access to his
 
 - Running `go version -m` on a binary which uses a system crypto backend now shows the `microsoft_systemcrypto=1` build setting.
 
- - It is now possible to build a binary that doesn't depend on a crypto package using invalid configurations, for example `GOOS=linux CGO_ENABLED=0 GOEXPERIMENT=systemcrypto`.
+ - The build-time backend compatibility check now only runs when a crypto package is required for the build.
+   - If your app doesn't depend on a crypto package, you may, for example, use `GOOS=linux CGO_ENABLED=0 GOEXPERIMENT=systemcrypto`.
+   - If your app doesn't use a crypto package and you make a change that introduces a crypto package dependency, you will only encounter a compatibility check failure after the change. The change may be in your transitive dependencies: for example, depending on a new module that uses `crypto/sha256` may trigger the compatibility check. This is undesirable, but it's necessary to enable flexibility.
 
 - `GOFIPS=0` support has been removed. It now has no effect.
 
--- a/eng/doc/fips/README.md
+++ b/eng/doc/fips/README.md
@@ -306,7 +306,7 @@ Most programs aren't expected to use these options. Determining FIPS mode at run
 
 ### Build option to use Go crypto if the backend compatibility check fails
 
-When building a Go program with a crypto backend, the build will check that the build environment and target are compatible with that backend. If not, the build will fail with an error. For example, a common unsupported build configuration is `GOOS=linux CGO_ENABLED=0 GOEXPERIMENT=opensslcrypto`. The OpenSSL backend requires cgo, so the build fails:
+When building a Go program that imports a `crypto` package with a crypto backend, the build will check that the build environment and target are compatible with that backend. If not, the build will fail with an error. For example, a common unsupported build configuration is `GOOS=linux CGO_ENABLED=0 GOEXPERIMENT=opensslcrypto`. The OpenSSL backend requires cgo, so the build fails:
 
 ```
 # runtime
--- a/eng/doc/fips/README.md
+++ b/eng/doc/fips/README.md
@@ -435,6 +435,10 @@ This list of major changes is intended for quick reference and for access to his
 
 ### Go 1.25 (Aug 2025)
 
+- Running `go version -m` on a binary which uses a system crypto backend now shows the `microsoft_systemcrypto=1` build setting.
+
+ - It is now possible to build a binary that doesn't depend on a crypto package using invalid configurations, for example `GOOS=linux CGO_ENABLED=0 GOEXPERIMENT=systemcrypto`.
+
 - `GOFIPS=0` support has been removed. It now has no effect.
 
 - `GOEXPERIMENT=allowcryptofallback` has been downgraded to the `allowcryptofallback` build tag. This is an internal mechanism that is not intended for use when building a Go application. This document has always recommended against using it, so we anticipate that this change won't affect users of the Microsoft build of Go. Please [contact the maintainers of the Microsoft build of Go](https://github.com/microsoft/go/blob/microsoft/main/SUPPORT.md) if you need to use it so we can understand the scenario and help find a safer alternative.
--- a/eng/doc/fips/README.md
+++ b/eng/doc/fips/README.md
@@ -317,11 +317,12 @@ When building a Go program that imports a `crypto` package with a crypto backend
         Please check your build environment and build command for a reason one or more of these tags weren't specified.
 ```
 
-We recommend one of these fixes:
+We recommend fixing the build environment to allow the crypto backend to be used. (Enable cgo.)
 
-- Fix the build environment to allow the crypto backend to be used. (Enable cgo.)
+These are other fixes that may be used on a case-by-case basis:
+
 - Remove `GOEXPERIMENT` entirely. This intentionally doesn't comply with the internal Microsoft crypto policy or FIPS, so for builds within Microsoft, this should only be done under a documented exception.
-
+- Refactor the code to not use a `crypto` package. For example, when computing a hash for not cryptographic purposes, there are several alternatives in the Go standard library that don't require a crypto backend, such as `hash/fnv` or `hash/maphash`.
 
 > [!IMPORTANT]
 > Individual crypto calls may fall back to standard Go crypto at runtime if the selected backend doesn't support an API or the arguments used. See the [FIPS User Guide](UserGuide.md) for more information.
--- a/eng/doc/fips/README.md
+++ b/eng/doc/fips/README.md
@@ -306,7 +306,7 @@ Most programs aren't expected to use these options. Determining FIPS mode at run
 
 ### Build option to use Go crypto if the backend compatibility check fails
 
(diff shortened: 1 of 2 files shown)


eng/doc/fips/README.md
```diff
@@ -435,6 +435,10 @@ This list of major changes is intended for quick reference and for access to his
 
 ### Go 1.25 (Aug 2025)
 
+- Running `go version -m` on a binary which uses a system crypto backend now shows the `microsoft_systemcrypto=1` build setting.
+
+ - It is now possible to build a binary that doesn't depend on a crypto package using invalid configurations, for example `GOOS=linux CGO_ENABLED=0 GOEXPERIMENT=systemcrypto`.
```
dagood: This seems confusing, I think a combination of ambiguity plus the configuration not actually being "invalid" in this case.
(E.g. `to build a binary that doesn't depend on (a crypto package using invalid configurations)` vs.
`to build a binary (that doesn't depend on a crypto package) using invalid configurations`,
also not totally clear what part of that the example is demonstrating.)

```suggestion
 - The build-time backend compatibility check now only runs when a crypto package is required for the build.
   - If your app doesn't depend on a crypto package, you may, for example, use `GOOS=linux CGO_ENABLED=0 GOEXPERIMENT=systemcrypto`.
   - If your app doesn't use a crypto package and you make a change that introduces a crypto package dependency, you will only encounter a compatibility check failure after the change. The change may be in your transitive dependencies: for example, depending on a new module that uses `crypto/sha256` may trigger the compatibility check. This is undesirable, but it's necessary to enable flexibility.
```

dagood: The doc should also note this behavior in the `Build option to use Go crypto if the backend compatibility check fails` section: acknowledge that when someone sees this error, it might not be because of their build setup, but rather because their source code might not have used `crypto` before. Another possible fix is to remove `crypto`.

dagood (changes requested):

dagood:
Also related:
* https://github.com/microsoft/go/issues/963
* https://github.com/microsoft/go/issues/972

eng/doc/fips/README.md
qmuntal: Good call. The fix you mention seems unrealistic, though, so I won't add it 😛.

qmuntal:
I'm open to renaming the setting to `ms_systemcrypto`. @dagood you are the expert here, what do you prefer?

eng/doc/fips/README.md
dagood: I think it could easily happen: someone adds a dependency that uses a sha256 checksum for some boring reason, and they can decide to remove the dependency and instead implement the feature without a checksum (or with a non-crypto checksum) if no-cgo is important to them.

I don't think it's *horrible* to let someone deduce this for themselves, but might be good for us to help them connect the dots a little...

dagood: I think really though the path that needs to be clear is: a user gets this error and can easily look it up in the FIPS docs and understand why it came up for them. Transitive dependencies specifically seem to me like they might not be obvious. ("crypto" doesn't show up in the go.mod after all.)

qmuntal: I get your point, but there are very few cases where not using crypto is the right fix, and I don't want to encourage people implementing their bespoke crypto algorithms(or third-party deps) that don't depend on crypto just to make the compiler happy.

I'll try to come up with something that we both can agree, though 😄 .

dagood:
Yeah, I briefly considered `microsoft_systemcrypto=1` vs. `ms_systemcrypto=1`, and I don't have any strong argument here. 😄

* ms:
  * `ms_` seems broadly nice for consistency.
  * However, this setting isn't tied to anything actually called `ms_systemcrypto`, so there's no *direct* consistency (like if the goexperiment were `ms_systemcrypto`). I also don't think anything else in this particular context will show up with an `ms_` prefix.
* microsoft;
  * If someone external picks up a Go binary and sees this string, `microsoft_systemcrypto` is more obvious than `ms_systemcrypto`.
  * But probably, they'd put it in a search engine either way and find us, so doesn't seem all that different.

Maybe our users will care what we call it because we're embedding it in their binaries?

That all said, I think my default choice would be `ms_` just because it matches other bits.

eng/doc/fips/README.md
qmuntal: Done!

qmuntal:
> If someone external picks up a Go binary and sees this string, microsoft_systemcrypto is more obvious than ms_systemcrypto.

I like this argument. Better be explicit that leave things open to interpretation.

eng/doc/fips/README.md
```diff
@@ -306,7 +306,7 @@ Most programs aren't expected to use these options. Determining FIPS mode at run
 
 ### Build option to use Go crypto if the backend compatibility check fails
 
-When building a Go program with a crypto backend, the build will check that the build environment and target are compatible with that backend. If not, the build will fail with an error. For example, a common unsupported build configuration is `GOOS=linux CGO_ENABLED=0 GOEXPERIMENT=opensslcrypto`. The OpenSSL backend requires cgo, so the build fails:
+When building a Go program that imports a `crypto` package with a crypto backend, the build will check that the build environment and target are compatible with that backend. If not, the build will fail with an error. For example, a common unsupported build configuration is `GOOS=linux CGO_ENABLED=0 GOEXPERIMENT=opensslcrypto`. The OpenSSL backend requires cgo, so the build fails:
```
dagood: I think this wording ends up with a similar ambiguity. Shifting the backend mention later works nicely IMO.

```suggestion
When building a Go program that imports a `crypto` package, the build will check that the build environment and target are compatible with the crypto backend being used, if any. If it's incompatible, the build will fail with an error. For example, a common unsupported build configuration is `GOOS=linux CGO_ENABLED=0 GOEXPERIMENT=opensslcrypto`. The OpenSSL backend requires cgo, so the build fails:
```

eng/doc/fips/README.md
```diff
@@ -317,11 +317,12 @@ When building a Go program with a crypto backend, the build will check that the
...
 
-We recommend one of these fixes:
+We recommend fixing the build environment to allow the crypto backend to be used. (Enable cgo.)
 
-- Fix the build environment to allow the crypto backend to be used. (Enable cgo.)
-- Remove `GOEXPERIMENT` entirely. This intentionally doesn't comply with the internal Microsoft crypto policy or FIPS, so for builds within Microsoft, this should only be done under a documented exception.
+These are other fixes that may be used on a case-by-case basis:
 
+- Remove `GOEXPERIMENT` entirely. This intentionally doesn't comply with the internal Microsoft crypto policy or FIPS, so for builds within Microsoft, this should only be done under a documented exception.
+- Refactor the code to not use a `crypto` package. For example, when computing a hash for not cryptographic purposes, there are several alternatives in the Go standard library that don't require a crypto backend, such as `hash/fnv` or `hash/maphash`.
```
```suggestion
- Refactor the code to not use a `crypto` package. For example, when computing a hash for non-cryptographic purposes, there are several alternatives in the Go standard library that don't require a crypto backend, such as `hash/fnv` or `hash/maphash`.
```

dagood (approved):
Thanks, I like the swap to one recommendation too.

eng/doc/fips/README.md
dagood: > I don't want to encourage people implementing their bespoke crypto algorithms(or third-party deps) that don't depend on crypto just to make the compiler happy.

That's a good point, hopefully someone on their project will understand the crypto policy and make sure this doesn't happen, but the way the policy and enforcement works will put people in a difficult situation and I'm sure *someone* will take our doc as an inspiration for a bad workaround.

In the past we've taken an openness approach in this doc (describe how things work, even internal workarounds, because motivated people will find it anyway and just waste their time), but that was a while ago. Perhaps also something to reevaluate on a broader scale.

Issue #966, opened 2023-06-27
Move backend-related errors into build logic to produce more specific and well-formatted errors


dagood:
We've added files that use unused strings to convey error messages like this:

```
# runtime
../go/go/src/runtime/backenderr_gen_nofallback_openssl.go:12:2: `
        The goexperiment.opensslcrypto tag is specified, but other tags required to enable that backend were not met.
        Required build tags:
          goexperiment.opensslcrypto && linux && cgo && !android && !cmd_go_bootstrap && !msan
        Please check your build environment and build command for a reason one or more of these tags weren't specified.
        For more information, visit https://github.com/microsoft/go/tree/microsoft/main/eng/doc/fips
        ` (untyped string constant "\n\tThe goexperiment.opensslcrypto tag is specified, but other tags ...) is not used
```

* https://github.com/microsoft/go/pull/950
* https://github.com/microsoft/go/pull/965

This is a clever way to make the build fail based on build tag combinations while conveying a fairly readable message.

If we move the logic to be implemented somewhere around `go/build/build.go`, we could improve the formatting and also make it more specific. (E.g. print out the current tag values, in the case of the above error.)

The "error file" unused string approach is simple and easy to maintain, so we should probably wait and see if it actually causes anyone trouble before making more complicated changes.

PR #1713, opened 2025-07-01
Remove `allowcryptofallback`

qmuntal:
After #1723 `allowcryptofallback` is only used to allow building the Go toolchain when `CGO_ENABLED=0`.

We can get this same behavior by removing `systemcrypto` from the `GOEXPERIMENT` environment variable or by using `GOEXPERIMENT=nosystemcrypto`, so `allowcryptofallback` is not really necessary.

For #1352.

### DIFF
--- a/eng/doc/fips/README.md
+++ b/eng/doc/fips/README.md
@@ -437,15 +437,15 @@ This list of major changes is intended for quick reference and for access to his
 
 - Running `go version -m` on a binary which uses a system crypto backend now shows the `microsoft_systemcrypto=1` build setting.
 
- - The build-time backend compatibility check now only runs when a crypto package is required for the build.
-   - If your app doesn't depend on a crypto package, you may, for example, use `GOOS=linux CGO_ENABLED=0 GOEXPERIMENT=systemcrypto`.
-   - If your app doesn't use a crypto package and you make a change that introduces a crypto package dependency, you will only encounter a compatibility check failure after the change. The change may be in your transitive dependencies: for example, depending on a new module that uses `crypto/sha256` may trigger the compatibility check. This is undesirable, but it's necessary to enable flexibility.
+- The build-time backend compatibility check now only runs when a crypto package is required for the build.
+  - If your app doesn't depend on a crypto package, you may, for example, use `GOOS=linux CGO_ENABLED=0 GOEXPERIMENT=systemcrypto`.
+  - If your app doesn't use a crypto package and you make a change that introduces a crypto package dependency, you will only encounter a compatibility check failure after the change. The change may be in your transitive dependencies: for example, depending on a new module that uses `crypto/sha256` may trigger the compatibility check. This is undesirable, but it's necessary to enable flexibility.
 
 - `GOFIPS=0` support has been removed. It now has no effect.
 
 - `GOEXPERIMENT=boringcrypto` support has been removed.
 
-- `GOEXPERIMENT=allowcryptofallback` has been downgraded to the `allowcryptofallback` build tag. This is an internal mechanism that is not intended for use when building a Go application. This document has always recommended against using it, so we anticipate that this change won't affect users of the Microsoft build of Go. Please [contact the maintainers of the Microsoft build of Go](https://github.com/microsoft/go/blob/microsoft/main/SUPPORT.md) if you need to use it so we can understand the scenario and help find a safer alternative.
+- `GOEXPERIMENT=allowcryptofallback` has removed. Instead, if it's necessary to opt out from using a system crypto backend, use `GOEXPERIMENT=nosystemcrypto`. This is an internal mechanism that is not intended for use when building a Go application. This document has always recommended against using it, so we anticipate that this change won't affect users of the Microsoft build of Go. Please [contact the maintainers of the Microsoft build of Go](https://github.com/microsoft/go/blob/microsoft/main/SUPPORT.md) if you need to use it so we can understand the scenario and help find a safer alternative.
 
 ### Go 1.24 (Feb 2025)
 
--- a/eng/doc/fips/README.md
+++ b/eng/doc/fips/README.md
@@ -443,9 +443,9 @@ This list of major changes is intended for quick reference and for access to his
 
 - `GOFIPS=0` support has been removed. It now has no effect.
 
-- `GOEXPERIMENT=boringcrypto` support has been removed.
+- `GOEXPERIMENT=boringcrypto` has been removed.
 
-- `GOEXPERIMENT=allowcryptofallback` has removed. Instead, if it's necessary to opt out from using a system crypto backend, use `GOEXPERIMENT=nosystemcrypto`. This is an internal mechanism that is not intended for use when building a Go application. This document has always recommended against using it, so we anticipate that this change won't affect users of the Microsoft build of Go. Please [contact the maintainers of the Microsoft build of Go](https://github.com/microsoft/go/blob/microsoft/main/SUPPORT.md) if you need to use it so we can understand the scenario and help find a safer alternative.
+- `GOEXPERIMENT=allowcryptofallback` has been removed. Instead, if it's necessary to opt out from using a system crypto backend, use `GOEXPERIMENT=nosystemcrypto`. This is an internal mechanism that is not intended for use when building a Go application. This document has always recommended against using it, so we anticipate that this change won't affect users of the Microsoft build of Go. Please [contact the maintainers of the Microsoft build of Go](https://github.com/microsoft/go/blob/microsoft/main/SUPPORT.md) if you need to use it so we can understand the scenario and help find a safer alternative.
 
 ### Go 1.24 (Feb 2025)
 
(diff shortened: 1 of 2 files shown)


qmuntal:
Still one test failing, I'll fix that tomorrow. It is ready for review anyway.

eng/doc/fips/README.md
```diff
@@ -445,7 +445,7 @@ This list of major changes is intended for quick reference and for access to his
 
 - `GOEXPERIMENT=boringcrypto` support has been removed.
 
-- `GOEXPERIMENT=allowcryptofallback` has been downgraded to the `allowcryptofallback` build tag. This is an internal mechanism that is not intended for use when building a Go application. This document has always recommended against using it, so we anticipate that this change won't affect users of the Microsoft build of Go. Please [contact the maintainers of the Microsoft build of Go](https://github.com/microsoft/go/blob/microsoft/main/SUPPORT.md) if you need to use it so we can understand the scenario and help find a safer alternative.
+- `GOEXPERIMENT=allowcryptofallback` has been removed. Use `GOEXPERIMENT=nosystemcrypto` to explicitly opt-out from using a system crypto backend.
```
dagood: Worth explicitly preventing someone from possibly interpretating this as a command out of context?

```suggestion
- `GOEXPERIMENT=allowcryptofallback` has been removed. Instead, if it's necessary to opt out from using a system crypto backend, use `GOEXPERIMENT=nosystemcrypto`.
```

dagood: I'm not sure I agree with removing the bulk of this text. It seems like all that needs to change is the first sentence. Why the rest?

eng/pipeline/stages/go-builder-matrix-stages.yml
```diff
@@ -82,25 +82,25 @@ stages:
         - ${{ if parameters.innerloop }}:
           - { os: darwin, arch: amd64, config: devscript }
           - { os: darwin, arch: amd64, config: test }
-          - { experiment: darwincrypto, os: darwin, arch: amd64, config: test }
-          - { experiment: darwincrypto, os: darwin, arch: amd64, config: test, fips: true }
+          - { experiment: systemcrypto, os: darwin, arch: amd64, config: test }
+          - { experiment: systemcrypto, os: darwin, arch: amd64, config: test, fips: true }
```
dagood: Why these changes? Is it just tidy or was there some driving reason?

dagood (changes requested):
This seems to be doing a decent amount more than removing the tag, and it's not clear to me what the scope is from the title/desc. I didn't fully review but a little confused by the first few parts I read through.

dagood:
> This PR uses another approach: we will check if there is any conflicting experiment, e.g. `GOEXPERIMENT=cngcrypto,opensslcrypto`, or an unsupported platform, e.g. `GOEXPERIMENT=cngcrypto GOOS=linux`, when parsing the GOEXPERIMENT environment variable. The few remaining checks are kept in files that make compilation to fail.

* Nice, tracked: https://github.com/microsoft/go/issues/966

qmuntal:
> This seems to be doing a decent amount more than removing the tag, and it's not clear to me what the scope is from the title/desc. I didn't fully review but a little confused by the first few parts I read through.

Good point. I'll factor out the goexperiment validation to another PR.

Done:
- https://github.com/microsoft/go/pull/1719
- https://github.com/microsoft/go/pull/1723

eng/pipeline/stages/go-builder-matrix-stages.yml
qmuntal: I'll move this to another PR, but basically I do it because `systemcrypto` makes more test to pass. For example on Linux you can cross compile to Windows if using `systemcrypto`, but not `opensslcrypto`, and some tests do that. These tests were previously passing because of `allowcryptofallback`, but they won't anymore once we remove it.

qmuntal: https://github.com/microsoft/go/pull/1719

eng/doc/fips/README.md
qmuntal: Will re-add the deleted text 👍

qmuntal:
Ready for review!

eng/doc/fips/README.md
```diff
@@ -437,15 +437,15 @@ This list of major changes is intended for quick reference and for access to his
...
 
-- `GOEXPERIMENT=allowcryptofallback` has been downgraded to the `allowcryptofallback` build tag. This is an internal mechanism that is not intended for use when building a Go application. This document has always recommended against using it, so we anticipate that this change won't affect users of the Microsoft build of Go. Please [contact the maintainers of the Microsoft build of Go](https://github.com/microsoft/go/blob/microsoft/main/SUPPORT.md) if you need to use it so we can understand the scenario and help find a safer alternative.
+- `GOEXPERIMENT=allowcryptofallback` has removed. Instead, if it's necessary to opt out from using a system crypto backend, use `GOEXPERIMENT=nosystemcrypto`. This is an internal mechanism that is not intended for use when building a Go application. This document has always recommended against using it, so we anticipate that this change won't affect users of the Microsoft build of Go. Please [contact the maintainers of the Microsoft build of Go](https://github.com/microsoft/go/blob/microsoft/main/SUPPORT.md) if you need to use it so we can understand the scenario and help find a safer alternative.
```
```suggestion
- `GOEXPERIMENT=allowcryptofallback` has been removed. Instead, if it's necessary to opt out from using a system crypto backend, use `GOEXPERIMENT=nosystemcrypto`. This is an internal mechanism that is not intended for use when building a Go application. This document has always recommended against using it, so we anticipate that this change won't affect users of the Microsoft build of Go. Please [contact the maintainers of the Microsoft build of Go](https://github.com/microsoft/go/blob/microsoft/main/SUPPORT.md) if you need to use it so we can understand the scenario and help find a safer alternative.
```

eng/doc/fips/README.md
```diff
@@ -437,15 +437,15 @@ This list of major changes is intended for quick reference and for access to his
...
-   - If your app doesn't use a crypto package and you make a change that introduces a crypto package dependency, you will only encounter a compatibility check failure after the change. The change may be in your transitive dependencies: for example, depending on a new module that uses `crypto/sha256` may trigger the compatibility check. This is undesirable, but it's necessary to enable flexibility.
+- The build-time backend compatibility check now only runs when a crypto package is required for the build.
+  - If your app doesn't depend on a crypto package, you may, for example, use `GOOS=linux CGO_ENABLED=0 GOEXPERIMENT=systemcrypto`.
+  - If your app doesn't use a crypto package and you make a change that introduces a crypto package dependency, you will only encounter a compatibility check failure after the change. The change may be in your transitive dependencies: for example, depending on a new module that uses `crypto/sha256` may trigger the compatibility check. This is undesirable, but it's necessary to enable flexibility.
 
 - `GOFIPS=0` support has been removed. It now has no effect.
 
 - `GOEXPERIMENT=boringcrypto` support has been removed.
```
dagood: It never had "support":

> boringcrypto selects the upstream BoringCrypto backend, which is not supported nor compliant

Although unfortunately "support" has multiple meanings... I think adjusting this to fit with the next point makes sense:

```suggestion
- `GOEXPERIMENT=boringcrypto` has been removed.
```

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -130,30 +131,43 @@ index e17a5402af818d..e83c4c5b79d11b 100644
...
 --- a/src/cmd/dist/build.go
 +++ b/src/cmd/dist/build.go
-@@ -1548,6 +1548,11 @@ func cmdbootstrap() {
+@@ -1387,6 +1387,12 @@ func toolenv() []string {
+ 		// still have full paths for stack traces for compiler crashes and the like.
+ 		env = append(env, "GOFLAGS=-trimpath -ldflags=-w -gcflags=cmd/...=-dwarf=false")
+ 	}
++	if goexp := os.Getenv("GOEXPERIMENT"); goexp != "" {
++		// Remove systemcrypto from GOEXPERIMENT, the Microsoft build of Go
++		// should not use it to favor portability. It only uses crypto packages
```
```suggestion
+		// Remove systemcrypto from GOEXPERIMENT: to favor portability, the
+		// Microsoft build of Go should not use it. Go only uses crypto packages
```

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -451,6 +507,37 @@ index 483a9ec33c6798..fad922e3f844ab 100644
...
+ 	"strings"
+ )
+@@ -187,7 +188,17 @@ func Main(arch *sys.Arch, theArch Arch) {
+ 
+ 	buildVersion := buildcfg.Version
+ 	if goexperiment := buildcfg.Experiment.String(); goexperiment != "" {
+-		buildVersion += " X:" + goexperiment
++		// Don't report crypto experiments in build version.
++		// These are already reported in the buildInfo.
++		// If we don't do this, "go tool dist test" will always fail.
```
dagood: Possible to expand a little on why/where the test fails? (And why it's not "good" that it fails?)

dagood (approved):
Looks good, just comment/doc suggestions.

gdams (approved):
LGTM (with Davis' comments resolved)

patches/0003-Implement-crypto-internal-backend.patch
qmuntal: Hmm, difficult to explain 😓. I'll try another approach that is more laser focused on the failing tests issues (see next commit). Thanks for the feedback!

dagood (approved):

Issue #1474, opened 2025-01-09
Check that vendoring patch is correct in CI


dagood:
It's easy for a mistake in the vendoring patch to slip through. There are a few types of test we could add:

* Make sure the patch only modifies files under `src/vendor/`.
* Make sure the patch matches the result of a freshly performed `go mod vendor` exactly. (Not just the repo state, the patch itself--no extraneous adds.)

A tricky part of an approach that actually applies patches is identifying the commit in the repo associated with the vendoring patch. Matching against the commit title could work, but might be fragile. Something more stable could be to use a `patch number 50` command to get a fixed number, then CI uses that number to selectively apply pre-vendor patch files and also the vendor patch individually.

dagood:
We could also simply apply all patches, build `go`, then `go mod vendor`, and `git diff --exit-code`. We could incorporate this into the `.github/workflows/patch-build.yml` workflow, so we know the vendoring dir is correct after each patch, not just after all are applied.

Checking that vendoring changes are exclusive to the vendoring patch is a nice-to-have. Overall vendor dir validity is the most important.

gdams:
@dagood my recent changes to CI check that go mod vendor doesn't produce a diff but we don't check that the commit in the go.mod file exists in the backend repo. Do we still need to do this? Shall I spin it off into a separate issue?

dagood:
I don't see where that comes from in this issue, but I do think it'd be worth checking. I kind of expected that `go mod vendor` would inherently check that when it tries to resolve modules, but I guess I'm not entirely sure what happens if go.sum and vendor are all valid.

dagood:
From https://github.com/microsoft/go/pull/1717#issuecomment-3028610638:

> Will there be a followup to move it to the "each patch" workflow? Or is that not a useful part of this?
> 
> > We could incorporate this into the `.github/workflows/patch-build.yml` workflow, so we know the vendoring dir is correct after each patch, not just after all are applied.

PR #1717, opened 2025-07-02
ci: add workflow to check if vendor patch is clean

gdams:
Adds a workflow to ensure that we don't check in unclean changes to the vendor patch


qmuntal (approved):

dagood:
* For https://github.com/microsoft/go/issues/1474

Will there be a followup to move it to the "each patch" workflow? Or is that not a useful part of this?

> We could incorporate this into the `.github/workflows/patch-build.yml` workflow, so we know the vendoring dir is correct after each patch, not just after all are applied.

PR #1721, opened 2025-07-03
Support for excluding OS-disabled cipher suites in TLS handshake

qmuntal:
This PR does the following:

- Adds `GOEXPERIMENT=ms_tls_config_schannel`
- When the new experiment is enabled, all TLS cipher suites are filtered and reordered based on the Windows' Schannel configuration (as retrieved by [BCryptEnumContextFunctions](https://learn.microsoft.com/en-us/windows/win32/api/bcrypt/nf-bcrypt-bcryptenumcontextfunctions))
- For TLS 1.2, [tls.Config.CipherSuites](https://pkg.go.dev/crypto/tls#Config) can still be used to further limit the enabled cipher suites.

Note that Schannel might support more cipher suites than Go. These cipher suites won't be used.

This new feature is orthogonal to `GOEXPERIMENT=systemcrypto`, so it can be enabled independenlty.

For https://github.com/microsoft/go-lab/issues/215.

### DIFF
--- a/patches/0011-implement-ms_tls_config_schannel-experiment.patch
+++ b/patches/0011-implement-ms_tls_config_schannel-experiment.patch
@@ -1,27 +1,27 @@
 From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
 From: qmuntal <qmuntaldiaz@microsoft.com>
 Date: Thu, 3 Jul 2025 15:13:26 +0200
-Subject: [PATCH] implement ms_schannelsiphersuites experiment
+Subject: [PATCH] implement ms_tls_config_schannel experiment
 
 ---
  src/crypto/tls/cipher_suites_windows.go       | 74 +++++++++++++++++++
  src/crypto/tls/cipher_suites_windows_test.go  | 67 +++++++++++++++++
- src/crypto/tls/fips140_test.go                |  4 +
+ src/crypto/tls/fips140_test.go                |  7 ++
  src/crypto/tls/tls_test.go                    |  8 ++
- .../exp_ms_schannelciphersuites_off.go        |  8 ++
- .../exp_ms_schannelciphersuites_on.go         |  8 ++
+ .../exp_ms_tls_config_schannel_off.go         |  8 ++
+ .../exp_ms_tls_config_schannel_on.go          |  8 ++
  src/internal/goexperiment/flags.go            |  4 +
  .../syscall/windows/security_windows.go       | 28 +++++++
  .../syscall/windows/zsyscall_windows.go       | 16 ++++
- 9 files changed, 217 insertions(+)
+ 9 files changed, 220 insertions(+)
  create mode 100644 src/crypto/tls/cipher_suites_windows.go
  create mode 100644 src/crypto/tls/cipher_suites_windows_test.go
- create mode 100644 src/internal/goexperiment/exp_ms_schannelciphersuites_off.go
- create mode 100644 src/internal/goexperiment/exp_ms_schannelciphersuites_on.go
+ create mode 100644 src/internal/goexperiment/exp_ms_tls_config_schannel_off.go
+ create mode 100644 src/internal/goexperiment/exp_ms_tls_config_schannel_on.go
 
 diff --git a/src/crypto/tls/cipher_suites_windows.go b/src/crypto/tls/cipher_suites_windows.go
 new file mode 100644
-index 00000000000000..d9276b46fbadc9
+index 00000000000000..c1dcd97555fe4d
 --- /dev/null
 +++ b/src/crypto/tls/cipher_suites_windows.go
 @@ -0,0 +1,74 @@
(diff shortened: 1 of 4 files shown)


patches/0011-implement-ms_tls_config_schannel-experiment.patch
```diff
@@ -0,0 +1,357 @@
...
++	"internal/goexperiment"
+ 	"internal/obscuretestdata"
+ 	"internal/testenv"
+ 	"math/big"
+@@ -51,6 +52,9 @@ func generateKeyShare(group CurveID) keyShare {
+ }
+ 
+ func TestFIPSServerProtocolVersion(t *testing.T) {
++	if goexperiment.MS_TLS_Config_Schannel {
++		t.Skip("skipping Schannel cipher suite test")
```
dagood: But why? I think:

```suggestion
+		// Additional test patching to support the experiment wouldn't add valuable coverage.
+		t.Skip("skipping: doesn't account for Schannel cipher suites")
```

patches/0011-implement-ms_tls_config_schannel-experiment.patch
```diff
@@ -0,0 +1,357 @@
...
++	}
+ 	test := func(t *testing.T, name string, v uint16, msg string) {
+ 		t.Run(name, func(t *testing.T) {
+ 			serverConfig := testConfig.Clone()
+@@ -174,6 +178,9 @@ func isFIPSSignatureScheme(alg SignatureScheme) bool {
+ }
+ 
+ func TestFIPSServerCipherSuites(t *testing.T) {
++	if goexperiment.MS_TLS_Config_Schannel {
++		t.Skip("skipping Schannel cipher suite test")
```
dagood:
```suggestion
+		// Additional test patching to support the experiment wouldn't add valuable coverage.
+		t.Skip("skipping: doesn't account for Schannel cipher suites")
```

patches/0011-implement-ms_tls_config_schannel-experiment.patch
```diff
@@ -0,0 +1,357 @@
...
++	}
+ }
+ 
+ func TestDialTimeout(t *testing.T) {
+@@ -1545,6 +1550,9 @@ func TestClientHelloInfo_SupportsCertificate(t *testing.T) {
+ }
+ 
+ func TestCipherSuites(t *testing.T) {
++	if goexperiment.MS_TLS_Config_Schannel {
++		t.Skip("Skipping CipherSuites test in Schannel mode")
```
dagood:
```suggestion
+		// Additional test patching to support the experiment wouldn't add valuable coverage.
+		t.Skip("skipping: doesn't account for Schannel cipher suites")
```

patches/0011-implement-ms_tls_config_schannel-experiment.patch
```diff
@@ -0,0 +1,357 @@
...
++	}
++	for _, c := range InsecureCipherSuites() {
++		if c.Name == name {
++			return c.ID, true
++		}
++	}
++	return 0, false
++}
++
++// cipherSuitesSchannel returns all the cipher suites that Schannel supports in preference order.
```
dagood:
```suggestion
+// cipherSuitesSchannel returns all the cipher suites that Schannel supports in Schannel's preference order.
```

(The comment on the expeirment makes it clear that it should be this way, so just bringing that info a little closer to the implementation.)

patches/0011-implement-ms_tls_config_schannel-experiment.patch
```diff
@@ -0,0 +1,357 @@
...
++		id, ok := cipherSuiteID(name)
++		if !ok {
++			continue // cipher suite not found in the provided list
++		}
++		suites = append(suites, id)
++	}
++	return suites, nil
++}
++
++// filterCipherSuites filters the provided cipher suites, removing those that are not in the allowed list.
```
dagood:
```suggestion
+// filterCipherSuites filters the provided cipher suites, creating a new slice that only includes those that are in the allowed list.
```

With `slices.DeleteFunc` etc., "removing" tends to mean modifying the original.

(I originally added a review comment to potentially use `slices.DeleteFunc`, but now that I've taken a closer look--that would turn out badly.)

patches/0011-implement-ms_tls_config_schannel-experiment.patch
```diff
@@ -0,0 +1,357 @@
...
+ 	defer runtime.KeepAlive(sid)
+ 	return *(*uint8)(unsafe.Pointer(getSidSubAuthorityCount(sid)))
+ }
++
++const (
++	CRYPT_LOCAL = 0x00000001
++
++	CRYPT_UM = 0x00000001
++
++	NCRYPT_SCHANNEL_INTERFACE = 0x00010002
```
dagood, line 392: Odd spacing (unless there's a URL reference for the values to add in)?

dagood (approved):
Nice! Just comment suggestions.

patches/0011-implement-ms_tls_config_schannel-experiment.patch
qmuntal: I like to put related constants (in this case all belong to BCrypt) in the same block, but keeping them separated if they refer to different concepts (that is, they are used in different functions or parameters). This idiom is also followed in upstream Go (sometimes 😄)

gdams (approved):

PR #1723, opened 2025-07-04
Validate backend constraints while parsing goexperiments

### LINKED ISSUE #966: Move backend-related errors into build logic to produce more specific and well-formatted errors

qmuntal:
We previously used a set of autogenerated files that would abort compilation if there was an invalid condition AND `allowcryptofallback` was not set.

This PR uses another approach: we will check if there is any conflicting experiment, e.g. `GOEXPERIMENT=cngcrypto,opensslcrypto`, or an unsupported platform, e.g. `GOEXPERIMENT=cngcrypto GOOS=linux`, when parsing the GOEXPERIMENT environment variable. The few remaining checks are kept in files that make compilation to fail.

This means that we get part of the goodness we lost in #1704, as these checks are useful regardless of whether `crypto` is used or not, while still only running the cgo-related checks when `crypto` is imported. On the other hand, we no longer can check tags (e.g. `-tags=goexperiment.cngcrypto GOOS=linux`), but that will still not compile, it's just that the error will be more cryptic, and that's an advanced settings anyway.

There are some tests that need to be skipped because `allowcryptofallback` hasn't been ported to the new way of doing checks (we want to get rid of it, but it still applies to the cgo checks, will fix that in a follow up PR).

CI now always uses `systemcrypto` instead of the backend-specific goexperiments to make some tests happy (that were previously passing due to `allowcryptofallback`) instead of skipping them one by one (we also want to get rid of backend-specific goexperiments).

Closes #966.

### DIFF
--- a/patches/0003-Implement-crypto-internal-backend.patch
+++ b/patches/0003-Implement-crypto-internal-backend.patch
@@ -2358,7 +2358,7 @@ index 00000000000000..5e4b436554d44d
 +// (because the implementation might be here).
 diff --git a/src/crypto/systemcrypto_nocgo.go b/src/crypto/systemcrypto_nocgo.go
 new file mode 100644
-index 00000000000000..13782a8241b99a
+index 00000000000000..99cf22dc13ef89
 --- /dev/null
 +++ b/src/crypto/systemcrypto_nocgo.go
 @@ -0,0 +1,15 @@
@@ -2366,7 +2366,7 @@ index 00000000000000..13782a8241b99a
 +// Use of this source code is governed by a BSD-style
 +// license that can be found in the LICENSE file.
 +
-+//go:build !cgo && (goexperiment.opensslcrypto || goexperiment.darwincrypto)
++//go:build !cgo && (goexperiment.opensslcrypto || goexperiment.darwincrypto) && !allowcryptofallback
 +
 +package crypto
 +
--- a/patches/0003-Implement-crypto-internal-backend.patch
+++ b/patches/0003-Implement-crypto-internal-backend.patch
@@ -19,6 +19,7 @@ desired goexperiments and build tags.
  src/cmd/go/systemcrypto_test.go               | 144 +++++++
  .../go/testdata/script/env_cross_build.txt    |   2 +
  .../script/test_android_issue62123.txt        |   2 +
+ src/cmd/internal/testdir/testdir_test.go      |   7 +
  src/cmd/link/link_test.go                     |   8 +
  src/crypto/internal/backend/backend_test.go   |  30 ++
  src/crypto/internal/backend/bbig/big.go       |  17 +
@@ -50,7 +51,7 @@ desired goexperiments and build tags.
  src/internal/buildcfg/exp.go                  |  54 +++
  src/runtime/runtime_boring.go                 |   5 +
  src/syscall/exec_linux_test.go                |   2 +-
- 41 files changed, 2162 insertions(+), 12 deletions(-)
+ 42 files changed, 2169 insertions(+), 12 deletions(-)
  create mode 100644 src/cmd/go/systemcrypto_test.go
  create mode 100644 src/crypto/internal/backend/backend_test.go
  create mode 100644 src/crypto/internal/backend/bbig/big.go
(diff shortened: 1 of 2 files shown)


gdams (approved):

qmuntal:
@dagood I'll merge this PR to unblock some other PRs in the queue. Feel free to put your comments in here, I'll take them into account latter.

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -338,8 +394,8 @@ index 00000000000000..b2298b8157ae92
 +			out := outPath(t)
 +
 +			// Build the binary with the specified GOEXPERIMENT.
-+			fail, _ := execGoTool(t, true, []string{"GOEXPERIMENT=" + tt.exp}, "build", "-o", out, file)
-+			if fail && tt.enabled {
++			succed, _ := execGoTool(t, true, []string{"GOEXPERIMENT=" + tt.exp}, "build", "-o", out, file)
++			if !succed {
```
dagood, line 398: Succeed* (or success, succeeded?)

Or: `ok`, perhaps flip the order.

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -1843,6 +1653,27 @@ index 00000000000000..a8ff666b6d182e
...
++
++//go:build requirefips && !goexperiment.systemcrypto
++
++package fips140
++
++func init() {
++	`
++	The requirefips tag is enabled, but no crypto backend is enabled.
++	A crypto backend is required to enable FIPS mode.
++	For more information, visit https://github.com/microsoft/go/tree/microsoft/main/eng/doc/fips
```
dagood, line 1674: I wonder if we need to adjust https://github.com/microsoft/go/tree/microsoft/main/eng/doc/fips#build-option-to-require-fips-mode to make it clearer that programs that contain no crypto code will not examine FIPS mode. I don't think this is necessarily related to any of our recent changes, but given the current docs, I think someone could make a simple "hello world" app with requirefips and expect it to function as a FIPS check canary.

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -2951,6 +2596,83 @@ index 9b7613a62b5a31..eddaf170772e19 100644
...
++	type platform struct {
++		goos   string
++		goarch []string // if nil or empty, enabled on all architectures
++	}
++	var backends = []struct {
++		name     string
++		enabled  bool
++		platform []platform
++	}{
++		{"systemcrypto", flags.SystemCrypto, []platform{
```
dagood, line 2632: It seems worth mentioning that the conflict logic requires `systemcrypto` to appear first in the slice--I found it a bit unexpected and had to figure out later how an order dependency lets the `backend2` loop omit the `backend.name != "systemcrypto"` check.

patches/0003-Implement-crypto-internal-backend.patch
qmuntal: "ok + flip the order" 👍

patches/0003-Implement-crypto-internal-backend.patch
qmuntal: will add an explicit check.

patches/0003-Implement-crypto-internal-backend.patch
qmuntal: > but given the current docs, I think someone could make a simple "hello world" app with requirefips and expect it to function as a FIPS check canary.

Good point, will add a line in the docs about this.

PR #1727, opened 2025-07-08
[release-branch.go1.23] Support for excluding OS-disabled cipher suites in TLS handshake

qmuntal:
This is a port of #1721.

I've added some improvements that I'll forward port to main. The main change is that TLS 1.2 cipher suites manually configured in Config.CipherSuite are now honored, but still using the preference order taken from Schannel.

### DIFF
--- a/patches/0016-implement-ms_tls_config_schannel-experiment.patch
+++ b/patches/0016-implement-ms_tls_config_schannel-experiment.patch
@@ -1,0 +1,550 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: qmuntal <qmuntaldiaz@microsoft.com>
+Date: Tue, 8 Jul 2025 16:04:25 +0200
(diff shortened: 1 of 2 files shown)


gdams (approved):

gdams (approved):

patches/0016-implement-ms_tls_config_schannel-experiment.patch
```diff
@@ -0,0 +1,568 @@
...
++	}
++	for _, id := range suites {
++		if !slices.Contains(out, id) {
++			out = append(out, id)
++		}
++	}
++	return out
++}
++
++// disableCipherSuites modifies the disabled map adding the cippher suites in supported
```
dagood, line 259:
```suggestion
+// disableCipherSuites modifies the disabled map, adding the cipher suites in supported
```

dagood (approved):
LGTM. (Not worth fixing the comment before merging IMO, to avoid respins.)

PR #1744, opened 2025-07-15
Upgrade go-infra/telemetry

qmuntal:
The newest version of the telemetry client enables uploading telemetry for unknown Go versions, which is necessary for when we release Go 1.25.

It also decouples the telemetry configuration from the client.

### DIFF
--- a/patches/0010-add-appinsights-telemetry.patch
+++ b/patches/0010-add-appinsights-telemetry.patch
@@ -5,14 +5,25 @@ Subject: [PATCH] Add appinsights telemetry
 
 ---
  README.md                                     | 15 ++++++
+ src/cmd/go.mod                                |  3 +-
+ src/cmd/go.sum                                |  6 ++-
  src/cmd/go/main.go                            |  1 +
  src/cmd/go/mstelemetry_test.go                | 52 +++++++++++++++++++
  src/cmd/go/script_test.go                     |  7 +++
- src/cmd/internal/telemetry/counter/counter.go | 31 ++++++++++-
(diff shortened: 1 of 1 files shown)


gdams (approved):

patches/0010-add-appinsights-telemetry.patch
```diff
@@ -38,6 +49,37 @@ index 71c9d1dc299388..8cab5daec4da54 100644
...
+diff --git a/src/cmd/go.mod b/src/cmd/go.mod
+index 1d4b7f30894cbd..edaa5d3cb8b2c5 100644
+--- a/src/cmd/go.mod
++++ b/src/cmd/go.mod
+@@ -4,7 +4,8 @@ go 1.25
+ 
+ require (
+ 	github.com/google/pprof v0.0.0-20250208200701-d0013a598941
+-	github.com/microsoft/go-infra/telemetry v0.0.0-20250603171931-8d11a613d784
++	github.com/microsoft/go-infra/telemetry v0.0.0-20250715111012-2d08bb42a36b
```
dagood, line 61: Looks like the upgrade ended up in the "wrong" patch file.

dagood: Pulled out the patch matching feature from the patch validation item into its own issue for this:
* https://github.com/microsoft/go/issues/1748

qmuntal: Fixed in https://github.com/microsoft/go/pull/1749.

PR #1746, opened 2025-07-15
Do not skip allocation tests and remove extra allocations from hash structs

mertakman:
Previously we were skipping allocation tests, but now with fixed amount of allocations on expected results, rather than passing we running them with correct numbers. Also removed extra allocations from Darwin backend due to returning pointer of hash structs..


qmuntal (changes requested):
The tests should still check for zero allocations when `goexperiment.SystemCrypto` is false.

qmuntal (approved):

gdams (approved):

Issue #1598, opened 2025-03-19
Azure Pipelines CI test logs are excessively long


dagood:
It took a good 20 seconds to load(/render?) the logs in this build:

https://dev.azure.com/dnceng-public/public/_build/results?buildId=985995&view=logs&jobId=c420baff-3a59-5926-6621-c8ddad8a0f0d&j=14ce9276-1b71-5587-760e-edce85b6645f&t=0d36b789-fcd3-5422-a8f5-dd823a4fa99d

The JSON logs appear to be making it to the AzDO display layer, which results in a huge amount of output: 507,000 lines. Maybe this would be useful to save as a file, but it makes it difficult to poke through a build.

(Also, once the logs load, clicking other parts of the AzDO UI sometimes cause my browser to briefly become unresponsive.)

qmuntal:
We pass all logs verbatim to `os.Stdout` in here: https://github.com/microsoft/go/blob/aee5af64845f9f6ce38e55d468fc23db31acdec4/eng/_util/cmd/build/build.go#L218.

We could filter them a bit to remove noise. In fact I initially did so, but end up reverting (see #1566) because I missed some important lines and didn't want to spend more time refining the filter. It probably deserves another shot.

dagood:
I think keeping the original around as an artifact is the way to go, because it means it's not critical for the filtering to be perfect. Could try https://learn.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands?view=azure-devops&tabs=bash#upload-upload-an-artifact to avoid AzDO task clutter.

PR #1751, opened 2025-07-17
Set definitive instrumentation key and endpoint

qmuntal:
We already have the definitive App Insights instrumentation key and endpoint. Let's use it.

This means that when we release Go 1.25, the toolchain will upload data by default. For development versions of the toolchain, `MS_GOTOOLCHAIN_TELEMETRY_ALLOW_GO_DEVEL=1` is still required.

While here, change `MS_GOTOOLCHAIN_TELEMETRY_STATS` to be enabled if it is set to `1` instead of `on` for consitency.

For #1672.

### DIFF
--- a/patches/0010-add-appinsights-telemetry.patch
+++ b/patches/0010-add-appinsights-telemetry.patch
@@ -8,8 +8,8 @@ Subject: [PATCH] Add appinsights telemetry
  .../internal/telemetrystats/telemetrystats.go | 13 +++++
  src/cmd/go/main.go                            |  1 +
  src/cmd/go/mstelemetry_test.go                | 52 +++++++++++++++++++
- src/cmd/go/script_test.go                     |  7 +++
- src/cmd/internal/telemetry/counter/counter.go | 36 ++++++++++++-
+ src/cmd/go/script_test.go                     |  6 +++
+ src/cmd/internal/telemetry/counter/counter.go | 37 ++++++++++++-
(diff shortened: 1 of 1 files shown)


gdams (approved):

dagood:
Tests failed, the result upload didn't work, seemingly because of `##[warning]Failed to read /Users/runner/work/1/s/eng/artifacts/TestResults.xml. Error : '', hexadecimal value 0x01, is an invalid character. Line 139113, position 21..`

Spotted these lines with the char:
```
2025-07-17T08:27:54.9850060Z {"Time":"2025-07-17T08:26:53.994198Z","Action":"output","Package":"internal/fuzz","Test":"FuzzRuneRoundTrip/seed#3","Output":"        rune('�')\n"}
2025-07-17T08:27:55.0303760Z {"Time":"2025-07-17T08:26:53.996313Z","Action":"output","Package":"internal/fuzz","Test":"FuzzStringRoundTrip/seed#2","Output":"    encoding_test.go:392: marshaled \"�\":\n"}
2025-07-17T08:27:55.0305100Z {"Time":"2025-07-17T08:26:53.996318Z","Action":"output","Package":"internal/fuzz","Test":"FuzzStringRoundTrip/seed#2","Output":"        string(\"�\")\n"}
```

* Would make this case easier to dig through: https://github.com/microsoft/go/issues/1598

PR #1753, opened 2025-07-17
backport: add telemetry support to go1.24

gdams:


dagood (changes requested):
I know it's a draft, but without information about this I want to make sure it gets planning attention because we announced this for 1.25.

PR #1757, opened 2025-07-21
[release-branch.go1.23] Remove curve prefix from Schannel cipher suite names

qmuntal:
The ARC team reported that GOEXPERIMENT=ms_tls_config_schannel is not working on Windows Server 2012. The issue is that cipher suite names are not using the standard form (which is also the one understood by Go), but they have the ECC curve name added as a prefix.

Windows Server 2012 is not supported by Go, although ARC still supports it. Anyway, newer Windows do still support these prefixes for the case a newer Windows version got upgraded from Windows Server 2012.

### DIFF
--- a/patches/0016-implement-ms_tls_config_schannel-experiment.patch
+++ b/patches/0016-implement-ms_tls_config_schannel-experiment.patch
@@ -8,16 +8,16 @@ Subject: [PATCH] implement ms_tls_config_schannel experiment
  src/crypto/tls/bogo_shim_test.go              |   7 +
  src/crypto/tls/handshake_server_test.go       |  20 ++
  src/crypto/tls/handshake_test.go              |   6 +
- src/crypto/tls/schannel_windows.go            | 142 +++++++++++++
- src/crypto/tls/schannel_windows_test.go       | 190 ++++++++++++++++++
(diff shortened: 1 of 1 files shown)


gdams (approved):

patches/0016-implement-ms_tls_config_schannel-experiment.patch
```diff
@@ -238,6 +242,23 @@ index 00000000000000..a77186c66d3095
...
++// This is not common, as since Windows 2016 the curve prefix is discarded by Schannel,
++// but it can still be present in case Windows was upgraded from an earlier version.
++func removeCipherSuiteCurvePrefix(name string) string {
++	const prefixLength = 5 // Length of "_P256", "_P384", or "_P521"
++	if len(name) < prefixLength {
++		return name // No curve prefix to remove
++	}
++	switch name[len(name)-prefixLength:] {
++	case "_P256", "_P384", "_P521":
++		name = name[:len(name)-prefixLength] // Remove the curve prefix
```
dagood, line 257: That's not a prefix, it's a suffix. ("Affix" being the general term.)

qmuntal: d'oh, this is one of my achilles heels, I always swap them 😅

Issue #402, opened 2022-02-02
Merge `_core` and `_util` into one module, write no-dependency enforcement test


dagood:
Right now, `_core` has no external dependencies, and `_util` has a few. The idea is that by isolating these modules from each other, we can easily spot external dependencies being mistakenly added to `_core` in code review and stop the PR from being merged. By having parts of CI use `_core` only, we are sure those parts don't take on external dependencies.

This makes it hard to share code between `_core` and `_util`.

It also makes it hard to use our go-infra repository as a dependency from `_core`. If we vendor the dependency, it would be fine, *but* it makes catching problems in manual review riskier: someone would need to spot 1 -> 2 dependencies rather than 0 -> 1 dependencies.

I think we should write tests that verify the set of dependencies being used to build specific tools. Then we can merge `_core` and `_util` and not rely as much on manual dependency change reviews.

(This is inspired by seeing that Go itself does some similar tests. 😄)

PR #1762, opened 2025-07-21
ci: enable linter on files in eng/_util

### LINKED ISSUE #78: Run a linter on Go files in `eng/` as a PR test
I've started running `golint` manually on new scripts I put in `microsoft/`, but it would be nice to have this as a PR acceptance test. This could make it nicer to run locally, too.

gdams:
Fixes: https://github.com/microsoft/go/issues/78

Applies some minimal fixes to get the linter green


.github/workflows/linter.yml
```diff
@@ -0,0 +1,18 @@
...
+  pull-requests: read
+
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Run golangci-lint
+        run: docker run --rm -v $(pwd):/app -w /app/eng/_util golangci/golangci-lint:v2.0 golangci-lint run -v
```
dagood: `actions/checkout` and the image are unpinned dependencies.

For the golangci-lint one, could we use a `go tool` go.mod dependency instead, so that I can be sure I'm running the same checks locally?

gdams: updated PTAL

dagood (approved):

qmuntal (approved):

eng/_util/golangci/go.mod
```diff
@@ -0,0 +1,214 @@
+// Copyright (c) Microsoft Corporation.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+module github.com/microsoft/go/_util/golangci
```
dagood: Why a new go.mod? That somewhat defeats the point of `go tool [x]` IMO.

dagood (changes requested):

eng/_util/golangci/go.mod
gdams: Quim didn't like how many dependencies the tool was pulling into our go.mod so he told me to create a separate one

dagood: While it feels reasonable, when I've considered this in the past I wasn't able to justify it to myself. If you really want to know the dependencies something uses, you must use `go list`--not every dependency in go.mod is involved in every build.

While in theory a human could scan the go.mod file and see what *could* be used, is that *actually* valuable?

We already have gone beyond manual review: for certain tools, we have tests on the dependencies used:

* https://github.com/microsoft/go/issues/402
* https://github.com/microsoft/go/pull/1287

IMO adding positive dependency tests is better than manually reviewing two go.mod files to try to "catch" bad dependencies being added due to e.g. otherwise routine updates and seemingly-small dependency changes.

---

In other cases, the reason to split the go.mod has been so that the module can be individually tagged and we can avoid spawning too many dependency updates for people who *use* that specific module. (Telemetry, install script.) Not related to what we have here.

gdams: okay, I've moved it back to the main `go.mod` file.

dagood: I'm ok leaving this out for now (just split out the lint fixes and merge them so they don't get stale) until @qmuntal has a chance to take a look at this decision.

.github/dependabot.yml
```diff
@@ -14,3 +14,8 @@ updates:
   schedule:
     interval: daily
   open-pull-requests-limit: 10
+- package-ecosystem: gomod
```
gdams: This needs removing now

eng/_util/golangci/go.mod
gdams: I've split out the lint fixes in https://github.com/microsoft/go/pull/1803

dagood: Perhaps it's enough to use the latest Docker container image and just make sure the raw image sha/digest shows up in logs, so that it *could* be reproduced locally in the unexpected case that it makes a difference to which errors are found by CI vs. locally. Pinning dependencies is my knee-jerk reaction to seeing a new one pop up, but in this case it's "just" a linter, it *should* have no breaking changes, it has no security privilege (right?), and the maintainers actively suggest not to build it from source.

gdams: I've reverted back to the docker approach. It does show the digest in the `docker run` command so this should be good enough for now

dagood (approved):

PR #1763, opened 2025-07-21
codeql: set GOEXPERIMENT=nosystemcrypto

gdams:
Without this the codeql run for linux arm32/aarch64 fails with:

```output
# crypto
../Binaries Signed/go/src/crypto/systemcrypto_nocgo.go:10:2: `
	Using a crypto backend requires CGO_ENABLED=1.
	
	` (untyped string constant "\n\tUsing a crypto backend requires CGO_ENABLED=1.\n\t\n\tFor more i...) is not used
```

Test run here: https://dev.azure.com/dnceng/internal/_build/results?buildId=2755002&view=results

### DIFF
--- a/eng/pipeline/stages/run-codeql.yml
+++ b/eng/pipeline/stages/run-codeql.yml
@@ -118,6 +118,11 @@ stages:
                 chmod 755 "${WORKAROUND_DIR}/go"
                 export PATH="${WORKAROUND_DIR}:${PATH}"
 
+                # Set GOEXPERIMENT for linux arm32/arm64 to avoid crypto backend issues
+                if [ "${{ c.os }}" = "linux" ] && ([ "${{ c.arch }}" = "arm" ] || [ "${{ c.arch }}" = "arm64" ]); then
+                  export GOEXPERIMENT=nosystemcrypto
+                fi
+
                 # Build Go itself.
                 go build std
                 go build cmd/...
--- a/eng/pipeline/stages/run-codeql.yml
+++ b/eng/pipeline/stages/run-codeql.yml
@@ -118,14 +118,11 @@ stages:
                 chmod 755 "${WORKAROUND_DIR}/go"
                 export PATH="${WORKAROUND_DIR}:${PATH}"
 
-                # Set GOEXPERIMENT for linux arm32/arm64 to avoid crypto backend issues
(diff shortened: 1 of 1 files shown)


eng/pipeline/stages/run-codeql.yml
```diff
@@ -118,6 +118,11 @@ stages:
                 chmod 755 "${WORKAROUND_DIR}/go"
                 export PATH="${WORKAROUND_DIR}:${PATH}"
 
+                # Set GOEXPERIMENT for linux arm32/arm64 to avoid crypto backend issues
+                if [ "${{ c.os }}" = "linux" ] && ([ "${{ c.arch }}" = "arm" ] || [ "${{ c.arch }}" = "arm64" ]); then
+                  export GOEXPERIMENT=nosystemcrypto
```
dagood: How about setting it for all, for now, to reach parity?

A != check rather than = seems like the way to go, otherwise.

eng/pipeline/stages/run-codeql.yml
```diff
@@ -118,6 +118,11 @@ stages:
                 chmod 755 "${WORKAROUND_DIR}/go"
                 export PATH="${WORKAROUND_DIR}:${PATH}"
 
+                # Set GOEXPERIMENT for linux arm32/arm64 to avoid crypto backend issues
+                if [ "${{ c.os }}" = "linux" ] && ([ "${{ c.arch }}" = "arm" ] || [ "${{ c.arch }}" = "arm64" ]); then
```
dagood: Use `bash`'s `env` parameter so that issues are raised at template-compile-time rather than runtime?

dagood:
* For https://github.com/microsoft/go-lab/issues/224

eng/pipeline/stages/run-codeql.yml
gdams: agreed, I've updates the job to set `GOEXPERIMENT=nosystemcrypto` for all platforms now

eng/pipeline/stages/run-codeql.yml
gdams: updated

qmuntal (approved):

dagood (approved):

PR #1765, opened 2025-07-22
Add Windows 2019 CI builder

qmuntal:
Currently we only test the windows port on Windows Server 2022.

Our CI pool infra (aka Helix .NET) also supports Windows Server 2019, so we should be running CI test on that pool too.


gdams (approved):

qmuntal:
Some pprof-related tests are failing on Windows Server 2019. Could be related to this version not having high resolution timers, see https://golang.org/issue/10842. Will mark the builder as broken until that is fixed.

gdams (approved):

eng/pipeline/stages/pool.yml
```diff
@@ -27,4 +27,8 @@ stages:
       official: ${{ parameters.inner.parameters.official }}
       os: ${{ parameters.inner.parameters.builder.os }}
       hostArch: ${{ parameters.inner.parameters.builder.hostArch }}
+      ${{ if eq(parameters.inner.parameters.builder.distro, '') }}:
+        distro: '2022' # Default distro if not specified.
```
dagood: Add a Windows condition, so we don't have a "linux 2022" flowing through the templates that *could* pop up unexpectedly in the future?

Also wondering if this should be moved to `shorthand-builders-to-builders.yml` (with an os == windows condition) and set to `hostDistro` or similar so the values propagate everywhere without pool-specific modifications?

(Like `${{ if and(not(builder.distro), eq(builder.os, 'windows')) }}: distro: '2022'`)

eng/pipeline/stages/go-builder-matrix-stages.yml
```diff
@@ -101,6 +101,7 @@ stages:
           - { experiment: nosystemcrypto, os: windows, arch: amd64, config: test }
           - { experiment: systemcrypto, os: windows, arch: amd64, config: test }
           - { experiment: systemcrypto, os: windows, arch: amd64, config: test, fips: true }
+          - { experiment: systemcrypto, os: windows, arch: amd64, config: test, distro: '2019', broken: true } # Affected by https://golang.org/issue/10842.
```
dagood, line 104: A known-broken Windows builder interacts **very** badly with the auto-retry logic, we need to make sure it doesn't retry in this case or else CI will always take a very long time.

dagood (changes requested):

eng/pipeline/stages/pool.yml
qmuntal: Makes sense, thanks!

eng/pipeline/stages/go-builder-matrix-stages.yml
qmuntal: Agree, fixed.

dagood (approved):
Looks great, thanks!

PR #1776, opened 2025-07-25
docs: Add Supported Platforms document

gdams:
First draft, will need some thorough reviews


SUPPORTED_PLATFORMS.md
```diff
@@ -0,0 +1,84 @@
...
+
+After the release of Go 1.25, Go 1.23 will no longer be supported.
+
+## Platform Support Matrix
+
+### Go 1.25 (Upcoming)
+
+| Platform | Architecture | Status | Notes |
+|----------|-------------|--------|-------|
+| Linux | amd64 | ✅ Supported | Primary platform, Azure Linux 3.0+, Ubuntu 22.04+ |
```
gdams: Not sure whether we want to call out Ubuntu at all here?

SUPPORTED_PLATFORMS.md
```diff
@@ -0,0 +1,84 @@
...
+## Platform-Specific Notes
+
+### Linux
+- **linux-amd64**: Fully supported and tested. This is the primary development and testing platform.
+- **linux-arm64**: Fully supported for ARM64-based systems.
+- **linux-armv6l**: Supported for ARM v6 systems, including Raspberry Pi devices.
+- **Minimum supported distributions**: 
+  - Azure Linux 3.0 (recommended)
+  - Ubuntu 22.04 and later
+  - Other modern Linux distributions with glibc 2.17+ or musl 1.1.16+
```
gdams: Is this glibc correct? Also not sure if musl should be called out here?

SUPPORTED_PLATFORMS.md
qmuntal: Good point. The least we officially support, the better, but I imagine many 1P customers are still using Ubuntu, so we probably need to state it as supported.

qmuntal: Also, we do support CBL-Mariner 2, even if it is deprecated (many 1P customers still use it).

SUPPORTED_PLATFORMS.md
```diff
@@ -0,0 +1,84 @@
...
+| Linux | amd64 | ✅ Supported | Primary platform, Azure Linux 3.0+, Ubuntu 22.04+ |
+| Linux | arm64 | ✅ Supported | Azure Linux 3.0+, Ubuntu 22.04+ |
+| Linux | armv6l | ✅ Supported | Modern Linux distributions |
+| Windows | amd64 | ✅ Supported | Windows Server 2016+, Windows 10+ |
+| macOS (Darwin) | amd64 | ❌ Not Available | |
+| macOS (Darwin) | arm64 | ❌ Not Available | |
+
+## Platform-Specific Notes
+
+### Linux
```
qmuntal: Also call out that we only support OpenSSL 1.1.1 and OpenSSL 3.
For OpenSSL 3, we support the built-in providers and [the SymCrypt provider](https://github.com/microsoft/SymCrypt-OpenSSL) v.1.6.0 or higher.

qmuntal (changes requested):
This document is much needed.

qmuntal (approved):

SUPPORTED_PLATFORMS.md
```diff
@@ -0,0 +1,90 @@
+# Supported Platforms
+
+This document outlines the platform support matrix for different versions of the Microsoft build of Go.
```
dagood: I think the meaning of "support" in this context needs to be clarified.

SUPPORTED_PLATFORMS.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Support Policy
+
+The Microsoft build of Go follows the upstream Go [Release Policy](https://go.dev/doc/devel/release#policy). This means we support each major release (1.X) until there are two newer major releases. A new Go major version is [released every six months](https://github.com/golang/go/wiki/Go-Release-Cycle), so each Go major version is supported for about one year.
```
dagood: Partial duplicate of information in README.md? I think sticking to platforms in this document makes more sense.

SUPPORTED_PLATFORMS.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Currently Supported Versions
+
+As of July 2025, the following Go versions are actively supported:
+
+- **Go 1.25** (upcoming release)
```
dagood: I don't believe we actively support releases that haven't released. 😄 

More generally, I would recommend removing essentially all "current" mentions in this doc to avoid it becoming a maintenance sink. IMO it's best to simply have a list of all actively supported platforms such that it covers what's in `main` and active supported branches, plus a static changelist of what version adds/drops support for specific platforms. (Like the FIPS doc.)

SUPPORTED_PLATFORMS.md
```diff
@@ -0,0 +1,90 @@
...
+
+After the release of Go 1.25, Go 1.23 will no longer be supported.
+
+## Platform Support Matrix
+
+### Go 1.25 (Upcoming)
+
+| Platform | Architecture | Status | Notes |
+|----------|-------------|--------|-------|
+| Linux | amd64 | ✅ Supported | Primary platform, Azure Linux 3.0+, Ubuntu 22.04+ |
```
dagood: What does primary mean?

SUPPORTED_PLATFORMS.md
```diff
@@ -0,0 +1,90 @@
...
+| Linux | arm64 | ✅ Supported | Azure Linux 3.0+, Ubuntu 22.04+ |
+| Linux | armv6l | ✅ Supported | Modern Linux distributions |
```
dagood: Unclear what Modern Linux distributions actually exist on armv6l, and would we actually support them? (All of them?)

SUPPORTED_PLATFORMS.md
```diff
@@ -0,0 +1,90 @@
...
+| Windows | amd64 | ✅ Supported | Windows Server 2016+, Windows 10+ |
```
dagood: Where's Windows arm64?

SUPPORTED_PLATFORMS.md
dagood: There is no specific glibc or musl, it depends on the machine that ran the build. (And the toolset itself doesn't require glibc or an equivalent, AFAIK.)

dagood: We need to test what happens to a program built on musl and running on musl, I think I'm extrapolating a bit there and I'm not sure how that works practically.

What I do know is that someone has successfully run a program built with systemcrypto on glibc on Alpine/musl once they installed the glibc-compat package. (Even without manually targeting a low glibc version when building the program.)

gdams: I'll just remove the `Other modern Linux distributions` for now

SUPPORTED_PLATFORMS.md
gdams: updated

SUPPORTED_PLATFORMS.md
gdams: agreed, removed it

SUPPORTED_PLATFORMS.md
gdams: agreed, updated

SUPPORTED_PLATFORMS.md
gdams: yeah that's weird terminology, removed

SUPPORTED_PLATFORMS.md
gdams: This one I'm not sure about, what was the original customer request for `armv6l`? Do we know what platform they required?

SUPPORTED_PLATFORMS.md
gdams: updated

gdams:
@dagood updated PTAL

SUPPORTED_PLATFORMS.md
dagood: Forwarded you the original email thread.

README.md
```diff
@@ -67,6 +67,10 @@ cross-compilation may be necessary.
 * `linux_armv6l`
 * `linux_arm64`
 * `windows_amd64`
+* `darwin_amd64`
+* `darwin_arm64`
```
dagood: These are still preview, right?

SUPPORTED_PLATFORMS.md
```diff
@@ -0,0 +1,62 @@
+# Supported Platforms
```
dagood: Is there a particular reason to put this in the root dir rather than `eng/doc`?

SUPPORTED_PLATFORMS.md
```diff
@@ -0,0 +1,62 @@
...
+
+This document outlines the platform support matrix for different versions of the Microsoft build of Go.
+
+"Support" in this context means that Microsoft actively tests these platforms and will accept bug reports for them. Platforms outside this scope are provided support on a best-effort basis with no guarantees.
```
dagood: Test in what sense? That the toolset works, or that it works at runtime? It seems like this doc is only be about runtime behavior (it lists Windows arm64, which we can target but we don't ship a prebuilt toolset for).

Do we really test Windows arm64 and Linux armv6l?

Does a listing on this page imply that there's a supported platform-provided crypto backend?

SUPPORTED_PLATFORMS.md
```diff
@@ -0,0 +1,62 @@
...
+| macOS (Darwin) | arm64 | 🔄 Preview | macOS 13+, not for production |
+
+## Platform Support History
+
+### Go 1.24
+- Added macOS (Darwin) amd64 and arm64 support in preview mode
+
+### Go 1.23
+- macOS (Darwin) platforms not available
+
```
dagood: So... they were available in 1.22? 😄

```suggestion
```

SUPPORTED_PLATFORMS.md
```diff
@@ -0,0 +1,62 @@
...
+## Platform-Specific Notes
+
+### Linux
+- **linux-amd64**: Fully supported and tested. This is the primary development and testing platform.
+- **linux-arm64**: Fully supported for ARM64-based systems.
+- **linux-armv6l**: Supported for ARM v6 systems, including Raspberry Pi devices.
+- **Minimum supported distributions**: 
+  - Azure Linux 3.0 (recommended)
+  - Ubuntu 22.04 and later
+  - CBL-Mariner 2.0
```
dagood: Isn't this a duplicate of the table?

I'm not sure a table is necessary, if we want to provide more details with more platform-specific context anyway.

README.md
dagood: Should be in the same order as https://github.com/microsoft/go/blob/microsoft/main/eng/doc/Downloads.md IMO.

qmuntal:
Common @gdams! Finish this PR 😸 🗡️

karianna:
> Common @gdams! Finish this PR 😸 🗡️

Your claw needs a grammar checker ;-)

qmuntal:
> Your claw needs a grammar checker ;-)

My claw is my 🧠, there is no grammar checker that can fix it!

gdams:
Thanks for the ping, I've significantly simplified this doc for now. We can iterate on it further in follow-up PRs

eng/doc/SupportedPlatforms.md
```diff
@@ -0,0 +1,37 @@
...
+
+| OS/Arch | Minimum Versions | Crypto Backend | Notes |
+|---------|-----------------|----------------|-------|
+| linux-amd64 | Azure Linux 3.0+, Ubuntu 22.04+ | OpenSSL | |
+| linux-arm64 | Azure Linux 3.0+, Ubuntu 22.04+ | OpenSSL | |
+| linux-armv6l | Ubuntu 22.04+ | OpenSSL | |
+| windows-amd64 | Windows 10+, Windows Server 2016+ | CNG | |
+| windows-arm64 | Windows 11+, Windows Server 2025+ | CNG | |
+| darwin-amd64 | macOS 13+ | CryptoKit/CommonCrypto | Added in Go 1.24 |
+| darwin-arm64 | macOS 13+ | CryptoKit/CommonCrypto | Added in Go 1.24 |
```
qmuntal: Follow Go conventions. Also, do we "support" linux/arm32?

```suggestion
| linux/amd64 | Azure Linux 3.0+, Ubuntu 22.04+ | OpenSSL | |
| linux/arm64 | Azure Linux 3.0+, Ubuntu 22.04+ | OpenSSL | |
| linux/arm32 | Ubuntu 22.04+ | OpenSSL | |
| windows/amd64 | Windows 10+, Windows Server 2016+ | CNG | |
| windows/arm64 | Windows 11+, Windows Server 2025+ | CNG | |
| darwin/amd64 | macOS 13+ | CryptoKit/CommonCrypto | Added in Go 1.24 |
| darwin/arm64 | macOS 13+ | CryptoKit/CommonCrypto | Added in Go 1.24 |
```

qmuntal (changes requested):

qmuntal (approved):

PR #1786, opened 2025-07-29
ed25519: check for boring enabled, and rand at the same time

xnox:
When compiling upstream projects (in particular
https://github.com/nats-io/nkeys) it is possible to reach Unreachable
code path.

It seems much safer to check for rand reader at the same time as
boring.Enabled and boring.SupportsEd25519 as otherwise automatic
fallback to the other implementation does not happen.

If the intention is to automatically allow using fallback, as the code
suggests to be the case.

Reported-by: @javacruft

### DIFF
--- a/patches/0099-ed25519-check-for-boring-enabled-and-rand-at-the-sam.patch
+++ b/patches/0099-ed25519-check-for-boring-enabled-and-rand-at-the-sam.patch
@@ -1,0 +1,59 @@
+From 3accfa5a92997343c59ccbd52cee112fa30d971f Mon Sep 17 00:00:00 2001
+From: Dimitri John Ledkov <dimitri.ledkov@surgut.co.uk>
+Date: Tue, 29 Jul 2025 21:50:56 +0100
+Subject: [PATCH] ed25519: check for boring enabled, and rand at the same time
+
+When compiling upstream projects (in particular
+https://github.com/nats-io/nkeys) it is possible to reach Unreachable
+code path.
+
+It seems much safer to check for rand reader at the same time as
+boring.Enabled and boring.SupportsEd25519 as otherwise automatic
+fallback to the other implementation does not happen.
+
+If the intention is to automatically allow using fallback, as the code
+suggests to be the case.
+---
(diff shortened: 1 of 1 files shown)


dagood:
https://github.com/microsoft/go/blob/2c7145e945502cf5e09ed6b793849820debee56f/eng/doc/fips/UserGuide.md#func-generatekey-1 does document a panic in this situation (leading me to believe there might be some reason), although I'm not sure why. @qmuntal, @gdams ?

"Unreachable" does seem wrong to me at face value, at the very least.

If it's ok to fall back (and based on what I've seen poking around, I don't see why not), the change looks good to me.

---

Ideally this change should be made to `0004-Use-crypto-backends.patch` (rebase/fixup-like) to keep the count down. (I noticed we never got rid of the individual-author `From` line, so opened https://github.com/microsoft/go/pull/1787 in in case that's misleading there.)

xnox:
> https://github.com/microsoft/go/blob/2c7145e945502cf5e09ed6b793849820debee56f/eng/doc/fips/UserGuide.md#func-generatekey-1 does document a panic in this situation (leading me to believe there might be some reason), although I'm not sure why. @qmuntal, @gdams ?
> 
> "Unreachable" does seem wrong to me at face value, at the very least.

This. When expected exception is raised, I am expecting a human readable wording and a developer focused explanation as to what is happening. Similar to the guidance shown upon unable to discover fips module, or if openssl is not FIPS capable. Etc.

> 
> If it's ok to fall back (and based on what I've seen poking around, I don't see why not), the change looks good to me.
> 

Yeah, "if it's ok to fallback" indeed.

> Ideally this change should be made to `0004-Use-crypto-backends.patch` (rebase/fixup-like) to keep the count down. (I noticed we never got rid of the individual-author `From` line, so opened #1787 in in case that's misleading there.)

I do expect it needs to be folded there, but thought it is easier to understand proposed change like this 0099 patch to not clash with any other PRs, and clearly show what change is being proposed.

And possibly tests need to be adjusted, if the unreachable is being removed.

gdams (approved):

dagood (approved):
Thanks!

PR #1791, opened 2025-07-31
Fix missing distro parameter causing Windows agent failures

gdams:
fixes: https://github.com/microsoft/go-lab/issues/227

Added missing distro: '2022' to signing and publishing stage builder objects to fix "Image 1es-windows- doesn't exist" error.


gdams:
I kicked off https://dev.azure.com/dnceng/internal/_build/results?buildId=2761884&view=results to test the changes

dagood (approved):
Would be nicer to manage these from a central place, but that's probably harder to do than it's worth due to AzDO's limited reuse features.

PR #1794, opened 2025-08-04
microbuild: switch to prod signing

gdams:
As per https://dev.azure.com/devdiv/DevDiv/_wiki/wikis/DevDiv.wiki/46279/Real-signing-with-PME-Enforcement?anchor=non-windows

Test run here: https://dev.azure.com/dnceng/internal/_build/results?buildId=2765516

@dagood I wonder if we should set SignType to `test` by default to prevent using PME in test environments? Thoughts?

### DIFF
--- a/eng/pipeline/stages/sign-stage.yml
+++ b/eng/pipeline/stages/sign-stage.yml
@@ -44,6 +44,8 @@ stages:
               signing:
                 enabled: true
                 signType: $(SigningType)
+                ${{ if eq(variables['SigningType'], 'real') }}:
+                  signWithProd: true
                 zipSources: false
                 feedSource: 'https://dnceng.pkgs.visualstudio.com/_packaging/MicroBuildToolset/nuget/v3/index.json'
             outputs:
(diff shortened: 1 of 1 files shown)


dagood:
> @dagood I wonder if we should set SignType to `test` by default to prevent using PME in test environments? Thoughts?

In the past test signing didn't work for us. We'll have to try it out again, or avoid using "signing" completely and fake it ourselves (rather than asking MicroBuild to fake it) for the sake of making the pipeline pass.

dagood:
I didn't end up including this in https://github.com/microsoft/go/pull/1795, but https://github.com/microsoft/go/pull/1795 adds a way to pass this through. (Compile-time `variables` references still have complicated undocumented behavior.)

gdams:
> I didn't end up including this in #1795, but #1795 adds a way to pass this through. (Compile-time `variables` references still have complicated undocumented behavior.)

That's fine, I'll rebase one we've merged #1795

gdams:
As expected we need to request permissions:

Screenshot 2025-08-05 at 21 58 18

I've reached out to the appropriate team

dagood:
https://dev.azure.com/dnceng/internal/_build/results?buildId=2765516&view=results looks good to me.

dagood (approved):

PR #1802, opened 2025-08-05
Stop batching the rolling pipeline CI trigger

dagood:


dagood:
If this becomes a problem, we could split this into two pipelines, one for release branches and one for the main branch. (Or rather: split the current two pipelines into four.) Hopefully not.

karianna (approved):

PR #1804, opened 2025-08-06
Document installation of Microsoft build of Go for Ubuntu 

### LINKED ISSUE #1430: document ubuntu repo
- add documentation for the packages.microsoft.com Ubuntu 22.04 jammy repo
- add a note saying more ubuntu distro support is coming soon, and go versions might lag a bit due to work needed for automation

example:
```bash
curl -SLf https://packages.microsoft.com/config/ubuntu/22.04/packages-microsoft-prod.deb > /tmp/msrepos.deb
sudo dpkg -i /tmp/msrepos.deb
sudo apt-get update && sudo apt-get install msft-golang
```

mertakman:
This PR documents installing Microsoft build of Go on Ubuntu via apt package manager.
Fixes: #1430

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ includes builds of this Go fork.
 
 ### Ubuntu
 
-To install the Microsoft Build of Go using an Ubuntu package, open a terminal and run the following commands:
+To install the Microsoft build of Go using an Ubuntu package, open a terminal and run the following commands:
 
 ```bash
 # Supported Ubuntu versions: 22.04 and 24.04.
--- a/README.md
+++ b/README.md
@@ -90,18 +90,9 @@ includes builds of this Go fork.
 
 ### Ubuntu
 
-To install the Microsoft build of Go using an Ubuntu package, open a terminal and run the following commands:
-
-```bash
-# Supported Ubuntu versions: 22.04 and 24.04.
-# For other versions, please use the tar.gz package.
-# See Microsoft's official instructions for up-to-date steps:
-# https://learn.microsoft.com/linux/packages
-ubuntu_release=`lsb_release -rs`
-wget https://packages.microsoft.com/config/ubuntu/${ubuntu_release}/packages-microsoft-prod.deb -O packages-microsoft-prod.deb
-sudo dpkg -i packages-microsoft-prod.deb
-```
-After the repository is added, install the Microsoft Build of Go by running the following commands:
+To install the Microsoft build of Go using an Ubuntu package, first set up the [Linux package repository for Microsoft Products](https://learn.microsoft.com/en-us/linux/packages).
+Packages are available in the Ubuntu 22.04 and 24.04 repositories.
+After the repository is added, install the Microsoft build of Go by running the following commands:
 
 ```bash
 sudo apt-get update && sudo apt-get install msft-golang
--- a/README.md
+++ b/README.md
@@ -92,6 +92,7 @@ includes builds of this Go fork.
 
 To install the Microsoft build of Go using an Ubuntu package, first set up the [Linux package repository for Microsoft Products](https://learn.microsoft.com/en-us/linux/packages).
 Packages are available in the Ubuntu 22.04 and 24.04 repositories.
+
 After the repository is added, install the Microsoft build of Go by running the following commands:
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -93,8 +93,10 @@ includes builds of this Go fork.
 To install the Microsoft Build of Go using an Ubuntu package, open a terminal and run the following commands:
 
 ```bash
-# Valid values are only '22.04' and '24.04'
-# For other versions of Ubuntu, please use the tar.gz package
+# Supported Ubuntu versions: 22.04 and 24.04.
+# For other versions, please use the tar.gz package.
+# See Microsoft's official instructions for up-to-date steps:
+# https://learn.microsoft.com/linux/packages
 ubuntu_release=`lsb_release -rs`
 wget https://packages.microsoft.com/config/ubuntu/${ubuntu_release}/packages-microsoft-prod.deb -O packages-microsoft-prod.deb
 sudo dpkg -i packages-microsoft-prod.deb
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ includes builds of this Go fork.
 
 ### Ubuntu
 
-To install the Microsoft Build of Go on Ubuntu 22.04 or 24.04 LTS, open a terminal and run the following commands:
+To install the Microsoft Build of Go on Ubuntu, open a terminal and run the following commands:
 ```bash
 # Valid values are only '22.04' and '24.04'
 # For other versions of Ubuntu, please use the tar.gz package
--- a/README.md
+++ b/README.md
@@ -103,6 +103,24 @@ After the repository is added, install the Microsoft Build of Go by running the 
 sudo apt-get update
 sudo apt-get install msft-golang
 ```
+Installing a Specific Version
+To install a specific version of the Microsoft Build of Go:
+
+List available versions:
+
+```bash
+apt-cache madison msft-golang
+```
+Install a specific version by specifying it explicitly:
+```bash
+sudo apt-get install msft-golang=<version>
+```
+For example:
+```bash
+sudo apt-get install msft-golang=1.22.3-1
+```
+Replace 1.22.3-1 with the version string listed in the apt-cache madison output.
+
 To verify the installation:
 ```bash
 go version
--- a/README.md
+++ b/README.md
@@ -88,6 +88,26 @@ includes builds of this Go fork.
 * In Azure Linux 2.0, the package `msft-golang` installs this fork.
 * In Azure Linux 3.0, the `golang` package installs this fork.
 
+### Ubuntu
+
+To install the Microsoft Build of Go on Ubuntu 22.04 or 24.04 LTS, open a terminal and run the following commands:
+```bash
+# Valid values are only '22.04' and '24.04'
+# For other versions of Ubuntu, please use the tar.gz package
+ubuntu_release=`lsb_release -rs`
+wget https://packages.microsoft.com/config/ubuntu/${ubuntu_release}/packages-microsoft-prod.deb -O packages-microsoft-prod.deb
+sudo dpkg -i packages-microsoft-prod.deb
+```
+After the repository is added, install the Microsoft Build of Go by running the following commands:
+```bash
+sudo apt-get update
+sudo apt-get install msft-golang
+```
+To verify the installation:
+```bash
+go version
+```
+
 ### Binary archive
 
 [Signed builds of Go](https://github.com/microsoft/go/blob/microsoft/main/eng/doc/Downloads.md)
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ includes builds of this Go fork.
 
 ### Ubuntu
 
-To install the Microsoft Build of Go on Ubuntu, open a terminal and run the following commands:
+To install the Microsoft Build of Go using an Ubuntu package, open a terminal and run the following commands:
 
 ```bash
 # Valid values are only '22.04' and '24.04'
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -88,6 +88,26 @@ includes builds of this Go fork.
 * In Azure Linux 2.0, the package `msft-golang` installs this fork.
 * In Azure Linux 3.0, the `golang` package installs this fork.
 
+### Ubuntu
+
+To install the Microsoft Build of Go on Ubuntu 22.04 or 24.04 LTS, open a terminal and run the following commands:
```
gdams:
```suggestion
To install the Microsoft Build of Go on Ubuntu, open a terminal and run the following commands:
```

README.md
```diff
@@ -88,6 +88,26 @@ includes builds of this Go fork.
...
+ubuntu_release=`lsb_release -rs`
+wget https://packages.microsoft.com/config/ubuntu/${ubuntu_release}/packages-microsoft-prod.deb -O packages-microsoft-prod.deb
+sudo dpkg -i packages-microsoft-prod.deb
+```
+After the repository is added, install the Microsoft Build of Go by running the following commands:
+```bash
+sudo apt-get update
+sudo apt-get install msft-golang
+```
+To verify the installation:
```
gdams: This is probably overkill but I'll leave it to @dagood to decide if he thinks it's useful

README.md
mertakman: Thanks

README.md
mertakman: Okey, keeping this for now.

README.md
```diff
@@ -88,6 +88,26 @@ includes builds of this Go fork.
...
+# Valid values are only '22.04' and '24.04'
+# For other versions of Ubuntu, please use the tar.gz package
+ubuntu_release=`lsb_release -rs`
+wget https://packages.microsoft.com/config/ubuntu/${ubuntu_release}/packages-microsoft-prod.deb -O packages-microsoft-prod.deb
+sudo dpkg -i packages-microsoft-prod.deb
+```
+After the repository is added, install the Microsoft Build of Go by running the following commands:
+```bash
+sudo apt-get update
+sudo apt-get install msft-golang
```
gdams: perhaps add a note on how to install a specific version?

mertakman: Sure,added

README.md
```diff
@@ -88,6 +88,44 @@ includes builds of this Go fork.
 * In Azure Linux 2.0, the package `msft-golang` installs this fork.
 * In Azure Linux 3.0, the `golang` package installs this fork.
 
+### Ubuntu
+
+To install the Microsoft Build of Go on Ubuntu, open a terminal and run the following commands:
+```bash
```
gdams: I always prefer codeblocks to be wrapped with a linebreak


```suggestion

```bash
```

README.md
```diff
@@ -88,6 +88,44 @@ includes builds of this Go fork.
...
+# Valid values are only '22.04' and '24.04'
+# For other versions of Ubuntu, please use the tar.gz package
+ubuntu_release=`lsb_release -rs`
+wget https://packages.microsoft.com/config/ubuntu/${ubuntu_release}/packages-microsoft-prod.deb -O packages-microsoft-prod.deb
+sudo dpkg -i packages-microsoft-prod.deb
+```
```
gdams, line 100: Same as above

```suggestion

```
```

File: README.md [hunk 6]
```diff
@@ -88,6 +88,44 @@ includes builds of this Go fork.
 * In Azure Linux 2.0, the package `msft-golang` installs this fork.
 * In Azure Linux 3.0, the `golang` package installs this fork.
 
+### Ubuntu
+
+To install the Microsoft Build of Go on Ubuntu, open a terminal and run the following commands:
+```bash
+# Valid values are only '22.04' and '24.04'
+# For other versions of Ubuntu, please use the tar.gz package
+ubuntu_release=`lsb_release -rs`
```
dagood: We can improve this in a few ways:

* `set -eu` to exit upon failure and detect unbound vars.
* `lsb_release` doesn't exist by default in some environments, don't use it.
* Check the requirements and tell the user about it. Don't ask the user to do it.

```sh
(
  set -eu
  ubuntu_release=$(. /etc/os-release; echo "$VERSION_ID")
  supported_versions='22.04 24.04'
  for v in $supported_versions; do
    if [ "$v" = "$ubuntu_release" ]; then
      wget "https://packages.microsoft.com/config/ubuntu/${ubuntu_release}/packages-microsoft-prod.deb" -O packages-microsoft-prod.deb
      sudo dpkg -i packages-microsoft-prod.deb
      exit 0
    fi
  done
  echo "Unsupported Ubuntu release: $ubuntu_release. Supported: $supported_versions. Consider using a different install method." >&2
  exit 1
)
```

Every Microsoft team having their own different, loose, instructions really bugs me. 😅

I know it's not typical to run in a subshell and use `set -eu` with a doc snippet, but it makes it a lot more like running a proper script, lets you use `exit`, allows normal safeguards, and avoids leaking env variables to the user's session.

We could consider including a `curl` fallback too, and handling already running as the root user (no `sudo` dependency in that case).

README.md
```diff
@@ -88,6 +88,44 @@ includes builds of this Go fork.
...
+wget https://packages.microsoft.com/config/ubuntu/${ubuntu_release}/packages-microsoft-prod.deb -O packages-microsoft-prod.deb
+sudo dpkg -i packages-microsoft-prod.deb
+```
+After the repository is added, install the Microsoft Build of Go by running the following commands:
+```bash
+sudo apt-get update
+sudo apt-get install msft-golang
```
dagood: Use `&&` to avoid easily missing errors that happen during the update. (Or use a subshell and `set -e` to exit early, but probably overkill for this line.)

```suggestion
sudo apt-get update && sudo apt-get install msft-golang
```

README.md
```diff
@@ -88,6 +88,44 @@ includes builds of this Go fork.
...
+```
+Installing a Specific Version
+To install a specific version of the Microsoft Build of Go:
```
dagood: Too much for the readme IMO. None of the other instructions in the readme file include how to get other versions.

(We could have a new file in `eng/docs` that we link to that goes into detail about installing specific Ubuntu packages if we really want to maintain this info.)

README.md
```diff
@@ -88,6 +88,44 @@ includes builds of this Go fork.
...
+```
+For example:
+```bash
+sudo apt-get install msft-golang=1.22.3-1
+```
+Replace 1.22.3-1 with the version string listed in the apt-cache madison output.
+
+To verify the installation:
+```bash
+go version
```
dagood: IMO this is too much info for the readme.

Also: this `go` could also be unrelated to the package. 😄 If you don't confirm that `go version` fails to find `go` before running the steps, you can't really know where this `go` came from.

dagood (commented):
I think in general this is probably too bulky for the main readme. The other install methods are very small and point to a doc to expand on it--none of the sections have enough standalone information to do an actual install without consulting another doc. I think this should be more like that.

README.md
```diff
@@ -88,6 +88,44 @@ includes builds of this Go fork.
 * In Azure Linux 2.0, the package `msft-golang` installs this fork.
 * In Azure Linux 3.0, the `golang` package installs this fork.
 
+### Ubuntu
+
+To install the Microsoft Build of Go on Ubuntu, open a terminal and run the following commands:
```
dagood:
```suggestion
To install the Microsoft Build of Go using an Ubuntu package, open a terminal and run the following commands:
```

Nitpick, but other methods are also valid on Ubuntu, and this is only useful to those who prefer the central package feed. (I don't tend to prefer Microsoft packages personally because of the setup complication and the risk of adding the Microsoft repository.)

README.md
gdams: I wonder if we should provide a shell script to simplify this process?

e.g

```bash
/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/microsoft/go/HEAD/setup-pmc.sh)"
```

Which would check if `packages.microsoft.com` has been added or not, setup the repo if it's not and provide a few "getting started" hints so people can then just `apt-get install msft-golang`. WFYT?

gdams: I played around with this idea here: https://github.com/microsoft/go-infra/pull/304

dagood: Piping a shell script is one of those "considered dangerous" topics, I don't think we should encourage it. (Especially from `raw.githubusercontent.com`.)

It looks like the packages.microsoft.com team has actually expanded their online presence! We should just link to https://learn.microsoft.com/en-us/linux/packages.

dagood: (And if that site doesn't look good enough, we should get them to make it good, IMO.)

dagood: > It looks like the packages.microsoft.com team has actually expanded their online presence! We should just link to https://learn.microsoft.com/en-us/linux/packages.

(Highlighting the specific action item here to make it clear because my review was re-requested. 😄)

Pointing out here which distro-versions actually have packages available is still useful, but the instructions PMC maintains should be what people use.

dagood (changes requested):

README.md
```diff
@@ -88,6 +88,25 @@ includes builds of this Go fork.
...
 * In Azure Linux 3.0, the `golang` package installs this fork.
 
+### Ubuntu
+
+To install the Microsoft Build of Go using an Ubuntu package, open a terminal and run the following commands:
+
+```bash
+# Supported Ubuntu versions: 22.04 and 24.04.
+# For other versions, please use the tar.gz package.
+# See Microsoft's official instructions for up-to-date steps:
```
dagood: IMO we shouldn't maintain our own copy of the repo setup steps at all, only a pointer to the docs, a list of which Ubuntu versions have packages available, and `sudo apt-get update && sudo apt-get install msft-golang`.

README.md
```diff
@@ -88,6 +88,25 @@ includes builds of this Go fork.
 * In Azure Linux 2.0, the package `msft-golang` installs this fork.
 * In Azure Linux 3.0, the `golang` package installs this fork.
 
+### Ubuntu
+
+To install the Microsoft Build of Go using an Ubuntu package, open a terminal and run the following commands:
```
dagood:
```suggestion
To install the Microsoft build of Go using an Ubuntu package, open a terminal and run the following commands:
```

(For some reason we have the "Microsoft Build of OpenJDK" and "Microsoft build of Go". 🤷)

README.md
```diff
@@ -88,6 +88,25 @@ includes builds of this Go fork.
...
+```bash
+# Supported Ubuntu versions: 22.04 and 24.04.
+# For other versions, please use the tar.gz package.
+# See Microsoft's official instructions for up-to-date steps:
+# https://learn.microsoft.com/linux/packages
+ubuntu_release=`lsb_release -rs`
+wget https://packages.microsoft.com/config/ubuntu/${ubuntu_release}/packages-microsoft-prod.deb -O packages-microsoft-prod.deb
+sudo dpkg -i packages-microsoft-prod.deb
+```
+After the repository is added, install the Microsoft Build of Go by running the following commands:
```
dagood:
```suggestion
To install the Microsoft build of Go using an Ubuntu package, first set up the [Linux package repository for Microsoft Products](https://learn.microsoft.com/en-us/linux/packages).
Packages are available in the Ubuntu 22.04 and 24.04 repositories.

After the repository is added, install the Microsoft build of Go by running the following commands:
```

README.md
```diff
@@ -88,6 +88,16 @@ includes builds of this Go fork.
 * In Azure Linux 2.0, the package `msft-golang` installs this fork.
 * In Azure Linux 3.0, the `golang` package installs this fork.
 
+### Ubuntu
+
+To install the Microsoft build of Go using an Ubuntu package, first set up the [Linux package repository for Microsoft Products](https://learn.microsoft.com/en-us/linux/packages).
+Packages are available in the Ubuntu 22.04 and 24.04 repositories.
+After the repository is added, install the Microsoft build of Go by running the following commands:
```
dagood, line 96:
```suggestion

After the repository is added, install the Microsoft build of Go by running the following commands:
```

Missed newline. (So that those who have the repo already set up can easily visually scan to the next step.)

dagood (approved):

README.md
mertakman: Fixed

dagood:
Thanks!

PR #1822, opened 2025-08-11
[release-branch.go1.25] Add branch to .gitmodules file

gdams:
Currently when running `git submodule update --remote`, the `go` submodule is updated to the latest `master` commit. By explictly setting the branch to be `release-branch.go1.25` this ensures we never pull in the wrong submodule commit.

### DIFF
--- a/.github/workflows/patch-build.yml
+++ b/.github/workflows/patch-build.yml
@@ -10,9 +10,10 @@ on:
   pull_request:
     branches: [ microsoft/* ]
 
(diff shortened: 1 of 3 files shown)


dagood (changes requested):
If we decide to do this, it should be automated by the branch update process.

`git submodule update --remote` seems like it would only be useful very rarely, so I don't think we should do this (and establish an expectation) without automation.

PR #1825, opened 2025-08-11
Split validation pipeline for resource trigger

dagood:

I noticed the unofficial pipeline was triggering for every build along with the official pipeline. Only the official one should trigger. The unofficial one is for dev work.

See https://github.com/microsoft/go-infra/blob/main/docs/pipeline-yml-style.md#pipeline-templates-to-reduce-duplication-with-officialunofficial-split

The split removes the need for `parameters` because we can treat unofficial as for-developers, which is lucky because we can't share parameters through pipeline yml templates.

Test official pipeline: https://dev.azure.com/dnceng/internal/_build/results?buildId=2769728&view=results
Test unofficial pipeline: https://dev.azure.com/dnceng/internal/_build/results?buildId=2769727&view=results

### DIFF
--- a/eng/pipeline/rolling-internal-validation-pipeline-unofficial.yml
+++ b/eng/pipeline/rolling-internal-validation-pipeline-unofficial.yml
@@ -1,0 +1,34 @@
+# Copyright (c) Microsoft Corporation.
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file.
+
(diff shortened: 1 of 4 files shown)


dagood:
It might not be necessary to backport this (there's little harm in triggering more builds), but I plan to, because we'll have more changes to this pipeline in the near future and this could be a source of merge conflict.

karianna (approved):

PR #1844, opened 2025-08-19
Implement ms_tls_config_schannel experiment

qmuntal:
This PR forward-ports the `ms_tls_config_schannel` feature from Go 1.23 to `main`.

For https://github.com/microsoft/go-lab/issues/215.

### DIFF
--- a/patches/0011-Implement-ms_tls_config_schannel-experiment.patch
+++ b/patches/0011-Implement-ms_tls_config_schannel-experiment.patch
@@ -1,0 +1,671 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: bot-for-go[bot] <199222863+bot-for-go[bot]@users.noreply.github.com>
+Date: Tue, 19 Aug 2025 15:16:09 +0200
+Subject: [PATCH] Implement ms_tls_config_schannel experiment
+
+---
+ src/crypto/tls/fips140_test.go                |   4 +
+ src/crypto/tls/schannel_disabled_test.go      |  11 +
+ src/crypto/tls/schannel_windows.go            | 166 +++++++++++++
+ src/crypto/tls/schannel_windows_test.go       | 225 ++++++++++++++++++
+ src/crypto/tls/tls_test.go                    |  16 ++
+ .../exp_ms_tls_config_schannel_off.go         |   8 +
+ .../exp_ms_tls_config_schannel_on.go          |   8 +
+ src/internal/goexperiment/flags.go            |   8 +
+ .../syscall/windows/security_windows.go       |  28 +++
+ .../syscall/windows/version_windows.go        |  11 +
+ .../syscall/windows/version_windows_test.go   |   7 +
+ .../syscall/windows/zsyscall_windows.go       |  16 ++
+ 12 files changed, 508 insertions(+)
+ create mode 100644 src/crypto/tls/schannel_disabled_test.go
+ create mode 100644 src/crypto/tls/schannel_windows.go
+ create mode 100644 src/crypto/tls/schannel_windows_test.go
+ create mode 100644 src/internal/goexperiment/exp_ms_tls_config_schannel_off.go
+ create mode 100644 src/internal/goexperiment/exp_ms_tls_config_schannel_on.go
+
+diff --git a/src/crypto/tls/fips140_test.go b/src/crypto/tls/fips140_test.go
+index 711ce750a71a93..05312aaf9bcb9d 100644
+--- a/src/crypto/tls/fips140_test.go
++++ b/src/crypto/tls/fips140_test.go
+@@ -14,6 +14,7 @@ import (
+ 	"crypto/x509/pkix"
+ 	"encoding/pem"
+ 	"fmt"
++	"internal/goexperiment"
+ 	"internal/obscuretestdata"
+ 	"internal/testenv"
+ 	"math/big"
+@@ -187,6 +188,9 @@ func TestFIPSServerCipherSuites(t *testing.T) {
+ 		}
+ 		serverConfig.BuildNameToCertificate()
+ 		t.Run(fmt.Sprintf("suite=%s", CipherSuiteName(id)), func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


gdams (approved):

karianna (approved):

dagood:
There are some notable changes from https://github.com/microsoft/go/blob/microsoft/release-branch.go1.23/patches/0016-implement-ms_tls_config_schannel-experiment.patch... any notes on that to make it easier to review?

qmuntal:
> There are some notable changes from https://github.com/microsoft/go/blob/microsoft/release-branch.go1.23/patches/0016-implement-ms_tls_config_schannel-experiment.patch... any notes on that to make it easier to review?

The functionality is the same, it has just been adapted to the code in master.

patches/0011-Implement-ms_tls_config_schannel-experiment.patch
```diff
@@ -0,0 +1,681 @@
...
++		return
++	}
++	cipherSuitesSchannel, versionsSchannel, err := getCipherSuitesSchannel()
++	if err != nil {
++		panic(err) // This should never happen, as Schannel is always available on Windows.
++	}
++	setSchannelCipherSuites(cipherSuitesSchannel, versionsSchannel)
++}
++
++// loadSchannelCipherSuites modifies the default cipher suites, the preference order, and the disabled cipher suites
```
dagood:
```suggestion
+// setSchannelCipherSuites modifies the default cipher suites, the preference order, and the disabled cipher suites
```

patches/0011-Implement-ms_tls_config_schannel-experiment.patch
```diff
@@ -0,0 +1,681 @@
...
++	}
++	for _, id := range suites {
++		if !slices.Contains(out, id) {
++			out = append(out, id)
++		}
++	}
++	return out
++}
++
++// disableCipherSuites modifies the disabled map adding the cippher suites in supported
```
dagood:
```suggestion
+// disableCipherSuites modifies the disabled map, adding the cipher suites in supported
```

patches/0011-Implement-ms_tls_config_schannel-experiment.patch
```diff
@@ -0,0 +1,681 @@
...
++	cipherSuitesPreferenceOrderNoAES = cipherSuitesPreferenceOrder
++	defaultCipherSuitesTLS13NoAES = defaultCipherSuitesTLS13
++
++	// Now filter the supported versions to only include those that Schannel supports.
++	supportedVersions = filterSupportedVersions(supportedVersions, versions)
++	allowedSupportedVersionsFIPS = filterSupportedVersions(allowedSupportedVersionsFIPS, versions)
++}
++
++// cipherSuitedByName returns the [CipherSuite] with the given name.
++func cipherSuitedByName(name string) (*CipherSuite, bool) {
```
dagood:
```suggestion
+// cipherSuiteByName returns the [CipherSuite] with the given name.
+func cipherSuiteByName(name string) (*CipherSuite, bool) {
```

patches/0011-Implement-ms_tls_config_schannel-experiment.patch
```diff
@@ -0,0 +1,681 @@
...
++// Copyright 2025 The Go Authors. All rights reserved.
++// Use of this source code is governed by a BSD-style
++// license that can be found in the LICENSE file.
++
++//go:build !windows || !goexperiment.ms_tls_config_schannel
++
++package tls
++
++func isSchannelCipherSuite(id uint16) bool {
++	return false
```
dagood, line 63: `panic("unreachable")` might more clearly match intent?

patches/0011-Implement-ms_tls_config_schannel-experiment.patch
```diff
@@ -0,0 +1,681 @@
...
++		}
++		if state.CipherSuite != TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 {
++			t.Error("the preference order should not depend on Config.CipherSuites")
++		}
++		if state.Version != VersionTLS12 {
++			t.Error("the version should be TLS 1.2 when TLS 1.3 is not available")
++		}
++	})
++}
++func TestRemoveCipherSuiteCurvePrefix(t *testing.T) {
```
dagood, line 448: Newline bewteen funcs, name doesn't match.

```suggestion
+
+func TestRemoveCipherSuiteCurveSuffix(t *testing.T) {
```

dagood (approved):
Essentially just formatting suggestions. 👍

patches/0011-Implement-ms_tls_config_schannel-experiment.patch
qmuntal: This function can be called from files not gated behind the `ms_tls_config_schannel` build tag, in which case it has to return false.

patches/0011-Implement-ms_tls_config_schannel-experiment.patch
```diff
@@ -0,0 +1,681 @@
...
++	if err != nil {
++		return nil, nil, err
++	}
++	defer windows.BCryptFreeBuffer(unsafe.Pointer(funcs))
++
++	suites = make([]uint16, 0, funcs.Count)
++	for i := range funcs.Count {
++		name := windows.UTF16PtrToString(funcs.At(int(i)))
++		name = removeCipherSuiteCurveSuffix(name)
++		suite, ok := cipherSuitedByName(name)
```
qmuntal:
```suggestion
+		suite, ok := cipherSuiteByName(name)
```

patches/0011-Implement-ms_tls_config_schannel-experiment.patch
dagood: It *could* be called, but it isn't called from anywhere when the experiment isn't active, right?

I don't understand why `false` makes sense here--to me that means "this isn't an algorithm schannel supports", but in this case we don't even know what schannel would support. I could see `true` making just as much sense depending on what this func is actually used for. (But it isn't used in either way, so... no context clues.)

A comment could clear it up fine if `false` really seems best.

Issue #262, opened 2021-11-04
go version should inform that the go binary is generated by microsoft 


qmuntal:
The `go version` is used to know which is the version of `go` binary being used and it can also discover which `go` version was used to generated a binary. This version is taken from a `VERSION` file in the root dir, and if it does not exist it is generated on the fly following this logic: https://github.com/golang/go/blob/577287728a8e8abbcbc6c96fdd49feb5a99ced0c/src/cmd/dist/build.go#L348

For example, running `go version` on my computers it outputs `go version devel go1.18-505a7c80f2 Wed Nov 3 18:06:56 2021 -0500 windows/amd64`.

We should include Microsoft-specific information in that output, such as the `microsoft` tag and probably a AzDO build number (when doing intermediate releases).

I propose to do it as follows:
- `microsoft/main` does not have any committed `VERSION` file. 
- Our builders will generate on-the-fly with whatever pattern we decide IF the `VERSION` file does not exist. This auto-generated file will contain at least a `microsoft` tag and the AzDO build number.
- We will commit a hand-made `VERSION` file on every release branch with at least the `microsoft` tag and the released version number.
- `boringcrypto` and `openssl` branches will have a special tag that clearly identifies them

This approach has the nice property that no upstream code has to be modified as all the logic will be put on `microsoft/go-infra`. 

(Forked from #140)

dagood:
Minimal effort change to at least have *something* that's easy to spot:

We currently only include a `MICROSOFT_REVISION` file if the revision is necessary information--that is, if it's not `1`. We can simply always include the file even if it's `1`.

https://github.com/microsoft/go-infra/blob/7578a79a400497eb9d42a5a6fa745ac68400a3fd/sync/sync.go#L573-L581

A secondary benefit is that our builds are more similar between `1` and `2` revisions. This means we won't find ourselves in an awkward situation where the presence of the file itself causes a build break and we have a hard time releasing a `-2` build. Such a problem could otherwise be dormant for a while because we don't release new Microsoft revisions all that often.

dagood:
> Minimal effort change to at least have _something_ that's easy to spot:
> 
> We currently only include a `MICROSOFT_REVISION` file if the revision is necessary information--that is, if it's not `1`. We can simply always include the file even if it's `1`.

Actually, this doesn't make anything show up in the packed Go. This file never gets included. I was probably mixing something up with how it worked before we started using upstream's distpack tooling.

(The secondary reasons for this change are still valid.)

[...]

PR #1857, opened 2025-09-04
Include build number in go version output

gdams:
for: #262

Sample output:

```output
go version             
go version go1.26-devel_c1b4d756f4 Thu Sep 4 14:06:31 2025 +0100 (msft: 12345678) darwin/arm64
```


dagood (changes requested):
The raw build number isn't meant to be meaningful to our users--we should emit the version that includes the Microsoft build revision.

Also, is the result of `runVersion` really baked into the binary? Or is it run every time (so env can't be used)?

I think we should also test that this version makes it into binaries built by the Microsoft build of Go.

gdams:
speaking with @qmuntal I actually think it's worth raising an upstream proposal to include the vendor and optional build version in the `go version` output. This would be useful for all go vendors and would help the golang community differentiate their builds from other ones (e.g Ubuntu, Red Hat, Microsoft etc)

dagood:
> include the vendor and optional build version in the `go version` output

Not just `go version`, also a way to standardize embedding it in binaries so you can tell with `go version -m`, IMO.

Issue #1871, opened 2025-09-15
BN_lebin2bn used, but not available in AWS-LC-FIPS


xnox:
### Microsoft build of Go version

go1.25

### What is your operating system and platform?

Some Amazon Bottlerocket, likely 1.45

### Output of `go env` in your module/workspace:

```shell
n/a
```

### What did you do?

Build and run go binary on bottlerocket 1.45

### What did you see happen?

Cannot get required symbol "BN_lebin2bn" from libcrypto version 1.1
SIGABRT: abort
...
vendor/github.com/golang-fips/openssl/v2._Cfunc_go_openssl_load_functions
...

### What did you expect to see?

Some sensible warning that OpenSSL 1.0 is no longer supported.

Or the application just working, and using other symbols that are available in OpenSSL 1.0.2k-fips  26 Jan 2017

Based on known-versions I thought 1.0.2 is still supported.

Maybe ECDSA code paths are triggered, when ECDSA is not supported at all, or some such? Or is this an issue from bind-now linking?

dagood:
> Cannot get required symbol "BN_lebin2bn" from libcrypto version 1.1

I'm a little puzzled by this string. The "from libcrypto version" bit is present in 1.24, but not 1.25. (Or at least I can't find a place where it might come from in 1.25.) The `1.1` seems to indicate that it found OpenSSL 1.1 and is trying to load it, not a 1.0. Maybe the system has a weird symlink from 1.1 -> 1.0? Could try [`GO_OPENSSL_VERSION_OVERRIDE`](https://github.com/microsoft/go/blob/microsoft/main/eng/doc/fips/README.md#runtime-openssl-version-override) to work around that if that's the case.

As you suspect, 1.0 is no longer supported in 1.25.

* https://github.com/golang-fips/openssl/pull/248
* https://github.com/golang-fips/openssl/issues/244

---

About support: https://github.com/golang-fips/openssl?tab=readme-ov-file#multiple-openssl-versions-supported is up to date, but IMO we're missing updates to:

* https://github.com/microsoft/go/blob/microsoft/main/eng/doc/fips/README.md#go-125-aug-2025
* https://github.com/microsoft/go/blob/microsoft/main/eng/doc/fips/README.md#multiple-openssl-versions-allowed
* https://github.com/microsoft/go/blob/27cc46e34c32861299b408970b59cc97eac06d28/patches/0003-Implement-crypto-internal-backend.patch#L2000
[...]

PR #1881, opened 2025-09-19
Stop trying to load OpenSSL 1.0, update docs

dagood:
Addresses most changes I mentioned in:

* https://github.com/microsoft/go/issues/1871

Should probably be backported to 1.25 to avoid odd loading/usage errors with OpenSSL versions we don't really support.


karianna (approved):

qmuntal (approved):
Thanks

PR #1884, opened 2025-09-24
crypto/internal/backend: add deprecation warnings for specific crypto backends

gdams:
Add deprecation warnings for opensslcrypto, cngcrypto, and darwincrypto GOEXPERIMENT values in CI environments (GitHub Actions and Azure DevOps). These specific backend experiments are deprecated in favor of systemcrypto, which automatically selects the appropriate backend for the target platform.

The warnings use CI-specific formats:
- GitHub Actions: ::warning:: annotation
- Azure DevOps: ##vso[task.logissue type=warning] command

Warnings are only emitted when:
1. Building in a detected CI environment
2. The deprecated backend was explicitly specified by the user (not auto-selected by systemcrypto)

This helps users migrate away from the specific backend experiments without breaking existing tooling that might parse stdout/stderr, and avoids warning when systemcrypto automatically selects a backend.

A few test outputs to confirm functionality:

```bash

# GitHub actions env, using systemcrypto experiment
GITHUB_ACTIONS=true GOEXPERIMENT=systemcrypto ./go/bin/go run test_crypto.go 
SHA256: 2cf24dba5fb0a30e26e83b2ac5b9e29e1b161e5c1fa7425e73043362938b9824

# GitHub actions env, using darwincrypto experiment (expert a warning)
GITHUB_ACTIONS=true GOEXPERIMENT=darwincrypto ./go/bin/go run test_crypto.go
::warning::GOEXPERIMENT=darwincrypto is deprecated. Use GOEXPERIMENT=systemcrypto instead, or remove GOEXPERIMENT entirely as systemcrypto is now the default.
SHA256: 2cf24dba5fb0a30e26e83b2ac5b9e29e1b161e5c1fa7425e73043362938b9824

# GitHub actions env, providing no GOEXPERIMENT (default systemcrypto)
GITHUB_ACTIONS=true ./go/bin/go run test_crypto.go     
SHA256: 2cf24dba5fb0a30e26e83b2ac5b9e29e1b161e5c1fa7425e73043362938b9824

# GitHub actions env, using nosystemcrypto experiment
GITHUB_ACTIONS=true GOEXPERIMENT=nosystemcrypto ./go/bin/go run test_crypto.go
SHA256: 2cf24dba5fb0a30e26e83b2ac5b9e29e1b161e5c1fa7425e73043362938b9824


# Azure Devops env, using systemcrypto experiment
TF_BUILD=true GOEXPERIMENT=systemcrypto ./go/bin/go run test_crypto.go
SHA256: 2cf24dba5fb0a30e26e83b2ac5b9e29e1b161e5c1fa7425e73043362938b9824

# Azure Devops env, using darwincrypto experiment (expert a warning)
TF_BUILD=true GOEXPERIMENT=darwincrypto ./go/bin/go run test_crypto.go
##vso[task.logissue type=warning]GOEXPERIMENT=darwincrypto is deprecated. Use GOEXPERIMENT=systemcrypto instead, or remove GOEXPERIMENT entirely as systemcrypto is now the default.
SHA256: 2cf24dba5fb0a30e26e83b2ac5b9e29e1b161e5c1fa7425e73043362938b9824

# Azure Devops env, providing no GOEXPERIMENT (default systemcrypto)
TF_BUILD=true ./go/bin/go run test_crypto.go     
SHA256: 2cf24dba5fb0a30e26e83b2ac5b9e29e1b161e5c1fa7425e73043362938b9824

# Azure Devops env, using nosystemcrypto experiment
TF_BUILD=true GOEXPERIMENT=nosystemcrypto ./go/bin/go run test_crypto.go
SHA256: 2cf24dba5fb0a30e26e83b2ac5b9e29e1b161e5c1fa7425e73043362938b9824


# Non Azure Devops / GitHub actions env using  using darwincrypto experiment (expect no warning)
GOEXPERIMENT=darwincrypto ./go/bin/go run test_crypto.go
SHA256: 2cf24dba5fb0a30e26e83b2ac5b9e29e1b161e5c1fa7425e73043362938b9824
```

### DIFF
--- a/patches/0011-Add-deprecation-warnings-for-specific-crypto-backend.patch
+++ b/patches/0011-Add-deprecation-warnings-for-specific-crypto-backend.patch
@@ -1,8 +1,7 @@
 From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
 From: bot-for-go[bot] <199222863+bot-for-go[bot]@users.noreply.github.com>
 Date: Wed, 24 Sep 2025 12:46:04 +0100
-Subject: [PATCH] crypto/internal/backend: add deprecation warnings for
- specific crypto backends
+Subject: [PATCH] Add deprecation warnings for specific crypto backends
 
 Add deprecation warnings for opensslcrypto, cngcrypto, and darwincrypto
 GOEXPERIMENT values in CI environments (GitHub Actions and Azure DevOps).
@@ -24,9 +23,9 @@ warning when systemcrypto automatically selects a backend.
 ---
  src/crypto/internal/backend/cng_windows.go    |  4 +
  src/crypto/internal/backend/darwin_darwin.go  |  4 +
- .../backend/deprecation/deprecation.go        | 89 +++++++++++++++++++
+ .../backend/deprecation/deprecation.go        | 96 +++++++++++++++++++
  .../internal/opensslsetup/opensslsetup.go     |  4 +
- 4 files changed, 101 insertions(+)
+ 4 files changed, 108 insertions(+)
  create mode 100644 src/crypto/internal/backend/deprecation/deprecation.go
 
 diff --git a/src/crypto/internal/backend/cng_windows.go b/src/crypto/internal/backend/cng_windows.go
@@ -75,10 +74,10 @@ index fea0f284de2439..3b8da177cf3c40 100644
  		panic("darwincrypto: " + err.Error())
 diff --git a/src/crypto/internal/backend/deprecation/deprecation.go b/src/crypto/internal/backend/deprecation/deprecation.go
 new file mode 100644
-index 00000000000000..dd19b51a8112f1
+index 00000000000000..0e02fcad6bffa1
 --- /dev/null
 +++ b/src/crypto/internal/backend/deprecation/deprecation.go
-@@ -0,0 +1,89 @@
+@@ -0,0 +1,96 @@
 +// Package deprecation provides utilities for emitting deprecation warnings
 +// in CI environments without affecting normal program output.
 +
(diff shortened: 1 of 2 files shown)


gdams:
I managed to test this by switching one of our test jobs to use `cngcrypto` temporarily:

Screenshot 2025-09-24 at 13 02 08

dagood (changes requested):
This is better than emitting some unformatted string to stdout, but only because it's more likely to be seen, as far as I can tell. It can still cause breakage:

* It's still something being printed to a std output stream, just in a way that some environments can automatically parse.
* A warning is often considered an error in principle, even if that logic is not directly implemented in AzDO/Actions.
* Arguably it's worse to detect environment because then it's less reproducible. If the output is causing an error in a non-obvious way (e.g. if stderr is parsed by something else, never reported directly and therefore never seen by AzDO/Actions), this could be quite a challenge to debug.
  * Or: it might simply show up as an error, which is already bad.

But even without that, I don't think we have justification for removing or even discouraging the old experiment values. What's the problem if people keep using them? Who benefits from this warning?

patches/0011-Add-deprecation-warnings-for-specific-crypto-backend.patch
```diff
@@ -0,0 +1,126 @@
...
++		fmt.Fprintf(os.Stderr,
++			"::warning title=GOEXPERIMENT_DEPRECATED::GOEXPERIMENT=%s is deprecated. Use GOEXPERIMENT=systemcrypto instead, or remove GOEXPERIMENT entirely as systemcrypto is now the default.\n",
++			backend)
++	} else if isAzureDevOps() {
++		fmt.Fprintf(os.Stderr,
++			"##vso[task.logissue type=warning;code=GOEXPERIMENT_DEPRECATED;]GOEXPERIMENT=%s is deprecated. Use GOEXPERIMENT=systemcrypto instead, or remove GOEXPERIMENT entirely as systemcrypto is now the default.\n",
++			backend)
++	}
++	// For other environments, we don't emit warnings to avoid breaking
++	// users who might be parsing stdout/stderr
```
dagood, line 103: I don't think this logic holds: I would expect child processes to inherit the variables, so e.g. even if `isAzureDevOps()` is true, there might be a wrapper process doing parsing.

gdams:
Closing as we will panic in go 1.27

PR #1886, opened 2025-09-25
Support disabling systemcrypto using the MS_GO_NOSYSTEMCRYPTO env

### LINKED ISSUE #1880: Allow specifying `nosystemcrypto` in a way that's compatible with upstream "go build"
As of Microsoft build of Go 1.25, some people need to specify `GOEXPERIMENT=nosystemcrypto` to opt out of the now on-by-default `systemcrypto`. This makes it so upstream Go fails to build:
```
$ export GOEXPERIMENT=nosystemcrypto
$ msgo1.25.1-1 build .
$ go1.25.1 build .
go: unknown GOEXPERIMENT systemcrypto
```

If we were trying to *enable* an experiment, we could use `-tags` (or `GOFLAGS` with `-tags`) to put our environment in a state where either build will work. However, there's no mechanism like this for *disabling* an experiment. Currently you need to conditionally enable `nosystemcrypto`. In a lot of cases this is easy to do (just enable it for the CI pipelines that use the Microsoft build of Go), but in some, it's not.

As of writing, we think it makes sense to add a build-time env var like `MS_NOSYSTEMCRYPTO=1` (or `MS_STDLIBCRYPTO=1`, etc...) so there's a universal, simple way to opt out that doesn't make the upstream toolkit fail.

qmuntal:
Disabling `systemcrypto` via `GOEXPERIMENT=nosystemcrypto` has some tricky corner cases (see #1880). We better provide a dedicated environment variable for it. The side benefit is that this goes in the direction of decoupling systemcrypto from the GOEXPERIMENT machinery.

Closes #1880

### DIFF
--- a/eng/doc/MigrationGuide.md
+++ b/eng/doc/MigrationGuide.md
@@ -153,7 +153,7 @@ In this case, you should first try to install a C compiler, like `gcc`.
 
 You may also need to set the `CGO_ENABLED` environment variable to `1` or [otherwise enable cgo](https://pkg.go.dev/cmd/cgo).
 
-If this isn't feasible, see [migration to `systemcrypto`](#migration-to-systemcrypto).
+If this isn't feasible, see [disabling systemcrypto](#disabling-systemcrypto).
 
 > [!NOTE]
 > The macOS `systemcrytpo` implementation also requires cgo.
@@ -203,7 +203,7 @@ sudo tdnf install gcc glibc-devel binutils kernel-headers
 
 Alternatively, the `build-essential` package contains all of these packages and more.
 
-If this isn't feasible, see [migration to `systemcrypto`](#migration-to-systemcrypto).
+If this isn't feasible, see [disabling systemcrypto](#disabling-systemcrypto).
 
 #### Unknown GOEXPERIMENT systemcrypto
 
@@ -267,7 +267,7 @@ Otherwise, identify why these packages are not present.
 
 If your system has a `libcrypto.so[...]` file that doesn't follow [the expected naming conventions, you can set the `GO_OPENSSL_VERSION_OVERRIDE` environment variable](fips/README.md#runtime-openssl-version-override) to make your Go program look for a specific suffix.
 
-If this isn't feasible, see [migration to `systemcrypto`](#migration-to-systemcrypto).
+If this isn't feasible, see [disabling systemcrypto](#disabling-systemcrypto).
 
 #### FIPS mode requested but not available
 
@@ -327,7 +327,7 @@ It uses `musl` instead of `glibc`.
 Try installing the `gcompat` or `libc6-compat` Alpine packages to use a `glibc` compatibility layer.
 Gathering more information about behavior on Alpine is tracked by [microsoft/go#1867](https://github.com/microsoft/go/issues/1867).
 
-If this isn't feasible, see [migration to `systemcrypto`](#migration-to-systemcrypto).
+If this isn't feasible, see [disabling systemcrypto](#disabling-systemcrypto).
 
 #### Cryptography package failures while in FIPS mode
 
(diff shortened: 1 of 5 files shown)


eng/doc/MigrationGuide.md
```diff
@@ -359,13 +354,13 @@ For specific guidance within Microsoft:
 - Read [Microsoft.Security.Cryptography.10010 on the Liquid Microsoft-internal site.][msc10010]
 - Contact the crypto board
 
-## Migration to `systemcrypto`
+## Disabling `systemcrypto`
```
dagood, line 357: There are some broken links to this section. (I do think the rename is reasonable though.)

eng/doc/MigrationGuide.md
```diff
@@ -359,13 +354,13 @@ For specific guidance within Microsoft:
...
 
 The difficulty of migrating to using `systemcrypto` can vary significantly depending on the Go project.
 If the change requires further planning and if it's acceptable for your project to be temporarily out of compliance with Microsoft cryptography policy, you can disable `systemcrypto`.
 
-To do so, set the `GOEXPERIMENT` environment variable to `nosystemcrypto`.
-If you have already set `GOEXPERIMENT`, append `,nosystemcrypto` to the existing value.
+To do so, set the `MS_GO_NOSYSTEMCRYPTO` environment variable to `1` if using Go 1.25.2 or later, else set the `GOEXPERIMENT` environment variable to `nosystemcrypto`.
+In the latter case, if you have already set `GOEXPERIMENT`, append `,nosystemcrypto` to the existing value.
 After that, build commands won't encounter errors related to `systemcrypto`, and the resulting program won't attempt to use system-provided cryptography at runtime.
```
dagood, line 366: GitHub won't let me do this as a suggestion because it spans some deletes, but I think this is cramped. How about:

---

If the change requires further planning and if it's acceptable for your project to be temporarily out of compliance with Microsoft cryptography policy, you can disable `systemcrypto` by following these instructions:

* If you're using Go 1.25.2 or later, set the `MS_GO_NOSYSTEMCRYPTO` environment variable to `1`.
* Otherwise, set the `GOEXPERIMENT` environment variable to `nosystemcrypto`.
  * If you have already set `GOEXPERIMENT`, append `,nosystemcrypto` to the existing value.

After that, build commands won't encounter errors related to `systemcrypto`, and the resulting program won't attempt to use system-provided cryptography at runtime.

eng/doc/MigrationGuide.md
```diff
@@ -359,13 +354,13 @@ For specific guidance within Microsoft:
...
 
-## Migration to `systemcrypto`
+## Disabling `systemcrypto`
 
 The difficulty of migrating to using `systemcrypto` can vary significantly depending on the Go project.
 If the change requires further planning and if it's acceptable for your project to be temporarily out of compliance with Microsoft cryptography policy, you can disable `systemcrypto`.
 
-To do so, set the `GOEXPERIMENT` environment variable to `nosystemcrypto`.
-If you have already set `GOEXPERIMENT`, append `,nosystemcrypto` to the existing value.
+To do so, set the `MS_GO_NOSYSTEMCRYPTO` environment variable to `1` if using Go 1.25.2 or later, else set the `GOEXPERIMENT` environment variable to `nosystemcrypto`.
```
dagood: I think it's probably ok to update the FIPS readme with theoretical future version information (but we should probably mark it as "planned" in the changelog?), but not so sure about the migration guide. Could cause confusion: thinking I'm on an old version even though I think I'm up to date (because I am).

patches/0002-Add-crypto-backend-GOEXPERIMENTs.patch
```diff
@@ -387,10 +387,38 @@ index 00000000000000..eb8a026982259c
...
+ 	"fmt"
++	"os"
+ 	"reflect"
+ 	"strings"
+ 
+@@ -29,6 +30,19 @@ type ExperimentFlags struct {
+ // default in the current toolchain. This is, in effect, the "control"
+ // configuration and any variation from this is an experiment.
+ var Experiment ExperimentFlags = func() ExperimentFlags {
++	if os.Getenv("MS_GO_NOSYSTEMCRYPTO") == "1" {
```
dagood, line 405: Hmm, wondering if this needs to be added to `src/cmd/go/internal/envcmd/env.go` or somewhere to invalidate the build cache when it changes. It seems like since all it does is (always) modify `GOEXPERIMENT`, which is on that list, it might be fine.

eng/doc/MigrationGuide.md
qmuntal: Good catch.

patches/0002-Add-crypto-backend-GOEXPERIMENTs.patch
qmuntal: `GOEXPERIMENT` is modified at the very beginning of the build invocation, before the cache comes into play, so we are fine.

eng/doc/MigrationGuide.md
qmuntal: Agree in general, for this particular case, 1.25.2 is around the corner, so I'm OK leaving this sentence as is.

eng/doc/MigrationGuide.md
```diff
@@ -359,13 +354,17 @@ For specific guidance within Microsoft:
 - Read [Microsoft.Security.Cryptography.10010 on the Liquid Microsoft-internal site.][msc10010]
 - Contact the crypto board
 
-## Migration to `systemcrypto`
+## Disabling `systemcrypto`
 
 The difficulty of migrating to using `systemcrypto` can vary significantly depending on the Go project.
 If the change requires further planning and if it's acceptable for your project to be temporarily out of compliance with Microsoft cryptography policy, you can disable `systemcrypto`.
```
dagood: Duplicate sentence, should be joined up!

(I'm intending for this suggestion to also remove the newline by selecting both lines, but it looks like GitHub might not apply it right based on the preview...)

```suggestion
```

dagood (approved):

eng/doc/MigrationGuide.md
qmuntal: It did remove the newline 😄.

PR #1899, opened 2025-10-08
add Microsoft Build of Go - Additional Features document

gdams:
Adds a document that outlines the support status of each of out features (experiments)


dagood:
Incorporated in https://github.com/microsoft/go/pull/2086

Issue #1073, opened 2023-10-27
Generate build pipeline YML to avoid unclear template usage


dagood:
@gdams mentioned yml generation in a meeting (another team uses this general approach), and I really like it. So, I wrote it up as a pseudo-ADR, picking [MADR](https://github.com/adr/madr/blob/develop/template/adr-template.md) in particular, to try out that structure. (Although there are some "I" and clear opinions that I think I'd swap out if I were writing this to check in.)

## Context and Problem Statement

We use AzDO Pipeline templates a lot. To overcome reusability limitations in the pipeline yml language, we have templates that accept a "inner" template's filename as a parameter and calls the "inner" template with some additional parameters. This can be quite hard to unravel when reading straight through without already knowing the end goal.

This works ok once you're used to it, and sticking to template-time logic avoids a lot of other odd AzDO Pipelines evaluation behavior. However, adding additional layers of parameters-based-on-parameters is a significant mental tax for both implementers and readers, encouraging copy-paste and quick hacks over reuse, and resulting in unclear template name distinctions like `pool.yml` and `pool-core.yml` that only exist to create two template contexts.

This applies to microsoft/go for build/test/sign/publish, but also https://github.com/microsoft/go-infra for release automation pipelines. Those also use complicated pipeline yml patterns that would benefit from generation, perhaps even more than microsoft/go pipelines.

## Decision Drivers

* We want the pipelines to be maintainable.
* People new to the team should be able to follow how each pipeline works with minimal AzDO pipeline experience.

## Considered Options

* Leave it as it is. It's worked so far.
* Generate build pipeline YML using Go code.

## Decision Outcome

Generate build pipeline YML.

### Consequences

* Go is a much more featureful language than AzDO pipeline yml. We can more concisely and more clearly express the list of builders and the tweaks necessary for each platform/configuration.
* Someone who looks at our repo can easily read through the runnable pipeline yml and see the steps that are taken for a particular builder.
* We can have more confidence about how a change to core logic will affect (or not affect) every builder. This helps confidence with small fixes that are only intended to affect one platform.
  * It is technically possible to download AzDO-evaluated yml from an AzDO build, but that requires URL manipulation to get the data and the output is difficult to correlate to the actual templates (in my experience).
* We can include builder-specific details in yml comments to guide the reader, without compromising the maintainability of the yml.
[...]

PR #1938, opened 2025-11-04
Add pipelineymlgen; fix 1ES officialness of validation pipelines

dagood:
* Minimally introduce https://github.com/microsoft/go/issues/1073
* Address https://github.com/microsoft/go-lab/issues/254 for this particular pipeline, but there are more in other repos.
  * Specifically, https://github.com/microsoft/go-lab/issues/270

Incorporates https://github.com/microsoft/go-infra/tree/main/cmd/pipelineymlgen. Not the cleanest incorporation because we can't put our tooling Go module at the root of this repository. Execution happens through `pwsh eng/run.ps1 pipelineymlgen` to make it work the same as other generators we currently use in this repository.

Adds reproducibility evaluation to the existing `eng/_util` self-test suite. Since it relies on external files, re-runs aren't reliable, so added `-count 1` support to the self-test runner. This isn't a problem in CI because the runs are always fresh.

This only uses the tool to address a very specific problem with the behavior of `${{ if variables['Go1ESOfficial'] }}:` in the shared template by breaking up that template while avoiding duplication. I think there's still a significant amount of low-hanging fruit in this repo's pipelines for simplification later.

Test official pipeline: https://dev.azure.com/dnceng/internal/_build/results?buildId=2831266&view=results
Test unofficial pipeline: https://dev.azure.com/dnceng/internal/_build/results?buildId=2831265&view=results


karianna (approved):

karianna (dismissed):
Pipeline seems to fail

dagood:
The problem in CI is:

* https://github.com/microsoft/go-infra/pull/373

gdams (approved):

PR #1978, opened 2025-12-02
Implement upgrader tool

qmuntal:
Upgrading one or more vendored dependencies is always tedious and error prone. You have to go to the repo and copy the latest commit sha, then build the proper module path (without forgetting that the openssl module is v2), and then run `go mod tidy` and `go mod vendor`. Some dependencies are even in different folders, i.e. you can upgrade the telemetry module from `go/src`, but you need to be in `go/src/cmd`.

All this cumbersomeness can be easily automated, as all inputs either static or can be automatically retrieved using `git ls-remote`.

This PR adds a tool that upgrades one or more dependencies by just providing their well-known name. For example, to upgrade all the backends you can run the tool as follows:

```
go run eng/_util/cmd/upgrader/main.go openssl windows darwin
```

### DIFF
--- a/eng/_util/cmd/upgrader/main.go
+++ b/eng/_util/cmd/upgrader/main.go
@@ -1,0 +1,137 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"log"
+	"maps"
+	"os"
+	"os/exec"
+	"slices"
+	"strings"
+	"unicode"
+)
+
+type depsInfo struct {
+	url string // Repository URL
+	mod string // Module name
+	wd  string // Relative working directory from repo root
+}
+
+var infos = map[string]depsInfo{
+	"openssl": {
+		url: "https://github.com/golang-fips/openssl",
+		mod: "github.com/golang-fips/openssl/v2",
+		wd:  "./go/src",
+	},
+	"windows": {
+		url: "https://github.com/microsoft/go-crypto-winnative",
+		mod: "github.com/microsoft/go-crypto-winnative",
+		wd:  "./go/src",
+	},
+	"darwin": {
+		url: "https://github.com/microsoft/go-crypto-darwin",
+		mod: "github.com/microsoft/go-crypto-darwin",
+		wd:  "./go/src",
+	},
+	"telemetry": {
+		url: "https://github.com/microsoft/go-infra",
+		mod: "github.com/microsoft/go-infra/telemetry",
+		wd:  "./go/src/cmd",
+	},
+	"telemetryconfig": {
+		url: "https://github.com/microsoft/go-infra",
+		mod: "github.com/microsoft/go-infra/telemetry/config",
+		wd:  "./go/src/cmd",
+	},
+}
+
+func description() string {
+	deps := slices.Collect(maps.Keys(infos))
+	slices.Sort(deps)
+	return `
+This command upgrades the dependencies listed as arguments to their latest available commit.
+Known dependencies are:
+` + strings.Join(deps, " ") + `
+
+For example, to upgrade all crypto backends, run:
+	go run eng/_util/cmd/upgradedeps/main.go openssl windows darwin
+`
+}
+
+func main() {
+	help := flag.Bool("h", false, "Print this help message.")
+
+	flag.Usage = func() {
+		fmt.Fprintf(flag.CommandLine.Output(), "Usage of upgradedeps:\n")
+		flag.PrintDefaults()
+		fmt.Fprintf(flag.CommandLine.Output(), "%s\n", description())
+	}
+
+	flag.Parse()
+	if *help {
+		flag.Usage()
+		return
+	}
+
+	deps := flag.Args()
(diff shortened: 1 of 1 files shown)


gdams (approved):

eng/_util/cmd/upgrader/main.go
```diff
@@ -0,0 +1,137 @@
+package main
```
dagood, line 5:
```suggestion
// Copyright (c) Microsoft Corporation.
// Use of this source code is governed by a BSD-style
// license that can be found in the LICENSE file.

package main
```

eng/_util/cmd/upgrader/main.go
```diff
@@ -0,0 +1,137 @@
...
+
+	// Tidy-up go.mod
+	cmd = exec.Command("go", "mod", "tidy")
+	cmd.Dir = info.wd
+	if out, err = cmd.CombinedOutput(); err != nil {
+		return fmt.Errorf("failed to tidy go.mod: %v, output: %s", err, out)
+	}
+
+	// Vendor dependencies
+	cmd = exec.Command("go", "mod", "vendor")
```
dagood: Should this command try to find `go` in `./go/bin/go` before (or rather than?) looking at the user's PATH's `go`? I for one don't normally have the dev build of Go as my ambient version, although it does depend on context.

eng/_util/cmd/upgrader/main.go
```diff
@@ -0,0 +1,137 @@
...
+		}
+	}
+}
+
+func upgradeDependency(info *depsInfo) error {
+	// Placeholder for actual upgrade logic
+	log.Printf("Upgrading dependency: %s\n", info.mod)
+
+	// Fetch latest commit from default branch
+	out, err := exec.Command("git", "ls-remote", info.url, "HEAD").CombinedOutput()
```
dagood: It seems like this will only be useful in `microsoft/main`, not for upgrading major versions where there may be (should be) a version-specific support branch in some repos. Maybe take branch version as an input and check for a matching branch first?

eng/_util/cmd/upgrader/main.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	fields := strings.FieldsFunc(string(out), unicode.IsSpace)
+	if len(fields) == 0 {
+		return fmt.Errorf("failed to parse latest commit from output: %s", out)
+	}
+	sha := fields[0]
+	log.Printf("Latest commit for %s: %s\n", info.mod, sha)
+
+	// Update the dependency to the latest commit
+	cmd := exec.Command("go", "get", fmt.Sprintf("%s@%s", info.mod, sha))
```
dagood: Consider `go get foo@main`, `foo@v2`, etc.?

dagood (approved):
LGTM as a dev tool. Some suggestions that could be integrated later.

> Note that the tool only hard-code information that don't change over time, like the git repo url and the Go module name, so it is safe to install it (i.e. run `go install eng/_util/cmd/upgrader/main.go`) and then use it like this `upgrader openssl windows darwin`.

I would temper this a bit... it would normally be safe, but if we *do* need to add or move something, it's better to have it in microsoft/go living alongside the code so that when something does go wrong, it's easy to re-install the version-specific change or go back to `go run`.

Otherwise, it would seem like this really belongs in go-infra, like the patch tooling.

> go run eng/_util/cmd/upgrader/main.go openssl windows darwin

I haven't been using full-path in this case--I've used it elsewhere and it's a great way to avoid the go.mod-in-another-directory issue, but the pwsh is in my muscle memory. Good tip. 😄 

(And on that note, the pwsh run.ps1 seems fairly overcomplicated and I believe it could be simplified to `go run` too.)

dagood:
> I haven't been using full-path in this case--I've used it elsewhere and it's a great way to avoid the go.mod-in-another-directory issue, but the pwsh is in my muscle memory. Good tip.

Oh... this only works when it doesn't need to get any packages from a go.mod, like here.

(This led me to learn that Go now has a `-C` option like Git's, for `go install -C eng/_util ./cmd/foo`. Neat, but it doesn't work in this case: the working directory is actually changed, so the tools in _util won't work. Also, that means tab-completion of the tool package name doesn't work, so there's less of a point.)

eng/_util/cmd/upgrader/main.go
qmuntal: Good point 👍

eng/_util/cmd/upgrader/main.go
qmuntal: The sha should be enough now that ls-remote accepts a reference rather than hardcoding `HEAD`.

eng/_util/cmd/upgrader/main.go
qmuntal: I'll use `runtime.GOROOT`, so that we are sure we use the binary that has run the tool. It also works well if the tool is installed.

PR #1984, opened 2025-12-05
implement CSHAKE

gdams:
The CNG and OpenSSL backends have support for CSHAKE so we should allow people to use it


qmuntal (approved):
LGTM, I'm only missing an update to https://github.com/microsoft/go/blob/microsoft/main/eng/doc/CrossPlatformCryptography.md.

PR #1996, opened 2025-12-10
auto generate crypto platform doc and programatically store status

gdams:
By programtically defining the crypto algorithm support we can better test for regressions but also auto generate the CrossPlatformCryptography.md doc which is already getting complex with all the footnotes etc


qmuntal (commented):
Looks good! One meta-comment: for the `platforms` property, could we make it default to "supported" on all platforms so we only have to type those platforms that are not supported or that require a note?

eng/_util/cmd/gencryptodoc/crypto.json
```diff
@@ -0,0 +1,2214 @@
...
+              ]
+            },
+            "macos": {
+              "supported": "false"
+            }
+          }
+        },
+        {
+          "name": "SHA-3-224",
+          "notes": ["Available starting in the Microsoft build of Go 1.26."],
```
qmuntal: This comment (with different versions) appears many times, we could codify it by adding a "goVersion" property and then letting the Go generator map it to a proper note.

Same applies to other version requirements, e.g. OpenSSL, MacOS, Windows, ...

gdams: I've updated the macos, min go and openssl one. The Windows footnote is a bit more awkward so I've left it for now unless you think it's still worth doing

qmuntal (approved):

gdams:
I'm going to hit merge, we can always iterate on the schema in follow-up PRs

PR #2003, opened 2025-12-12
FIPS readme updates, clarify GOFIPS

### LINKED ISSUE #2000: Clarify the support of the GOFIPS environment variable in documentation
[eng/doc/fips/README.md](https://github.com/microsoft/go/blob/microsoft/main/eng/doc/fips/README.md#usage-runtime) mentions that

> Support for the GOFIPS environment variable will be removed in Go 1.25.

But it actually seems that `GOFIPS=1` is still supported in Go 1.25 and still has an impact. Only GOFIPS=0 has been dropped by https://github.com/microsoft/go/pull/1708.

dagood:
* Fix https://github.com/microsoft/go/issues/2000
* Make new docs more accessible from this entrypoint.
* Fix other inconsistencies with Go 1.25.
* Clean up some "pre-1.24" mentions (when they don't seem worth the noise for the historical note).
* Clarify FIPS mode preference behavior, unify the part of the process that is the same on all platforms.
  * I tried to be careful to document what the code does much more closely. Maybe it's too detailed, but there's some nuance in the code that's hard to describe, like `GOFIPS=asdf` meaning that `systemFIPSMode()` won't be called. It's also possible I misinterpreted it or we'll want to change it--please take a look at this, @qmuntal.

https://github.com/microsoft/go/blob/microsoft/release-branch.go1.25/patches/0003-Implement-crypto-internal-backend.patch#L1748

https://github.com/microsoft/go/blob/microsoft/release-branch.go1.24/patches/0003-Implement-crypto-internal-backend.patch#L1750

### DIFF
--- a/eng/doc/fips/README.md
+++ b/eng/doc/fips/README.md
@@ -1,7 +1,13 @@
-This directory contains documentation about FIPS and the FIPS implementation in the Microsoft fork of Go.
+This directory contains documentation about FIPS, details about the FIPS implementation in the Microsoft build of Go, and using system-provided cryptography via the Go standard library.
 
-* README.md (this file): a general overview and first steps.
+* README.md (this file): an overview of the design and how to use it.
 * [**FIPS 140 User Guide** (UserGuide.md)](UserGuide.md): notes on FIPS compliance of specific crypto APIs.
+
+See also:
+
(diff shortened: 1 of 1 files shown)


karianna (approved):
Looks fine overall, looking forward to a day where we can archive the variations!

qmuntal (approved):

Issue #1995, opened 2025-12-10
Align TLS settings with Microsoft policies


qmuntal:
We should be following this policy:

[Encrypt Data in Transit Using Approved Cryptographic Protocols and Algorithms - Liquid](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fliquid.microsoft.com%2FWeb%2FObject%2FRead%2FMS.Security%2FRequirements%2FMicrosoft.Security.Cryptography.10031%23Cipher%2520Suites&data=05%7C02%7Cjheglund%40microsoft.com%7Ca333486ed4524222659908de2b65462d%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C638995911122956599%7CUnknown%7CTWFpbGZsb3d8eyJFbXB0eU1hcGkiOnRydWUsIlYiOiIwLjAuMDAwMCIsIlAiOiJXaW4zMiIsIkFOIjoiTWFpbCIsIldUIjoyfQ%3D%3D%7C0%7C%7C%7C&sdata=seP5ONHh%2BAYx%2FnwcHHdcUsFOsI4xlfcEzelvwYsIIVI%3D&reserved=0)

PR #2027, opened 2026-01-07
Prefer AES GCM over CHACHA even when hardware AES is not supported

qmuntal:
`CHACHA20_POLY1305_SHA256` is not allowed by the Microsoft crypto policies, but we are using it when the hardware doesn't support AES instruction. All hardware build after 2010 should have those instructions, but disable the preferred CHACHA cipher suite paths just in case.

This change can be reverted by setting `GODEBUG=ms_tlsprofile=off`.


gdams (approved):

qmuntal:
For #1995.

PR #2041, opened 2026-01-13
Fallback to Go crypto when RSA key size is not supported

qmuntal:
Go RSA implementation only restricts the minimum RSA key size to 1024 bits (and that can be bypassed by setting (`GODEBUG=rsa1024min=0`). It doesn't limit its maximum value and allow sizes not divisible by 8.

All out backends have a much stricter RSA key size policy. Take them into account when checking if we have to fallback to Go.

This fixes several compatibility issues with a new FIPS provider being developed in AZL3 and allows to unskip some additional tests.

### DIFF
--- a/eng/doc/CrossPlatformCryptography.md
+++ b/eng/doc/CrossPlatformCryptography.md
@@ -146,8 +146,6 @@ This section includes the following packages:
 
 - [crypto/rsa](https://pkg.go.dev/crypto/rsa)
 
-[rsa.GenerateKey](https://pkg.go.dev/crypto/rsa#GenerateKey) only supports the following key sizes (in bits): 2048, 3072, 4096.
-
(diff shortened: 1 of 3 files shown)


gdams (approved):

eng/doc/CrossPlatformCryptography.md
```diff
@@ -146,8 +146,6 @@ This section includes the following packages:
 
 - [crypto/rsa](https://pkg.go.dev/crypto/rsa)
 
-[rsa.GenerateKey](https://pkg.go.dev/crypto/rsa#GenerateKey) only supports the following key sizes (in bits): 2048, 3072, 4096.
```
dagood, line 149: Was this just a mistake in the doc? Or is this something that was fixed, and is still true for some supported versions?

PR #2051, opened 2026-01-15
SkipTestKeysFromPreMasterSecret when running on FIPS mode

qmuntal:
Some OpenSSL providers don't support TLS 1.0 PRF in FIPS mode. The public `crypto/tls` API doesn't support it either, but these tests use internal functions.

Found while testing the new AZL3 FIPS provider.

Updates #266.


gdams (approved):

dagood:
> Updates #266.

I think that was meant to be:

* https://github.com/golang-fips/openssl/issues/266

Issue #172, opened 2021-08-13
Add windows-arm64 builders


dagood:
See https://github.com/microsoft/go/issues/173 for initial work.

PR #2053, opened 2026-01-15
CI: add build support for Windows arm64

gdams:
I can't find anything that confirms that there are 1ES Windows arm64 runner images so cross compiling for now

* https://github.com/microsoft/go/issues/172


dagood:
Have you run a test internal build?

gdams:
> Have you run a test internal build?

no I haven't, I assumed it would just work as it's running on the same AMD64 VMs?

dagood:
It's probably fine based on public CI, but it would be best to check that it doesn't hit problems with build manifest, signing, pdb gen.

dagood:
Let's just go ahead and merge this--the change itself is a simple configuration change that *should* be valid, and any fallout we can treat as its own build break. (Testing in the real build is more accurate anyway!)

dagood (approved):

PR #2063, opened 2026-01-16
CI: enable native macos arm64 testing

gdams:
macos 15 arm64 runners exist in CI now so we should be running tests with them. Also switched the build from cross compile to native compilation.


qmuntal (approved):

dagood:
> Also switched the build from cross compile to native compilation.

This needs its own justification--it looks like this job finished in half the time (probably mostly just larger agents?) so this is good!

gdams:
> > Also switched the build from cross compile to native compilation.
> 
> 
> 
> This needs its own justification--it looks like this job finished in half the time (probably mostly just larger agents?) so this is good!

Yeah this is great! Apple silicon chips are so much faster for compiling! Might be worth switching the intel build to cross compiling on arm64 in the near future.

dagood:
Well, devil's advocate: these could happen to be better machines in many ways other than the CPU architecture. 😄 But yeah, I think that seeking out the fastest agent to reserve+build on for every buildandpack job would be interesting to see.

PR #2078, opened 2026-01-22
[release-branch.go1.26] Update release notes for 1.26

dagood:


qmuntal (commented):
Added the missing WIP release notes in https://github.com/microsoft/go/pull/2096.

dagood:
Updated based on https://github.com/microsoft/go/pull/2096

qmuntal (approved):

docs/go1.26.md
```diff
@@ -22,14 +27,15 @@ Setting the FIPS preference to enabled will no longer cause a panic when the Win
...
+
 Binaries can now be built without using cgo by setting `GOEXPERIMENT=ms_nocgo_opensslcrypto`.
 
 #### Darwin
 
 The backend is no longer in preview and is now fully supported.
 It is enabled by default on macOS.
 
-Binaries can now be built without using cgo.
+Binaries are now built without using cgo.
```
dagood: Hmm, my intent here was to clarify that without something like `GOEXPERIMENT=ms_nocgo_opensslcrypto`, it always happens this way, but after reading it again, it could be taken to mean that even if you use cgo for some other reason, it will somehow be built without cgo (illogical, which could raise a question). More specific:

```suggestion
Building a program that uses this backend no longer requires cgo.
```

PR #2082, opened 2026-01-23
Add footnotes details elements for optional clickability

dagood:
Some footnotes can be a bit tricky to associate with the list at the end of the table, so try adding a `details` element to make them clickable:

> image

This should have no content change, just formatting.

### DIFF
--- a/eng/_util/cmd/updatecryptodocs/updatecryptodocs.go
+++ b/eng/_util/cmd/updatecryptodocs/updatecryptodocs.go
@@ -197,10 +197,11 @@ func printTable(w io.Writer, section Section) {
 			strs[i] = fmt.Sprintf("%d", idx)
 		}
 		var r strings.Builder
-		fmt.Fprintf(&r,
-			"<details><summary>%s<sup>%s</sup></summary>",
-			name, strings.Join(strs, ","))
-		for _, idx := range indices {
+		fmt.Fprintf(&r, "%s<sup title=\"", name)
+		for i, idx := range indices {
+			if i > 0 {
+				fmt.Fprintf(&r, " ")
+			}
 			note := ""
 			// Find note by index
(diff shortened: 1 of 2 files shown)


karianna (approved):

qmuntal (changes requested):
The footnotes are hard to follow, that's a real issue. On the other hand, I'm don't quite like the solution proposed here:

- Each table entry with a footnote now will have an arrow and a superscript number, which feels like too noisy.
- It is not directly evident that clicking the arrow will expand the entry, nor that the expanded content is the same as the footnote down below.
- The table layout changes when the entry is expanded, and expanding too many entries makes the table unreadable
- The markdown source code is uglier (nit)

I would prefer something like native GitHub Markdown footnotes instead, see https://github.blog/changelog/2021-09-30-footnotes-now-supported-in-markdown-fields/.

dagood:
> I would prefer something like native GitHub Markdown footnotes instead, see https://github.blog/changelog/2021-09-30-footnotes-now-supported-in-markdown-fields/.

Same, but they're all rendered way down at the end of the file. (And since this means everything piles up, the "back to source" buttons probably end up being overwhelming visual clutter too.) I also don't think this helps when there are multiple footnotes per cell.

I'll try `<sup title="foo">1,2</sup>` for mouse hover to see how that looks. Not copyable, depends on browser, no formatting and newlines, and doesn't improve raw markdown formatting, but less intrusive when rendered.

dagood:
Hover text would need more processing to strip links:

> image

At its best, looks like this:

> image

Abandoning this for now.

dagood:
Then next approach I'd try is to have every row link to its section in https://github.com/microsoft/go/blob/microsoft/main/eng/doc/fips/UserGuide.md, where each requirement is also laid out in full text.

Issue #146, opened 2021-07-29
Let devs install microsoft/go similarly to `go install golang.org/dl/go1.16.6`


dagood:
See https://golang.org/doc/manage-install#installing-multiple. (A partner team pointed it out to us as a way they get Go. In particular, to compare how different Go versions go when something strange is going on.)

https://github.com/golang/dl is a mirror of the `golang.org/dl` module's repo. It has a package for each version of Go: a small wrapper that can download that specific version to e.g. `~/sdk/go1.16.6`, and after that, it redirects `go1.16.6 [...]` commands to that Go binary it downloaded.

The simplest-seeming thing to do would be to copy `dl` and provide a `https://github.com/microsoft/go-dl` module with e.g. a `microsoft-go1.16.6` package that points at our blob storage instead of the tar.gz/zip files available on `dl.google.com/go/`.

dagood:
It occurred to me that if we publish prebuilt copies of these commands, they could be considered "installers", and because they're written in Go, we would be more familiar with how to maintain them than if we create MSI (https://github.com/microsoft/go/issues/29) or macOS pkg (https://github.com/microsoft/go/issues/1325), RPM (https://github.com/microsoft/go/issues/1318), or Deb (https://github.com/microsoft/go/issues/9) installers.

Publishing a prebuilt binary fixes the bootstrap issue: to `go install`, you need `go`, but you don't need Go to run a prebuilt Go app.

For context: upstream doesn't publish prebuilt copies of `golang.org/dl`. But they already support MSI and macOS pkg installers.

dagood:
A prototype version of this has been implemented internally (https://github.com/microsoft/go-lab/pull/142, thanks!) but it isn't done. A minimum Done for this issue IMO means:

* Available publicly.
* Automatically updates to include the latest version.

(Note that this is a public issue. We might want to create multiple internal issues to track parts of it if necessary.)

PR #2086, opened 2026-01-27
Update Additional Features draft doc, include version features

dagood:
* https://github.com/microsoft/go/pull/1899
* https://github.com/microsoft/go/issues/262
* https://github.com/microsoft/go/pull/2078 can condense a lot with a ref to this doc.

Simplify, focus on features without exhaustively explaining enable/disable, and fill out the list. Include a specific note about security (regarding our crypto changes) to have another chance to clear up any misunderstanding about that.

### DIFF
--- a/eng/doc/AdditionalFeatures.md
+++ b/eng/doc/AdditionalFeatures.md
@@ -1,63 +1,161 @@
-# Microsoft Build of Go - Additional Features
+# Additional Features Provided by the Microsoft build of Go
 
-This document provides a comprehensive overview of all Microsoft-specific features currently supported in the Microsoft build of Go.
+This document provides a comprehensive overview of all features currently implemented in the Microsoft build of Go that are not part of the official Go distribution.
 
-## Crypto Backend Features
+If you are trying to use the Microsoft build of Go but aren't looking for a precise guide to every feature, see [the Migration Guide](/eng/doc/MigrationGuide.md).
 
-These features enable system-provided cryptographic libraries for FIPS 140 compliance and improved security.
+The `patches` directory at each Git tag specifies the exact code changes we have made to the official Go toolchain of that version.
+If it's critical to you to understand the exact implementation details of all changes we've made, please review the patch files.
 
-### `GOEXPERIMENT=systemcrypto`
-- **Purpose**: Automatically selects the recommended crypto backend for the target platform
-- **Behavior**: 
-  - Linux: Enables `opensslcrypto` (OpenSSL)
-  - Windows: Enables `cngcrypto` (CNG - Cryptography Next Generation)
-  - macOS: Enables `darwincrypto` (CommonCrypto & CryptoKit) (Go 1.24+)
-- **Default**: Enabled by default on Linux and Windows since Go 1.25
-- **Notes**: Acts as an alias that also satisfies individual backend build constraints, set `MS_GO_NOSYSTEMCRYPTO: 1` if backend opt-out needed
+## Telemetry
 
-## TLS Features
+The Microsoft build of Go collects [opt-out (enabled by default) toolset telemetry](https://devblogs.microsoft.com/go/microsoft-go-telemetry/).
 
-### `GOEXPERIMENT=ms_tls_config_schannel`
-- **Purpose**: Filters and orders cipher suites according to Windows Schannel settings
-- **Platform**: Windows only (no effect on Windows Server 2012/Windows 8 and earlier)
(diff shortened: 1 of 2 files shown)


eng/doc/AdditionalFeatures.md
```diff
@@ -0,0 +1,161 @@
...
+Other methods of changing goexperiment settings also work.
+
+See [the Migration Guide](/eng/doc/MigrationGuide.md) for more information about when, and how, to configure this feature.
+
+More details are available in the [FIPS README](/eng/doc/fips/README.md).
+These changes also include improvements to crypto library behavior to make it easier to comply with some FIPS requirements.
+
+> [!NOTE]
+> Improving security is not the purpose of this feature.
+> We encourage Go developers who aren't subject to Microsoft internal cryptography policies to continue using the official Go cryptography implementation.
```
qmuntal:
```suggestion
> The purpose of this feature is to comply with the Microsoft internal cryptography policies, which mandate using system-provided cryptography modules.
> We encourage Go developers who aren't subject to those policies to continue using the official Go cryptography implementation.
```

eng/doc/AdditionalFeatures.md
```diff
@@ -0,0 +1,161 @@
...
+
+The Microsoft build of Go disables the `GOTOOLCHAIN` feature by default to ensure consistent builds and toolchain usage.
+Specifically, the default `GOTOOLCHAIN` value is `local` rather than `auto`.
+
+See <https://go.dev/doc/toolchain> for more information about `GOTOOLCHAIN`.
+
+This change [is recommended by the Go team](https://github.com/golang/go/issues/57001#issuecomment-1332657428) for Linux distributions of Go, and the reasoning applies to the Microsoft build of Go as well.
+We anticipate that otherwise, many who use the Microsoft build of Go would find themselves in a situation where their build switched to the official Go toolchain without their knowledge, and they would silently lose features they rely on for compliance.
+
+For developers who are aware of this behavior and want it, it's possible to re-enable `GOTOOLCHAIN=auto`.
```
qmuntal: We also don't allow setting `GOTOOLCHAIN=auto`, as users might cargo-cult it, and downgrading to upstream Go is a "catastrophic" event. To re-enable it one should set the following vars `MS_GOTOOLCHAIN_ALLOW_NON_LOCAL=1 GOTOOLCHAIN=auto`.

qmuntal (changes requested):
Also, add a TODO section for the TLS settings aligned with MS policies. That is orthogonal to the system-provided crypto. I'll fill it in a follow-up PR.

eng/doc/AdditionalFeatures.md
dagood: Yeah, the harsh start is my reaction to seeing "for security purposes" slip in when (e.g.) AI takes the wheel. Recommending that people keep using Go crypto *should* be enough.

eng/doc/AdditionalFeatures.md
dagood: Forgot about that safeguard, I think it actually makes sense to describe it here as a feature in itself.

dagood:
I added some content for the TLS alignment feature based on the patch's godebug doc, could still probably use filling in (potentially a ref to more details in FIPS doc).

qmuntal (approved):
LGTM++!

karianna:
Maybe not for this round but something kept bugging me and then I realised what it was.  Should we consider having the versions tabbed, or in a dropdown like I see on other major public fdocs pages for build tools like Gradle or versions of Java?  I might be very wrong abotu the Ux of that given what you're trying to convey here sometimes but thought I'd mention it,

gdams (approved):
LGTM, I agree with Martijn that we might be able to present this information in a better way but I like to move fast and iterate with docs

dagood:
> Should we consider having the versions tabbed, or in a dropdown like I see on other major public fdocs pages for build tools like Gradle or versions of Java?

Unfortunately, we don't have these kinds of tools in Markdown, the best we have is `<details><summary>…</summary>…</details>` which isn't all that ergonomic. But: what I expect is that when a version goes EOL, we remove references to that version in docs like this to keep it readable.

An idea @gdams brought up offline (or something similar that led to this? 😄) is to maybe get this info accessible via `go doc` so you can easily tell what your specific version will do without cross-referencing these docs vs. your version. (An obstacle for that is that it's annoying to acquire the Microsoft build of Go on a dev machine, which https://github.com/microsoft/go/issues/146 (dl) would help with.)

PR #2106, opened 2026-02-04
openssl: delete cgo-less experiment and promote to default

gdams:
This is for **go 1.27** and onwards


eng/pipeline/stages/go-builder-matrix-stages.yml
```diff
@@ -93,13 +93,13 @@ stages:
...
+          - { experiment: 'systemcrypto', os: linux, arch: amd64, config: nocgo, fips: true }
           - { experiment: systemcrypto, os: linux, arch: amd64, config: test, distro: ubuntu }
-          - { experiment: 'systemcrypto,ms_nocgo_opensslcrypto', os: linux, arch: amd64, config: nocgo, distro: ubuntu }
+          - { experiment: 'systemcrypto', os: linux, arch: amd64, config: nocgo, distro: ubuntu }
           - { experiment: systemcrypto, os: linux, arch: amd64, config: test, distro: azurelinux3 }
-          - { experiment: 'systemcrypto,ms_nocgo_opensslcrypto', os: linux, arch: amd64, config: nocgo, distro: azurelinux3 }
+          - { experiment: 'systemcrypto', os: linux, arch: amd64, config: nocgo, distro: azurelinux3 }
           - { experiment: systemcrypto, os: linux, arch: amd64, config: test, distro: azurelinux3, fips: true }
-          - { experiment: 'systemcrypto,ms_nocgo_opensslcrypto', os: linux, arch: amd64, config: nocgo, distro: azurelinux3, fips: true }
+          - { experiment: 'systemcrypto', os: linux, arch: amd64, config: nocgo, distro: azurelinux3, fips: true }
```
qmuntal: single quotes not needed

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -287,16 +286,11 @@ index 48c3aa5efdb008..3ac500b85e6988 100644
  		t.registerTest("GOOS=ios on darwin/amd64",
  			&goTest{
  				variant:  "amd64ios",
-@@ -882,14 +886,21 @@ func (t *tester) registerTests() {
+@@ -882,14 +886,16 @@ func (t *tester) registerTests() {
  
  	// Test internal linking of PIE binaries where it is supported.
  	if t.internalLinkPIE() && !disablePIE {
 +		var tags, env []string
```
qmuntal: tags and env can be removed and the changed below reverted.

qmuntal (changes requested):

patches/0003-Implement-crypto-internal-backend.patch
gdams: done, sorry pushed after requesting your review

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -539,12 +503,10 @@ index 00000000000000..b6ba8158598a6c
...
-+		{"linux", "amd64", "noms_nocgo_opensslcrypto", false}, // cgoless OpenSSL requires ms_nocgo_opensslcrypto on linux/amd64
-+		{"linux", "arm64", "noms_nocgo_opensslcrypto", false}, // cgoless OpenSSL requires ms_nocgo_opensslcrypto on linux/arm64
-+		{"linux", "ppc64le", "ms_nocgo_cngcrypto", false},     // cgoless OpenSSL not supported on linux/ppc64le
-+		{"linux", "arm", "ms_nocgo_cngcrypto", false},         // cgoless OpenSSL not supported on linux/arm
-+		{"linux", "amd64", "ms_nocgo_opensslcrypto", true},
-+		{"linux", "arm64", "ms_nocgo_opensslcrypto", true},
++		{"linux", "amd64", "", false},
++		{"linux", "arm64", "", false},
++		{"linux", "ppc64le", "", false},
++		{"linux", "arm", "", false},
```
qmuntal: All these should succed.

qmuntal (changes requested):

patches/0003-Implement-crypto-internal-backend.patch
gdams: ooops fixed :D

gdams:
> The telemetry configuration in patch 0001 to remove the ms_nocgo_opensslcrypto reference

This is part of the go-infra package so we can't remove this here. I also don't think it makes sense to remove it from go-infra for now as we need it for Go 1.26.

> The MigrationGuide.md to reflect that CGO-less OpenSSL is now the default on supported architectures and no longer requires explicit enablement

I've tried to get the balance right in the migration guide, considering that most users will be running Go 1.26 so mentioning the experiment still makes sense.

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -3163,25 +3119,26 @@ index 00000000000000..5e4b436554d44d
...
++//go:build goexperiment.opensslcrypto && !(cgo || 386 || amd64 || arm || arm64 || ppc64le || riscv64)
 +
 +package crypto
 +
 +func init() {
 +	`
 +	Using GOEXPERIMENT=systemcrypto on Linux requires CGO_ENABLED=1.
 +
-+	Consider using our cgo-less experiment by setting GOEXPERIMENT=ms_nocgo_opensslcrypto.
++	Our CGO-less OpenSSL implementation is only supported on certain architectures.
```
```suggestion
+	Our cgo-less OpenSSL implementation is only supported on certain architectures.
```

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -3163,25 +3119,26 @@ index 00000000000000..5e4b436554d44d
...
++	These architectures are: 386, amd64, arm, arm64, ppc64le and riscv64.
 +	
-+	For more information, visit https://github.com/microsoft/go/blob/microsoft/main/eng/doc/MigrationGuide.md#cgo-is-not-enabled
++	To request support for additional architectures, please file an issue at https://github.com/microsoft/go/issues/new.
```
dagood: I think we should point at a doc, and the doc should point at a label search where people can see existing requests. The doc can also explain why there is a per-platform issue.

eng/doc/MigrationGuide.md
```diff
@@ -146,7 +146,7 @@ For more information, visit https://github.com/microsoft/go/blob/microsoft/main/
 ```
 
 > [!NOTE]
-> As of Go 1.26, there is a cgo-less experiment available for Linux: `ms_nocgo_opensslcrypto`.
+> In Go 1.26, there is a cgo-less experiment available for Linux: `ms_nocgo_opensslcrypto`. For Go 1.27 this experiment will be removed and the CGO requirement for `systemcrypto` on Linux will be lifted.
```
dagood:
```suggestion
> In Go 1.26, there is a cgo-less experiment available for Linux: `ms_nocgo_opensslcrypto`. For Go 1.27 this experiment will be removed and the cgo requirement for `systemcrypto` on Linux will be lifted by default.
```

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -3163,25 +3119,26 @@ index 00000000000000..5e4b436554d44d
...
 --- /dev/null
 +++ b/src/crypto/systemcrypto_nocgo_linux.go
-@@ -0,0 +1,17 @@
+@@ -0,0 +1,18 @@
 +// Copyright 2025 The Go Authors. All rights reserved.
 +// Use of this source code is governed by a BSD-style
 +// license that can be found in the LICENSE file.
 +
-+//go:build goexperiment.opensslcrypto && !(cgo || (goexperiment.ms_nocgo_opensslcrypto && (386 || amd64 || arm || arm64 || ppc64le || riscv64)))
++//go:build goexperiment.opensslcrypto && !(cgo || 386 || amd64 || arm || arm64 || ppc64le || riscv64)
```
dagood, line 3130: It's not clear to me on an initial read, but based on this, is the plan to use cgo if cgo happens to be enabled for some other reason, and use cgoless if cgo isn't enabled globally? What if someone wants to use cgo e.g. because they're targeting a different platform? Or if they don't want to use cgo for the backend, but do want to use it for their own C library they're using?

What are the current limitations of the cgoless implementation? Are there performance differences, functional differences (even if slight)?

Worth an ADR to explore?

dagood (changes requested):

dagood:
> > The telemetry configuration in patch 0001 to remove the ms_nocgo_opensslcrypto reference
> 
> This is part of the go-infra package so we can't remove this here. I also don't think it makes sense to remove it from go-infra for now as we need it for Go 1.26.

I think the agent needs to be taught that it shouldn't pay attention to vendored Go code in general.

qmuntal:
> I think the agent needs to be taught that it shouldn't pay attention to vendored Go code in general.

The comment from the agent was useful to me, it surfaced the need to update the telemetry configuration at some point in the future. The recommendation is wrong, that's true, we don't want to do it now.

patches/0003-Implement-crypto-internal-backend.patch
qmuntal: > is the plan to use cgo if cgo happens to be enabled for some other reason, and use cgoless if cgo isn't enabled globally

Yes. This fact needs to be better documented, probably as part of this PR @gdams.

> What if someone wants to use cgo e.g. because they're targeting a different platform? Or if they don't want to use cgo for the backend, but do want to use it for their own C library they're using?

Mixing cgo with cgoless OpenSSL is possible, but I don't see why someone would want to do that, and that means more testing and complexity on our side.

> What are the current limitations of the cgoless implementation? Are there performance differences, functional differences (even if slight)?

This also needs to be documented. There are some few limitations and differences that I want to smooth out before Go 1.27.

> Worth an ADR to explore?

I should have made an ADR for all this enchilada one year ago. IMO it is a bit late now, we should focus instead on writing good docs.

gdams: I've updated the docs to explain this slighlt more clearly

qmuntal (approved):

patches/0003-Implement-crypto-internal-backend.patch
dagood: We talked about this in sync, and a feature doc sounds good. (IMO it can serve just fine as the de-facto Record of our Architecture Decision. 🙂) We also talked about potentially adding a dedicated roadmap somewhere, which could take pressure off this cgoless doc potentially having to explain the whole rollout plan.

eng/doc/MigrationGuide.md
```diff
@@ -146,9 +146,17 @@ For more information, visit https://github.com/microsoft/go/blob/microsoft/main/
 ```
 
 > [!NOTE]
-> As of Go 1.26, there is a cgo-less experiment available for Linux: `ms_nocgo_opensslcrypto`.
+> In Go 1.26, there is a cgo-less experiment available for Linux: `ms_nocgo_opensslcrypto`. For Go 1.27 this experiment will be removed and the cgo requirement for `systemcrypto` on Linux will be lifted by default.
```
dagood: Split the line for reading/scanning clarity--put the version noticeably at left edge.

```suggestion
> In Go 1.26, there is a cgo-less experiment available for Linux: `ms_nocgo_opensslcrypto`.
>
> In Go 1.27, the experiment will be removed and the cgo requirement for `systemcrypto` on Linux will be lifted by default.
>
```

eng/doc/MigrationGuide.md
```diff
@@ -146,9 +146,17 @@ For more information, visit https://github.com/microsoft/go/blob/microsoft/main/
...
 > This allows the use of OpenSSL without requiring cgo.
-> Currently this experiment is supported on the following architectures: `386`, `amd64`, `arm`, `arm64`, `ppc64le` and `riscv64`.
+> Currently this experiment is supported on the following architectures: **386**, **amd64**, **arm**, **arm64**, **ppc64le** and **riscv64**.
+>
+> **How the backend is selected:** If cgo is enabled (e.g., `CGO_ENABLED=1`), the cgo-based OpenSSL backend is always used.
```
dagood: This should be in its own doc IMO and linked from here as a "for more details". As long as this selection is done properly and there are no feature/perf gaps, it seems to me that it will "just work" and the vast majority reading this doc don't need the details.

eng/doc/MigrationGuide.md
```diff
@@ -146,9 +146,17 @@ For more information, visit https://github.com/microsoft/go/blob/microsoft/main/
...
+> The cgo-less backend is only used when cgo is disabled AND you're on a supported architecture.
```
dagood: I don't think we use caps for emphasis normally?

```suggestion
> The cgo-less backend is only used when cgo is disabled **and** you're on a supported architecture.
```

eng/doc/MigrationGuide.md
```diff
@@ -146,9 +146,17 @@ For more information, visit https://github.com/microsoft/go/blob/microsoft/main/
...
+> This means if you enable cgo for other reasons (e.g., your own C library or cross-compiling for a different platform), the crypto backend will use the cgo-based implementation.
```
dagood: Why would you enable cgo for cross-compiling for a different platform?

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -2717,15 +2673,15 @@ index 00000000000000..c8f281d7fd1274
...
+index 00000000000000..a7437047f91696
 --- /dev/null
 +++ b/src/crypto/internal/backend/openssl_linux.go
 @@ -0,0 +1,424 @@
 +// Copyright 2017 The Go Authors. All rights reserved.
 +// Use of this source code is governed by a BSD-style
 +// license that can be found in the LICENSE file.
 +
-+//go:build goexperiment.opensslcrypto && linux && (cgo || goexperiment.ms_nocgo_opensslcrypto)
++//go:build goexperiment.opensslcrypto
```
dagood: I might be forgetting a convention, but isn't it bad/unexpected to have a file named `openssl_linux.go` that doesn't have a `linux` build tag? (And does this really imply that we have Windows/macOS/etc. support here?)

patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -3163,25 +3119,26 @@ index 00000000000000..5e4b436554d44d
...
 +func init() {
 +	`
 +	Using GOEXPERIMENT=systemcrypto on Linux requires CGO_ENABLED=1.
 +
-+	Consider using our cgo-less experiment by setting GOEXPERIMENT=ms_nocgo_opensslcrypto.
++	Our cgo-less OpenSSL implementation is only supported on certain architectures.
++	These architectures are: 386, amd64, arm, arm64, ppc64le and riscv64.
 +	
-+	For more information, visit https://github.com/microsoft/go/blob/microsoft/main/eng/doc/MigrationGuide.md#cgo-is-not-enabled
++	For more information, including how to request support for additional architectures, visit https://github.com/microsoft/go/blob/main/eng/doc/MigrationGuide.md#cgo-is-not-enabled.
```
dagood: Leave out `.`. Some shells' URL parsers could probably handle it to make it clickable without breaking the URL, but not all. It might also be manually copy-pasted by selection, and it's easier to do that when there aren't adjacent garbage characters.

Also, branch is wrong.

```suggestion
+	For more information, including how to request support for additional architectures, visit https://github.com/microsoft/go/blob/microsoft/main/eng/doc/MigrationGuide.md#cgo-is-not-enabled
```

dagood (changes requested):

eng/doc/MigrationGuide.md
gdams: agreed, I've restrutured this

eng/doc/MigrationGuide.md
gdams: cross-compiling to an unsupported architecture right? I've expanded the wording slightly

patches/0003-Implement-crypto-internal-backend.patch
gdams: my understanding was that the filename was a build constraint itself? I do note however that we're inconsistent in our patches with this so we should probably agree on a standard and update all the patches.

gdams: @qmuntal confirmed to me that he prefered explicitly adding the `linux` tag so I've made the change

dagood: True, the [doc](https://pkg.go.dev/cmd/go#hdr-Build_constraints) says:

> If a file's name, after stripping the extension and a possible _test suffix, matches any of the following patterns: [...](example: source_windows_amd64.go) where GOOS and GOARCH represent any known operating system and architecture values respectively, then the file is considered to have an implicit build constraint requiring those terms (**in addition to** any explicit constraints in the file).

To me it might make more sense to take advantage and minimize a bit. Really just so the behavior and expectations are clear at a glance and modifying the constraints/filenames have the most intuitive result. For example, purego does this:

* [trampolines_amd64.s](https://github.com/ebitengine/purego/blob/071d22a94b4bc442118b3b0927274dd5ae4d7551/internal/fakecgo/trampolines_amd64.s#L4) includes `//go:build !cgo && (darwin || linux)`

For example, in this PR, there would be no way to accidentally remove the `linux` constraint if there weren't a technically redundant `linux` constraint to begin with! (Alternatively, we create some tests to validate that we always have the constraint in both places.)

dagood: Opened:
* https://github.com/microsoft/go/issues/2122

eng/doc/MigrationGuide.md
```diff
@@ -146,11 +146,10 @@ For more information, visit https://github.com/microsoft/go/blob/microsoft/main/
 ```
 
 > [!NOTE]
-> As of Go 1.26, there is a cgo-less experiment available for Linux: `ms_nocgo_opensslcrypto`.
-> This allows the use of OpenSSL without requiring cgo.
-> Currently this experiment is supported on the following architectures: `386`, `amd64`, `arm`, `arm64`, `ppc64le` and `riscv64`.
+> In Go 1.26, there is a cgo-less experiment available for Linux: `ms_nocgo_opensslcrypto`.
+> In Go 1.27, the experiment will be removed and the cgo requirement for `systemcrypto` on Linux will be lifted by default.
```
dagood, line 150: These are still on the same line (rendered):

```suggestion
>
> In Go 1.27, the experiment will be removed and the cgo requirement for `systemcrypto` on Linux will be lifted by default.
```

Or, should we even mention it here rather than in the other doc? 1.27 doesn't affect anyone yet, so it might not make sense for a migration doc.

```suggestion
```

eng/doc/NocgoOpenSSL.md
```diff
@@ -0,0 +1,37 @@
+# Cgo-less OpenSSL Backend
+
+This document describes the cgo-less OpenSSL backend for the Microsoft build of Go, available as an experimental feature starting in Go 1.26.
```
dagood, line 3: This only really describes how to turn it on, not how it works (which people will naturally be curious/cautious of, so we should do what we can to point people in the right direction IMO). Should we have docs in the backend repo maybe? Also for the macOS nocgo backend?

eng/doc/NocgoOpenSSL.md
```diff
@@ -0,0 +1,37 @@
...
+## Supported Architectures
+
+Currently this experiment is supported on the following architectures:
+
+- **386**
+- **amd64**
+- **arm**
+- **arm64**
+- **ppc64le**
+- **riscv64**
```
dagood, line 22: Like https://go.dev/dl/, I think we should highlight the ones that we expect most people to use, and therefore the ones they're probably looking for. (Otherwise, why highlight anything?)


```suggestion
- 386
- **amd64**
- arm
- **arm64**
- ppc64le
- riscv64
```

eng/doc/NocgoOpenSSL.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Overview
+
+In Go 1.26, there is a cgo-less experiment available for Linux: `ms_nocgo_opensslcrypto`.
+
+In Go 1.27, the experiment will be removed and the cgo requirement for `systemcrypto` on Linux will be lifted by default.
+
+This allows the use of OpenSSL without requiring cgo.
+
+## Supported Architectures
```
dagood, line 13:
```suggestion
## Supported architectures
```

We generally aren't using title case for subheaders in the repo.

eng/doc/NocgoOpenSSL.md
```diff
@@ -0,0 +1,37 @@
...
+- **arm**
+- **arm64**
+- **ppc64le**
+- **riscv64**
+
+The set of supported architectures is limited because each architecture requires a unique assembly implementation to call OpenSSL.
+Architectures are added based on demand and available resources.
+To see existing requests or request support for additional architectures, search for issues with the [`Area-Nocgo`](https://github.com/microsoft/go/labels/Area-Nocgo) label.
+
+## How the Backend is Selected
```
dagood, line 28:
```suggestion
## How the backend is selected
```

eng/doc/NocgoOpenSSL.md
```diff
@@ -0,0 +1,37 @@
...
+
+If cgo is enabled (e.g., `CGO_ENABLED=1`), the cgo-based OpenSSL backend is always used.
+The cgo-less backend is only used when cgo is disabled **and** you're on a supported architecture.
+
+This means if you enable cgo for other reasons (e.g., linking your own C library or cross-compiling to an unsupported architecture), the crypto backend will use the cgo-based implementation.
```
dagood, line 33: > cross-compiling to an unsupported architecture

This still seems muddy to me... unsupported by whom? Why is cross-compilation significant vs. native compilation?

A related scenario that might be worth documenting is the idea that if you want to compile for an architecture not supported by the cgoless Linux backend, you can do so by enabling cgo. (Maybe it can replace this scenario.)

We could also mention that it might be enabled by default on your platform.

```suggestion
This means if cgo is enabled by default on your platform or you enable cgo for other reasons (e.g., linking your own C library), the crypto backend will use the cgo-based implementation.
You can also intentionally enable cgo to successfully build for a platform without cgo-less OpenSSL support.
```

eng/doc/NocgoOpenSSL.md
```diff
@@ -0,0 +1,37 @@
+# Cgo-less OpenSSL Backend
```
dagood, line 1: Fairly different from the filename? `CgolessOpenSSL.md` or:


```suggestion
# No-cgo OpenSSL Backend
```

eng/doc/NocgoOpenSSL.md
```diff
@@ -0,0 +1,37 @@
...
+
+This document describes the cgo-less OpenSSL backend for the Microsoft build of Go, available as an experimental feature starting in Go 1.26.
+
+## Overview
+
+In Go 1.26, there is a cgo-less experiment available for Linux: `ms_nocgo_opensslcrypto`.
+
+In Go 1.27, the experiment will be removed and the cgo requirement for `systemcrypto` on Linux will be lifted by default.
```
dagood, line 9:
```suggestion
In Go 1.26, there is a cgo-less experiment available for Linux: `ms_nocgo_opensslcrypto`.

> [!NOTE]
> While `systemcrypto` is a fully supported `GOEXPERIMENT` value (it is not "experimental"), `ms_nocgo_opensslcrypto` **is** experimental in Go 1.26 and may have limitations.

In Go 1.27, the experiment will be removed and the cgo requirement for `systemcrypto` on Linux will be lifted by default.
```

eng/doc/NocgoOpenSSL.md
```diff
@@ -0,0 +1,37 @@
...
+## How the Backend is Selected
+
+If cgo is enabled (e.g., `CGO_ENABLED=1`), the cgo-based OpenSSL backend is always used.
+The cgo-less backend is only used when cgo is disabled **and** you're on a supported architecture.
+
+This means if you enable cgo for other reasons (e.g., linking your own C library or cross-compiling to an unsupported architecture), the crypto backend will use the cgo-based implementation.
+
+## Experimental Status
+
+While `systemcrypto` is a fully supported `GOEXPERIMENT` value (it is not "experimental"), `ms_nocgo_opensslcrypto` **is** experimental in Go 1.26 and may have limitations.
```
dagood, line 37: I think this makes more sense as a note, earlier. I would expect a section like this (if we have it) to be more detailed about why we're releasing it as an experiment first, and what our criteria will be for making it mainstream and calling it fully supported.


```suggestion
```

eng/doc/MigrationGuide.md
```diff
@@ -146,11 +146,10 @@ For more information, visit https://github.com/microsoft/go/blob/microsoft/main/
...
 >
-> While `systemcrypto` is a fully supported `GOEXPERIMENT` value (it is not "experimental"), `ms_nocgo_opensslcrypto` **is** experimental as of Go 1.26 and may have limitations.
+> For more details about supported architectures and backend selection, see [cgo-less OpenSSL Backend](NocgoOpenSSL.md).
```
dagood, line 152: Keep it simple and open ended?

```suggestion
> For more details, see [cgo-less OpenSSL Backend](NocgoOpenSSL.md).
```

eng/doc/NocgoOpenSSL.md
```diff
@@ -0,0 +1,37 @@
...
+- **386**
+- **amd64**
+- **arm**
+- **arm64**
+- **ppc64le**
+- **riscv64**
+
+The set of supported architectures is limited because each architecture requires a unique assembly implementation to call OpenSSL.
+Architectures are added based on demand and available resources.
+To see existing requests or request support for additional architectures, search for issues with the [`Area-Nocgo`](https://github.com/microsoft/go/labels/Area-Nocgo) label.
```
dagood, line 26: Prettier:

```suggestion
To see existing requests or request support for additional architectures, use the [![](https://img.shields.io/github/labels/microsoft/go/Area-Nocgo)](https://github.com/microsoft/go/labels/Area-Nocgo) label.
```

> To see existing requests or request support for additional architectures, use the [](https://github.com/microsoft/go/labels/Area-Nocgo) label.

dagood (approved):
Suggestions, some maybe for followup, otherwise LGTM.

dagood:
Oh, automerge. I guess it's followup then. 😄

PR #2126, opened 2026-02-10
Consolidate crypto backend GOEXPERIMENTs into systemcrypto

gdams:
Remove the per-platform GOEXPERIMENTs (opensslcrypto, cngcrypto, darwincrypto) and use only systemcrypto with platform build constraints.

Before this change, there was lots of confusion around the platform-specific experiements:
  - Linux: GOEXPERIMENT=opensslcrypto
  - Windows: GOEXPERIMENT=cngcrypto
  - Darwin: GOEXPERIMENT=darwincrypto

Platform-specific code now uses build constraints like:
  //go:build goexperiment.systemcrypto && linux
  //go:build goexperiment.systemcrypto && windows
  //go:build goexperiment.systemcrypto && darwin

Runtime checks use:
  goexperiment.SystemCrypto && runtime.GOOS == "linux"

This simplifies the developer experience and makes cross-platform scripts and documentation easier to maintain.

Changes:
- Throws errors for OpenSSLCrypto, CNGCrypto, DarwinCrypto in goexperiment.Flags
- Update all build constraints to use systemcrypto + platform
- Update runtime checks to use SystemCrypto + runtime.GOOS
- Update test infrastructure to only check for systemcrypto


patches/0003-Implement-crypto-internal-backend.patch
```diff
@@ -3217,65 +3201,45 @@ index 87cd1f4de3246a..c8d7fb4cffba66 100644
...
+ 				f, val = f[2:], false
+ 			}
++			// Check for deprecated per-platform crypto experiments.
++			switch f {
++			case "opensslcrypto":
++				return nil, fmt.Errorf("GOEXPERIMENT opensslcrypto is deprecated, use GOEXPERIMENT=systemcrypto instead")
++			case "cngcrypto":
++				return nil, fmt.Errorf("GOEXPERIMENT cngcrypto is deprecated, use GOEXPERIMENT=systemcrypto instead")
++			case "darwincrypto":
++				return nil, fmt.Errorf("GOEXPERIMENT darwincrypto is deprecated, use GOEXPERIMENT=systemcrypto instead")
```
qmuntal: "deprecated" implies that it is still somehow supported. Also, GOEXPERIMENT=systemcrypto is enabled by default, so it is not necessary mentioning it.

I would say something like this

```go
switch f {
  case "opensslcrypto", "cngcrypto", "darwincrypto":
      switch {
         case goos == "darwin" || (goos == "windows" && goarch != "386") || (goos == "linux" && all the goarchs that default to systemcrypto...):
              return nil, fmt.Errorf("systemcrypto is enabled by default, GOEXPERIMENT %s is not necessary", f)
         default:
              return nil, fmt.Errorf("GOEXPERIMENT %s has been removed, use GOEXPERIMENT=systemcrypto instead")
      }
}
```

qmuntal (changes requested):
Docs need to be updated (including release notes, why not 🐱 )

qmuntal (approved):

Issue #2138, opened 2026-02-13
Go 1.26.0-1 doesn't include revision in microsoft_toolset_version


dagood:
In `mcr.microsoft.com/oss/go/microsoft/golang:1.26-azurelinux3.0`, I'm not seeing the revision:

```
# GODEBUG=ms_version=1 go version
go version go1.26.0-microsoft linux/amd64
```

```
# go version -m main
main: go1.26.0
        path    command-line-arguments
        build   microsoft_systemcrypto=1
        build   microsoft_toolset_version=go1.26.0-microsoft
[...]
# go version -m $(go env GOROOT)/bin/go
/usr/local/go/bin/go: go1.26.0
        path    cmd/go
        build   microsoft_toolset_version=go1.26.0-microsoft
[...]
```

dagood:
The MICROSOFT_REVISION file is packed up, but it isn't present when the build is taking place (so, not present when `src/internal/buildcfg/zbootstrap.go` is generated). I tested the release branch flow with a revision, but it's possible I only tested it with one of the earlier approaches to implement the feature.

https://github.com/microsoft/go/blob/e5b716ebaeb3d51740d8e3804fa2b410b7ef954f/eng/_util/cmd/build/build.go#L284-L285

I think this just needs to be moved earlier in the process so MICROSOFT_REVISION is present during the build.

dagood:
Confirmed working for `go version go1.26.1-1-microsoft linux/amd64`

PR #2140, opened 2026-02-13
Include revision in Microsoft-specific version string

dagood:
* https://github.com/microsoft/go/issues/2138

### DIFF
--- a/eng/_util/cmd/build/build.go
+++ b/eng/_util/cmd/build/build.go
@@ -143,33 +143,9 @@ func build(o *options) (err error) {
 		return err
 	}
 
-	// Next, make sure VERSION and MICROSOFT_REVISION files are in the Go root. We need them for
-	// different reasons.
-	//
-	// A VERSION file is necessary to make distpack run. The build runs fine without this file,
-	// but for consistency with MICROSOFT_REVISION, we include VERSION during the build as well.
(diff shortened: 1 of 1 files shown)


eng/_util/cmd/build/build.go
```diff
@@ -342,6 +350,7 @@ func build(o *options) (err error) {
 }
 
 func writeDevelVersionFile(goRootDir, executableExtension string) (string, error) {
+	// Use "go tool dist version" to directly get the version the toolset would call itself.
 	cmd := exec.Command(filepath.Join(goRootDir, "bin", "go"+executableExtension), "tool", "dist", "version")
```
dagood, line 351: This now runs before the build happens, which doesn't work. I suppose the VERSION file handling actually needs to stay where it is.

dagood:
A cherry-pick to 1.26 (making sure to clean the submodule before building) also works now:

```
$ GODEBUG=ms_version=1 go/bin/go version
go version go1.26.0-1-microsoft linux/amd64

$ go/bin/go version -m go/bin/go
go/bin/go: go1.26.0
        path    cmd/go
        build   microsoft_toolset_version=go1.26.0-1-microsoft
```

gdams (approved):

PR #2149, opened 2026-02-19
Enable GODEBUG=fips140=only with systemcrypto backends

### LINKED ISSUE #1656: Support `GODEBUG=fips140=only`
`GODEBUG=fips140=only` is not well supported by us. Some checks are not properly enforced. We need to investigate what's missing and fix it.

gdams:
Fixes: #1656

### DIFF
--- a/eng/doc/fips/README.md
+++ b/eng/doc/fips/README.md
@@ -81,7 +81,7 @@ The following table summarizes common configurations and how suitable each one i
 | Default | Default | Compliant | Can be used to create a compliant app. FIPS mode is determined by system-wide configuration. Make sure you are familiar with your platform's system-wide FIPS switch, described in [Usage: Runtime](#usage-runtime). |
 | `GOEXPERIMENT=systemcrypto` | Default | Compliant | Can be used to create a compliant app. The `GOEXPERIMENT` setting is redundant because `systemcrypto` is enabled by default. |
 | Default | `GODEBUG=fips140=on` or `GOFIPS=1` | Compliant | Can be used to create a compliant app. Depending on platform, the app enables FIPS mode, ensures it is already enabled, or doesn't do any additional checks. The app panics if there is a problem. See [Usage: Runtime](#usage-runtime). |
-| Default | `GODEBUG=fips140=only` | Compliant | Same as `fips140=on`, but also panics if a non-FIPS-approved algorithm is used. Note: if the backend does not support a particular algorithm, the call panics rather than falling back to Go standard library crypto. See [Usage: Runtime](#usage-runtime). |
+| Default | `GODEBUG=fips140=only`, Go 1.27+ | Compliant | Same as `fips140=on`, but also panics if a non-FIPS-approved algorithm is used. Note: if the backend does not support a particular algorithm, the call panics rather than falling back to Go standard library crypto. See [Cross-Platform Cryptography](../CrossPlatformCryptography.md) to check algorithm support per platform, and [Usage: Runtime](#usage-runtime). |
 | Default | `GO_OPENSSL_VERSION_OVERRIDE=1.1.1k-fips` | Compliant | Can be used to create a compliant app. On Linux, this environment variable causes the runtime to load `libcrypto.so.1.1.1k-fips` instead of using the automatic search behavior. This environment variable has no effect on Windows or macOS. |
 | `-tags=requirefips` | Default | Compliant | Can be used to create a compliant app. The behavior is the same as `GODEBUG=fips140=on` and `GOFIPS=1`, but no runtime configuration is necessary. See [the `requirefips` section](#build-option-to-require-fips-mode) for more information on when this "locked-in" approach may be useful rather than the flexible approach. |
 | `MS_GO_NOSYSTEMCRYPTO=1` or `GOEXPERIMENT=nosystemcrypto` | Default | Not compliant | Crypto usage is not FIPS compliant. |
@@ -164,13 +164,14 @@ If FIPS mode preference is Enabled ✅, then:
 The Go runtime FIPS mode may be important to distinguish its FIPS mode from the system FIPS mode or crypto engine's FIPS mode.
 The [Go Runtime FIPS mode](#go-runtime-fips-mode) section describes this in more detail.
 
-There is also `GODEBUG=fips140=only`.
+Since Go 1.27, there is also `GODEBUG=fips140=only`.
 It acts as `GODEBUG=fips140=on`, but also makes a best effort to panic if a non-FIPS 140-3 compliant algorithm is used.
 
 > [!NOTE]
 > When `fips140=only` is set with a system crypto backend, the enforcement depends on the backend's algorithm support.
 > If the backend does not support a particular FIPS-approved algorithm (e.g. SHA-512/224 on macOS, or CTR mode on macOS), a call to that algorithm will panic rather than falling back to the Go standard library implementation.
 > This means that `fips140=only` may restrict the set of usable algorithms compared to `fips140=on`, depending on the platform.
+> See [Cross-Platform Cryptography](../CrossPlatformCryptography.md) to check which algorithms are supported on each platform.
 
 > [!NOTE]
 > The `GODEBUG`, `GOFIPS`, and `GOLANG_FIPS` options described in this section have no effect at build time, only runtime. When the Go program starts up, it examines its environment variables and other platform-specific configurations. This is normally the desired behavior. See [`requirefips`](#build-option-to-require-fips-mode) for info about an optional build tag that may affect FIPS mode.
@@ -422,6 +423,7 @@ This list of major changes is intended for quick reference and for access to his
 
 ### Go 1.27 (Aug 2026)
 
+- Support for `GODEBUG=fips140=only` has been added. It acts as `fips140=on`, but also panics if a non-FIPS-approved algorithm is used.
 - The per-platform GOEXPERIMENTs (`opensslcrypto`, `cngcrypto`, `darwincrypto`) have been removed.
   - The `systemcrypto` GOEXPERIMENT is now enabled by default on all platforms where it's supported (Windows, Linux and macOS). To disable it, set `GOEXPERIMENT=nosystemcrypto`.
   - Using any of the removed experiments will result in a build error.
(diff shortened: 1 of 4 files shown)


patches/0004-Use-crypto-backends.patch
```diff
@@ -1090,7 +1172,18 @@ index b336f32eb69ef5..1eb2604505e229 100644
  		b, err := boringPrivateKey(priv)
  		if err != nil {
  			return nil, err
-@@ -494,7 +494,7 @@ func addASN1IntBytes(b *cryptobyte.Builder, bytes []byte) {
+@@ -416,7 +416,9 @@ func signFIPS[P ecdsa.Point[P]](c *ecdsa.Curve[P], priv *PrivateKey, rand io.Rea
+ 	// Always using SHA-512 instead of the hash that computed hash is
+ 	// technically a violation of draft-irtf-cfrg-det-sigs-with-noise-04 but in
+ 	// our API we don't get to know what it was, and this has no security impact.
+-	sig, err := ecdsa.Sign(c, sha512.New, k, rand, hash)
++	// Use the FIPS module's SHA-512 directly to ensure the service indicator
```
gdams: I don't think this change is neccessary, adding a comment to remind me

patches/0004-Use-crypto-backends.patch
```diff
@@ -1506,6 +1599,40 @@ index e7976e25193dfe..9578e1e985c02d 100644
...
+ func New(h func() hash.Hash, key []byte) hash.Hash {
+-	if boring.Enabled {
+-		hm := boring.NewHMAC(h, key)
+-		if hm != nil {
+-			return hm
+-		}
+-		// BoringCrypto did not recognize h, so fall through to standard Go code.
+-	}
+-	h = fips140hash.UnwrapNew(h)
++	fh := fips140hash.UnwrapNew(h)
```
qmuntal: Keep the original variable name, `h`.

patches/0004-Use-crypto-backends.patch
```diff
@@ -3425,7 +3810,7 @@ index 080bf694f03652..0c899f5c407ffa 100644
  	cryptotest.SkipTestAllocations(t)
 +	expectedAllocations := 0.0
 +	if boring.Enabled {
-+		expectedAllocations = 2
++		expectedAllocations = 6
```
gdams: @qmuntal I wanted you to sanity check this increase in allocations please :)

gdams: Note that on Windows and Darwin the number of allocations is even higher (10)

patches/0004-Use-crypto-backends.patch
```diff
@@ -1635,8 +1743,46 @@ index 2b6cf4b75fc6b7..3a1b9e5cc67ecc 100644
...
+--- a/src/crypto/internal/fips140only/fips140only.go
++++ b/src/crypto/internal/fips140only/fips140only.go
+@@ -24,9 +24,11 @@ func ApprovedHash(h hash.Hash) bool {
+ 	switch h.(type) {
+ 	case *sha256.Digest, *sha512.Digest, *sha3.Digest:
+ 		return true
+-	default:
+-		return false
+ 	}
++	if fa, ok := h.(interface{ FIPSApproved() bool }); ok {
```
dagood: Why is this designed as an interface vs. adding something like the `Supports` funcs? Maybe it doesn't even need to live in the backend repo?

(https://github.com/golang-fips/openssl/pull/373 says it's required, but I don't know why it's required.)

patches/0004-Use-crypto-backends.patch
```diff
@@ -307,6 +311,76 @@ index 097c37e343fdb8..a5d603896d3890 100644
...
  func Enabled() bool {
+diff --git a/src/crypto/cipher/cbc.go b/src/crypto/cipher/cbc.go
+index 87bafee08ade4f..d724c21e76f7f1 100644
+--- a/src/crypto/cipher/cbc.go
++++ b/src/crypto/cipher/cbc.go
+@@ -44,6 +44,10 @@ type cbcEncAble interface {
+ 	NewCBCEncrypter(iv []byte) BlockMode
+ }
+ 
++type fipsCBCEncAble interface {
```
dagood, line 324: Possible to describe what this is for in a comment? (Like `cbcEncAble` is.)

patches/0004-Use-crypto-backends.patch
```diff
@@ -1506,6 +1580,40 @@ index e7976e25193dfe..9578e1e985c02d 100644
...
+-		// BoringCrypto did not recognize h, so fall through to standard Go code.
+-	}
+-	h = fips140hash.UnwrapNew(h)
++	fh := fips140hash.UnwrapNew(h)
+ 	if fips140only.Enforced() {
+ 		if len(key) < 112/8 {
+ 			panic("crypto/hmac: use of keys shorter than 112 bits is not allowed in FIPS 140-only mode")
+ 		}
+-		if !fips140only.ApprovedHash(h()) {
++		if !fips140only.ApprovedHash(fh()) {
```
dagood: Is it valid/intended to call `fips140hash.UnwrapNew` on every `hmac.New` call on the given `h`, whether or not it's from systemcrypto? I'm not super familiar with `fips140hash` but it seems like it might only be intended to be compatible with hashes from the upstream FIPS crypto module.

patches/0004-Use-crypto-backends.patch
```diff
@@ -3364,8 +3709,48 @@ index 1435eac1f5b5dc..17e8501154762a 100644
  	"crypto/internal/fips140/sha512"
  	"hash"
  )
+@@ -58,6 +58,9 @@ func New() hash.Hash {
+ // [encoding.BinaryUnmarshaler] to marshal and unmarshal the internal
+ // state of the hash.
+ func New512_224() hash.Hash {
++	if boring.Enabled && boring.SupportsHash(crypto.SHA512_224) {
```
dagood, line 3721: Is this PR also adding support (or maybe theoretical support) for more hashes?

patches/0004-Use-crypto-backends.patch
dagood: Would it be worthwhile upgrading the backends in a separate PR rather than together with systemcrypto changes to make sure the diff is coming from backend changes?

qmuntal: Agree with @dagood.

> @qmuntal I wanted you to sanity check this increase in allocations please :)

We can probably improve 512_224 and 512_256 allocations a bit, but they will never be 0 as those contains fallback code that preclude many compiler optimizations.

qmuntal (changes requested):

patches/0004-Use-crypto-backends.patch
qmuntal: > Why is this designed as an interface vs. adding something like the Supports funcs?

It could also be a top-level function, good point. In fact this might be slightly better, because then we don't add new methods to the hash objects returned by sha256.New (and friends), even when those methods are not documented and can only be accessed via type-casting.

@gdams could you update the backend to replace `(h *Hash) FIPSApproved() bool` with `func FIPSApprovedHash(h hash.Hash) bool`?

> Maybe it doesn't even need to live in the backend repo?

It does. The [hash.Hash](https://pkg.go.dev/hash#Hash) interface doesn't allow you to know which kind of hash it is, and not all hashes provided by the backend are FIPS compliant (e.g. MD5). We need to check backend-private fields to know that.

patches/0004-Use-crypto-backends.patch
gdams: > Would it be worthwhile upgrading the backends in a separate PR rather than together with systemcrypto changes to make sure the diff is coming from backend changes?

Agreed, I've moved these changes to https://github.com/microsoft/go/pull/2153

patches/0004-Use-crypto-backends.patch
gdams: it turns out we were missing this logic despite our docs claiming support. I've wrapped this up into https://github.com/microsoft/go/pull/2153 so will pull these changes out

patches/0004-Use-crypto-backends.patch
gdams: Yes, it's valid. `UnwrapNew` is safe to call unconditionally. For non-SHA3 hashes it's a no-op identity function. For SHA3 hashes with systemcrypto, it correctly extracts the backend's hash object, which is what we want

patches/0004-Use-crypto-backends.patch
gdams: > @gdams could you update the backend to replace (h *Hash) FIPSApproved() bool with func FIPSApprovedHash(h hash.Hash) bool?

I've pushed a fix

qmuntal (approved):

eng/doc/fips/README.md
```diff
@@ -81,6 +81,7 @@ The following table summarizes common configurations and how suitable each one i
 | Default | Default | Compliant | Can be used to create a compliant app. FIPS mode is determined by system-wide configuration. Make sure you are familiar with your platform's system-wide FIPS switch, described in [Usage: Runtime](#usage-runtime). |
 | `GOEXPERIMENT=systemcrypto` | Default | Compliant | Can be used to create a compliant app. The `GOEXPERIMENT` setting is redundant because `systemcrypto` is enabled by default. |
 | Default | `GODEBUG=fips140=on` or `GOFIPS=1` | Compliant | Can be used to create a compliant app. Depending on platform, the app enables FIPS mode, ensures it is already enabled, or doesn't do any additional checks. The app panics if there is a problem. See [Usage: Runtime](#usage-runtime). |
+| Default | `GODEBUG=fips140=only` | Compliant | Same as `fips140=on`, but also panics if a non-FIPS-approved algorithm is used. Note: if the backend does not support a particular algorithm, the call panics rather than falling back to Go standard library crypto. See [Usage: Runtime](#usage-runtime). |
```
dagood: Missing version applicability note?

eng/doc/fips/README.md
```diff
@@ -165,8 +166,11 @@ The [Go Runtime FIPS mode](#go-runtime-fips-mode) section describes this in more
 
 There is also `GODEBUG=fips140=only`.
 It acts as `GODEBUG=fips140=on`, but also makes a best effort to panic if a non-FIPS 140-3 compliant algorithm is used.
-The `only` setting is not yet supported in the Microsoft build of Go.
-(See [microsoft/go#1656 Support `GODEBUG=fips140=only`](https://github.com/microsoft/go/issues/1656).)
+
+> [!NOTE]
+> When `fips140=only` is set with a system crypto backend, the enforcement depends on the backend's algorithm support.
```
dagood, line 171: Missing version applicability note?

eng/doc/fips/README.md
dagood: Refer to https://github.com/microsoft/go/blob/microsoft/main/eng/doc/CrossPlatformCryptography.md ?

gdams:
I'm going to enable automerge on this PR now. If we have follow-up docs queries I'm happy to raise follow-up PRs :)

PR #2150, opened 2026-02-19
eng/pipeline: add prerequisite patches check stage

### LINKED ISSUE #2148: Move "Patches Apply Cleanly" check to the top
Right now the CI results look like this when patches don't apply cleanly:

Image

It takes several clicks to actually get from the PR screen to the pipeline and the patches step to see that that's all that went wrong.

Instead, the patches-specific check should be on top, for easier diagnosis:

Image

Currently the checks appear to be in alphabetical order, so we might have to use an ordering trick like renaming the check to

`0: Test / Patches Apply Cleanly`

Comes up a lot, e.g. in submodule updates:

* https://github.com/microsoft/go/pull/2147

A release involves a submodule update, so this improvement could help the release workflow.

gdams:
fixes #2148 

Add a "Patches Apply Cleanly" stage that runs before all build/test stages. This surfaces patch application failures at the top of the PR checks list, so developers don't have to drill into the pipeline to find them.

Previously, patch failures were only visible by clicking through to the pipeline run and finding the "Apply patches" step inside a specific build job. This is especially tedious during submodule updates and releases where patch conflicts are common.

The new stage runs on Ubuntu 22.04 with a 10-minute timeout. All build, test, and CodeQL stages now depend on it and will be skipped if patches fail to apply.

Example of how things look when the patches don't apply cleanly:

image


dagood (changes requested):
Unfortunately AzDO obscures the time spent waiting for an agent so it's hard to get exact numbers, but I don't like that this PR adds extra latency to every test run--an entire "wait for agent" delay that will always be on the hot path. (Let alone the ~1m of actual execution time.)

gdams:
closing for now

PR #2153, opened 2026-02-24
hash: implement missing boring logic for SHA512_224/256

gdams:
Our docs have incorrectly claimed that we already support these on OpenSSL. We should also probably backport this to our active release branches


qmuntal (approved):

qmuntal:
> We should also probably backport this to our active release branches

Not sure about this, IMO better fix the docs than to change behavior mid-release.

PR #2160, opened 2026-02-25
[release-branch.go1.26] Include revision in Microsoft-specific version string (#2140)

dagood:
* Include revision in Microsoft-specific version string

* Split VERSION and MICROSOFT_REVISION handling

(cherry picked from commit 165873c80410bd67ed47e281eb52e442c2680a45)

* Port https://github.com/microsoft/go/pull/2140
* For https://github.com/microsoft/go/issues/2138

### DIFF
--- a/eng/_util/cmd/build/build.go
+++ b/eng/_util/cmd/build/build.go
@@ -110,6 +110,7 @@ func build(o *options) (err error) {
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


dagood:
I don't think we *need to* backport this, but it's almost entirely an opt-in feature, so risk is low, and it would be best for early adopters to experience the actual format/data we're aiming for.

karianna (approved):

PR #2171, opened 2026-03-03
Enforce FIPS 140-only mode for ChaCha20Poly1305 in backends

gdams:
Add fips140only.Enforced() checks to NewChaCha20Poly1305 in all three backend implementations (darwin, linux, windows) so that ChaCha20Poly1305 is rejected in FIPS 140-only mode even when a crypto backend is enabled.

Remove the workaround in fips140only_test.go that skipped the HPKE ChaCha20Poly1305 error checks when a backend was enabled.


qmuntal (approved):
Like this approach very much!

PR #2186, opened 2026-03-12
Add msgo/modulehash telemetry counter

gdams:
Emit a SHA-256 hash of the main module path as a numeric telemetry counter during build, install, and run commands.

Re-vendor go-infra telemetry to pick up wildcard prefix support in the upload filter.


qmuntal:
We are not adding this counter for now.

PR #2198, opened 2026-03-19
Improve NocgoOpenSSL.md

qmuntal:
Clarify how the nocgo mode is selected and add a runtime dependencies section.

### DIFF
--- a/eng/doc/NocgoOpenSSL.md
+++ b/eng/doc/NocgoOpenSSL.md
@@ -55,6 +55,8 @@ The following are the expected runtime dependencies other than OpenSSL itself on
 > The `libdl.so.2` and `libpthread.so.0` functionality is already provided by `libc.so.6` since glibc 2.34.
 > The Microsoft build of Go supports platforms with older versions of glibc, so it still links to `libdl.so.2` and `libpthread.so.0` directly.
 >
+> There is no runtime dependency on the same libc that the program was built with. You can, for example, build a program on a system with glibc 2.34 or later and run it on a system with an older glibc version.
+>
 > If your platform doesn't provide these libraries, the preferred solution is to install the appropriate runtime or compatibility packages so that `libdl.so.2` and `libpthread.so.0` are available in standard system library directories (for example `/lib` or `/usr/lib`).
 >
 > As an advanced workaround, you may instead symlink `libdl.so.2` and `libpthread.so.0` to `libc.so.6` in a directory searched by the dynamic linker (for example, a system library directory) and update the loader cache if required (for example by running `ldconfig`); this typically requires root privileges and should only be done if you understand the implications for your system's C library.
--- a/eng/doc/NocgoOpenSSL.md
+++ b/eng/doc/NocgoOpenSSL.md
@@ -32,8 +32,23 @@ To see existing requests or request support for additional architectures, use th
 
 ## How the backend is selected
 
-If cgo is enabled (e.g., `CGO_ENABLED=1`), the cgo-based OpenSSL backend is always used.
-The cgo-less backend is only used when cgo is disabled **and** you're on a supported architecture.
+* If cgo is explicitly enabled with `CGO_ENABLED=1`, the cgo-based OpenSSL backend is used.
+* If cgo is disabled with `CGO_ENABLED=0` and you're on a supported architecture, the cgo-less OpenSSL backend is used.
+* If cgo is not explicitly enabled or disabled, then the cgo-less OpenSSL backend is used if the Go toolchain decides that cgo should be disabled. The are the rules at the time of writing:
+  * cgo is disabled if the Go toolchain is build with `CGO_ENABLED=0`. Note that that's not currently the case for the Microsoft build of Go. 
+  * cgo is disabled when cross-compiling.
+  * cgo is diables if the `CC` environment variable is not set to an existing executable.
 
-This means if cgo is enabled by default on your platform or you enable cgo for other reasons (e.g., linking your own C library), the crypto backend will use the cgo-based implementation.
-You can also intentionally enable cgo to successfully build for a platform without cgo-less OpenSSL support.
+## Runtime dependencies
+
+The cgo-less OpenSSL backend still relies in the C machinery to load OpenSSL at runtime and manage threading.
+
+This is an exhaustive list of the runtime dependencies other than OpenSSL itself:
+
+* `libc.so.6` for various C library functions. There is no minimum version requirement.
+* `libdl.so.2` for loading OpenSSL at runtime using `dlopen`. There is no minimum version requirement.
+* `libpthread.so.0` for managing Go's threads. There is no minimum version requirement.
+
+Note that `libdl.so.2` and `libpthread.so.0` functionality is already provided by `libc.so.6` since glibc 2.34.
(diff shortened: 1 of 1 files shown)


gdams (approved):

gdams (approved):

eng/doc/NocgoOpenSSL.md
```diff
@@ -32,8 +32,24 @@ To see existing requests or request support for additional architectures, use th
 
 ## How the backend is selected
 
-If cgo is enabled (e.g., `CGO_ENABLED=1`), the cgo-based OpenSSL backend is always used.
-The cgo-less backend is only used when cgo is disabled **and** you're on a supported architecture.
+* If cgo is explicitly enabled with `CGO_ENABLED=1`, the cgo-based OpenSSL backend is used.
+* If cgo is disabled with `CGO_ENABLED=0` and you're on a supported architecture, the cgo-less OpenSSL backend is used.
```
dagood: I think it's worth clarifying what happens when you're not on a supported architecture.

```suggestion
* If cgo is disabled with `CGO_ENABLED=0`, then:
  * If you're on a supported architecture, the cgo-less OpenSSL backend is used.
  * If you're on an unsupported architecture, the build fails.
```

(Not 100% sure bifurcation is needed, maybe just a note as a sub-bullet, but I definitely think something is necessary.)

eng/doc/NocgoOpenSSL.md
```diff
@@ -32,8 +32,24 @@ To see existing requests or request support for additional architectures, use th
...
+* If cgo is not explicitly enabled or disabled, then the cgo-less OpenSSL backend is used if the Go toolchain decides that cgo should be disabled. These are the rules at the time of writing:
```
dagood: Include date so anyone who wants to know the time of writing doesn't have to dig through Git, also include an official link:


```suggestion
* If cgo is not explicitly enabled or disabled, then the cgo-less OpenSSL backend is used if the Go toolchain decides that cgo should be disabled. The rules are officially described by [cmd/cgo](https://pkg.go.dev/cmd/cgo). This is a summary, at the time of writing (2026-03-19):
```

eng/doc/NocgoOpenSSL.md
```diff
@@ -32,8 +32,24 @@ To see existing requests or request support for additional architectures, use th
...
+
+The cgo-less OpenSSL backend still relies on the C machinery to load OpenSSL at runtime and manage threads.
+
+The following are the expected runtime dependencies other than OpenSSL itself on glibc-based systems:
+
+* `libc.so.6` for various C library functions.
+* `libdl.so.2` for loading OpenSSL at runtime using `dlopen`.
+* `libpthread.so.0` for managing Go's threads.
+
+Note that `libdl.so.2` and `libpthread.so.0` functionality is already provided by `libc.so.6` since glibc 2.34.
```
dagood: Might also be worth noting that this means there is no runtime dependency on the *same* libc version that the program was built on? I think fully understanding this paragraph leaves you with the right impression, but this is a big scare with libc and I think it's worth addressing directly.

eng/doc/NocgoOpenSSL.md
```diff
@@ -32,8 +32,24 @@ To see existing requests or request support for additional architectures, use th
...
+  * cgo is disabled when cross-compiling.
+  * cgo is disabled if the `CC` environment variable is not set to an existing executable.
 
-This means if cgo is enabled by default on your platform or you enable cgo for other reasons (e.g., linking your own C library), the crypto backend will use the cgo-based implementation.
-You can also intentionally enable cgo to successfully build for a platform without cgo-less OpenSSL support.
+## Runtime dependencies
+
+The cgo-less OpenSSL backend still relies on the C machinery to load OpenSSL at runtime and manage threads.
+
+The following are the expected runtime dependencies other than OpenSSL itself on glibc-based systems:
```
dagood, line 48: Should we also document what's expected on musl systems?

eng/doc/NocgoOpenSSL.md
```diff
@@ -32,8 +32,24 @@ To see existing requests or request support for additional architectures, use th
...
+
+* `libc.so.6` for various C library functions.
+* `libdl.so.2` for loading OpenSSL at runtime using `dlopen`.
+* `libpthread.so.0` for managing Go's threads.
+
+Note that `libdl.so.2` and `libpthread.so.0` functionality is already provided by `libc.so.6` since glibc 2.34.
+The Microsoft build of Go supports platforms with older versions of glibc, so it still links to `libdl.so.2` and `libpthread.so.0` directly.
+If your platform doesn't provide these libraries, the preferred solution is to install the appropriate runtime or compatibility packages so that `libdl.so.2` and `libpthread.so.0` are available in standard system library directories (for example `/lib` or `/usr/lib`).
+As an advanced workaround, you may instead create symlinks to `libc.so.6` in a directory searched by the dynamic linker (for example, a system library directory) and update the loader cache if required (for example by running `ldconfig`); this typically requires root privileges and should only be done if you understand the implications for your system's C library.
```
dagood: This is pretty imposing when rendered. Maybe split, and make it a "real" note?

Also, nit: name the symlinks.


```suggestion
> [!NOTE]
> The `libdl.so.2` and `libpthread.so.0` functionality is already provided by `libc.so.6` since glibc 2.34.
> The Microsoft build of Go supports platforms with older versions of glibc, so it still links to `libdl.so.2` and `libpthread.so.0` directly.
>
> If your platform doesn't provide these libraries, the preferred solution is to install the appropriate runtime or compatibility packages so that `libdl.so.2` and `libpthread.so.0` are available in standard system library directories (for example `/lib` or `/usr/lib`).
>
> As an advanced workaround, you may instead symlink `libdl.so.2` and `libpthread.so.0` to `libc.so.6` in a directory searched by the dynamic linker (for example, a system library directory) and update the loader cache if required (for example by running `ldconfig`); this typically requires root privileges and should only be done if you understand the implications for your system's C library.
```

> [!NOTE]
> The `libdl.so.2` and `libpthread.so.0` functionality is already provided by `libc.so.6` since glibc 2.34.
> The Microsoft build of Go supports platforms with older versions of glibc, so it still links to `libdl.so.2` and `libpthread.so.0` directly.
>
> If your platform doesn't provide these libraries, the preferred solution is to install the appropriate runtime or compatibility packages so that `libdl.so.2` and `libpthread.so.0` are available in standard system library directories (for example `/lib` or `/usr/lib`).
>
> As an advanced workaround, you may instead symlink `libdl.so.2` and `libpthread.so.0` to `libc.so.6` in a directory searched by the dynamic linker (for example, a system library directory) and update the loader cache if required (for example by running `ldconfig`); this typically requires root privileges and should only be done if you understand the implications for your system's C library.

eng/doc/NocgoOpenSSL.md
qmuntal: Yes, but I still don't know what to expect in there. Will leave it for a future PR.

dagood (approved):

dagood:
/azp run microsoft-go

(IIRC this will get it to use a newer merge commit to skip https://github.com/microsoft/go/issues/2200...)

dagood:
> (IIRC this will get it to use a newer merge commit to skip #2200...)

Didn't work, but some more of the macOS failures passed. Maybe it's over? Hitting retry again.

Closing and re-opening would probably have worked, but that would redo the GitHub Actions workflows too I believe.

PR #2203, opened 2026-03-20
Parallelize Patch Build sequence test

dagood:
This workflow tests that Go is buildable after applying each patch. This takes a while, and scales O(N) per patch. This creates an artificial limit to how many patch files we add.

This PR parallelizes this workflow to O(1) per patch, as long as we're able to freely reserve agents.

Addresses https://github.com/microsoft/go/pull/2202#issuecomment-4099255874

### DIFF
--- a/.github/workflows/patch-build.yml
+++ b/.github/workflows/patch-build.yml
@@ -52,8 +52,9 @@ jobs:
 
       - name: Build
         run: |
-          cd ${{ github.workspace }}/go/src
+          set -x
           pwsh eng/run.ps1 build
+          cd ${{ github.workspace }}/go/src
           ${{ github.workspace }}/go/bin/go mod vendor
           cd ${{ github.workspace }}/go/src/cmd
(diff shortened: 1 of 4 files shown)


dagood:
(Delete was a misclick, seems like I misinterpreted review UI.)

dagood:
Interesting. Switching to `pwsh eng/run.ps1 build` (automatically get the same Microsoft-built bootstrap Go we use in other builds) also resulted in the race install running, getting us this in [Patch Build / Build up to 0003-Implement-crypto-internal-backend.patch (pull_request)](https://github.com/microsoft/go/actions/runs/23356991949/job/67950645346?pr=2203):

* https://github.com/microsoft/go/issues/2204

We might not actually care if this part of buildability matters, or only the "core" part of the build.

For now, skipping the race build. (This was the old behavior.)

gdams (approved):

qmuntal (approved):

PR #2209, opened 2026-03-23
Pass attempt label to JUnit output to distinguish test retries

### LINKED ISSUE #2134: Make test results distinguish the attempt
These two entries are two attempts, but there's no way to distinguish them except correlating which tests fail (if any) vs. the job logs:

> Image

michelle-clayton-work:
Followup PR to https://github.com/microsoft/go-infra/pull/449

- Add -attempt flag to run-builder, passed through to json2junit
- Each retry attempt writes a unique JUnit XML file (TestResults-attempt-N.xml)
- Publish all attempt files via wildcard in PublishTestResults
- Update go-infra dependency to pick up Attempt support in json2junit

This ensures test result entries in AzDO distinguish which attempt they correspond to, both via unique file names and via the attempt label embedded in JUnit test suite names.

Fixes: https://github.com/microsoft/go/issues/2134


eng/pipeline/stages/run-stage.yml
```diff
@@ -265,8 +265,9 @@ stages:
                       -builder '${{ parameters.builder.os }}-${{ parameters.builder.arch }}-${{ parameters.builder.config }}' `
                       $(if ('${{ parameters.builder.experiment }}') { '-experiment'; '${{ parameters.builder.experiment }}' }) `
                       $(if ('${{ parameters.builder.fips }}') { '-fipsmode' }) `
-                      -junitout '$(Build.SourcesDirectory)/eng/artifacts/RawTestOutput/TestResults.xml' `
-                      -rawtestout '$(Build.SourcesDirectory)/eng/artifacts/RawTestOutput/raw-json-attempt-${{ attempt }}.txt'
+                      -junitout '$(Build.SourcesDirectory)/eng/artifacts/RawTestOutput/TestResults-attempt-${{ attempt }}.xml' `
+                      -rawtestout '$(Build.SourcesDirectory)/eng/artifacts/RawTestOutput/raw-json-attempt-${{ attempt }}.txt' `
+                      -attempt '${{ attempt }}'
```
dagood: Oh, I wasn't clear about what kind of attempts I meant in https://github.com/microsoft/go/issues/2134... I was thinking about [`System.JobAttempt`](https://learn.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml). But it actually makes sense to keep each `parameters.retryAttempts` result as well so that we can use it to track test flakiness in the future. So, I guess we should keep/fix both `-attempt` and (say) `-jobattempt` types of retry here.

dagood:
> Fixes: #2134

GitHub quirk: if you put this in a list (without changing anything else), GitHub grabs the issue title. I do this basically always--more immediately readable and makes it obvious when I've grabbed the wrong issue.

`* Fixes: #2134`

* Fixes: #2134

michelle-clayton-work:
image

eng/pipeline/stages/run-stage.yml
```diff
@@ -265,8 +265,9 @@ stages:
                       -builder '${{ parameters.builder.os }}-${{ parameters.builder.arch }}-${{ parameters.builder.config }}' `
                       $(if ('${{ parameters.builder.experiment }}') { '-experiment'; '${{ parameters.builder.experiment }}' }) `
                       $(if ('${{ parameters.builder.fips }}') { '-fipsmode' }) `
-                      -junitout '$(Build.SourcesDirectory)/eng/artifacts/RawTestOutput/TestResults.xml' `
-                      -rawtestout '$(Build.SourcesDirectory)/eng/artifacts/RawTestOutput/raw-json-attempt-${{ attempt }}.txt'
+                      -junitout '$(Build.SourcesDirectory)/eng/artifacts/RawTestOutput/TestResults-attempt-${{ attempt }}.xml' `
+                      -rawtestout '$(Build.SourcesDirectory)/eng/artifacts/RawTestOutput/raw-json-attempt-${{ attempt }}.txt' `
+                      -jobattempt '$(System.StageAttempt)'
```
dagood: Should this one be System.JobAttempt? That would make more sense to me as a way to make the XML file end up with the job attempt baked in, and `testRunTitle` being what applies the stage attempt on screen.

```suggestion
                      -jobattempt '$(System.JobAttempt)'
```

I did forget that AzDO has this barely-documented pair... unfortunate.

dagood (commented):
Do you have a test run with a simulated failure showing what job/stage attempts look like with this change? I'm not quite sure how it'll show up, and why/when to use `jobattempt` vs. `testRunTitle`.

dagood (approved):

PR #2192, opened 2026-03-17
Revendor backends

michelle-clayton-work:

PR #2210, opened 2026-03-24
Test PR to exercise review: unrevendor backends

dagood:
* To simulate a vendor patch change, reverts microsoft/go#2192

It seems like Copilot review has changed to expose the session. This might help figure out what is going wrong with the reviewer when it sees a vendor patch change. Try it out.


dagood:
Well, that was uninformative. 😕

Issue #12, opened 2021-03-12
Improve how the build script sets up the "stage 0" Go toolset


dagood:
The build script currently downloads, verifies, and extracts the Go linux-amd64 binary release 1.16 to `~/.go-stage-0/1.16`. This is used in CI, and it can be used locally get a build going on a machine with minimal dependencies.

There are a few things to improve:

It should detect if Go is already installed and (optionally) use that.
* We should probably still make it possible to download a fresh copy of 1.16 anyway, in case it makes it easier to reproduce failures that happen in CI.

It should download Go somewhere under the repository directory, not `$HOME`.
  * IMO, building a repo should not modify state in my home directory unless it has an extremely good reason.
  * This makes it easy to clean up in case a fresh repro is needed (it gets deleted by `git clean -xdf`).
  * However, putting Go inside the repo causes the test suite to fail, probably due to upwards directory traversal from `go`.
    ```
    --- FAIL: TestAllDependencies (6.36s)
        --- FAIL: TestAllDependencies/std(quick) (1.80s)
            moddeps_test.go:63: /work/bin/go list -mod=vendor -deps ./...: exit status 1
                package std/bytes
                    bytes/bytes.go:10:2: use of internal package internal/bytealg not allowed
                [...]
            moddeps_test.go:64: (Run 'go mod vendor' in /work/microsoft/.go-stage-0/go/src to ensure that dependecies have been vendored.)
    FAIL
    FAIL    cmd/internal/moddeps    6.405s
    ```

This issue is referenced in source.

PR #2230, opened 2026-04-14
Install stage 0 Go into repo instead of home directory

michelle-clayton-work:
Move the stage 0 Go toolset download location from the user's home directory (~/.local/share/microsoft-go/) into the repository at eng/artifacts/_goStage0/. This has several benefits:

- Building the repo no longer modifies state in the home directory.
- Cleanup is easy: the directory is removed by 'git clean -xdf' since eng/artifacts/ is already gitignored.
- The '_' prefix on the directory name prevents Go's moddeps_test from traversing into the stage 0 installation directory and failing.

The existing MS_USE_PATH_GO=1 opt-in to use a pre-installed Go from PATH continues to work as before.

Fixes 
- https://github.com/microsoft/go/issues/12

### DIFF
--- a/eng/run.ps1
+++ b/eng/run.ps1
@@ -17,10 +17,9 @@ To list all possible tools:
 
 Builds 'eng/_util/cmd/<tool>/<tool>.go' and runs it using the list of arguments.
 
-This command automatically installs a known version of the Microsoft build Go that
-will be used to build the tools. The known version of Go will also be used to build the
-Go source code, if it's built. Set environment variable "MS_USE_PATH_GO" to 1 to
-your own Go from PATH instead.
+This command uses Go from PATH by default. Set "MS_USE_CI_STAGE0" to 1 to instead
+download the exact Go version used in CI into the repository at
+'eng/artifacts/_goStage0/'. This is useful for reproducing CI-specific failures.
 
 Every tool accepts a '-h' argument to show tool usage help.
 #>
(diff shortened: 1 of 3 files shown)


eng/run.ps1
```diff
@@ -92,6 +101,16 @@ if ($env:MS_USE_PATH_GO -eq "1") {
...
 } else {
+  # Check if Go is already on PATH and let the user know they can use it.
+  if ($env:MS_FORCE_STAGE0_DOWNLOAD -ne "1") {
+    $pathGo = Get-Command go -ErrorAction SilentlyContinue
+    if ($pathGo) {
+      $pathGoVersion = & go version 2>$null
+      if ($pathGoVersion) {
+        Write-Host "Note: Go is already on PATH ($pathGoVersion). Set MS_USE_PATH_GO=1 to use it instead of downloading."
+      }
+    }
```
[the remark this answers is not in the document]
michelle-clayton-work: Now uses $pathGo.Source to invoke the exact resolved executable path

eng/run.ps1
```diff
@@ -82,7 +87,11 @@ if (-not ($tool_source -is [System.IO.FileInfo])) {
 # Now that we have a single result, navigate upwards to see which module it's in.
 $tool_module = $tool_source.Directory.Parent.Parent.FullName
 
-# Download a consistent stage 0 version of Go unless opted out.
+# Set up a stage 0 Go toolset. See https://github.com/microsoft/go/issues/12
+#  - "MS_USE_PATH_GO=1": use Go already on PATH (skip download).
+#  - "MS_FORCE_STAGE0_DOWNLOAD=1": always download (useful for CI repro).
```
dagood: I'm curious about this use case... why would it be better to set `MS_FORCE_STAGE0_DOWNLOAD=1` rather than unsetting `MS_USE_PATH_GO=1`, given they're both variables that a dev would have intentionally set up? I feel like this makes more sense if the script by default has the `MS_USE_PATH_GO=1` behavior and we remove the `MS_USE_PATH_GO` env var entirely.

And... I actually think this would be reasonable to do, with a name like `MS_USE_CI_STAGE0=1`. 😄 The Results may differ from CI environment log should probably just note the env var by name, as a reminder.

I believe this is what I intended with #12, but the text isn't specific. For context, I filed that long before we added the `MS_USE_PATH_GO` check, and that was added to be able to build on macOS, not specifically for developers.

It seems to me (with more Go experience than when these pwsh scripts were setup initially) that it'll be abundantly clear when a build issue is caused by a wrong bootstrap version. If anything, it'll be good for a developer to run into any issues locally rather than waiting until we bump up the CI stage0 version.

@qmuntal, @gdams, any concern with changing the default?

michelle-clayton-work: Good call — I flipped the default so PATH Go is used by default, removed MS_USE_PATH_GO entirely, and added MS_USE_CI_STAGE0=1 as the opt-in to download the exact CI stage 0 version. CI pipelines set MS_USE_CI_STAGE0=1 in pipeline.yml so their behavior is unchanged. The "use PATH Go" path now logs a reminder about the env var in case someone needs to repro a CI-specific issue.

Definitely open to changing as well!

gdams (approved):

qmuntal (approved):

dagood (approved):

PR #2283, opened 2026-05-18
Add msgo/module/hash telemetry counter

gdams:
## Summary

Add a SHA-256 hash of the Go module path as a custom property on the `go/invocations` telemetry event, and vendor the updated `go-infra/telemetry` dependency (which includes support for counter properties and property filtering).

fixes: https://github.com/microsoft/go-lab/issues/382

### `msgo/module/hash` property on `go/invocations`

When `go/invocations` is recorded, the nearest `go.mod` file is located and the module path is hashed using SHA-256. The resulting hash is sent as a custom property (`msgo/module/hash`) on the `go/invocations` event. This is a one-way hash — the original module path cannot be recovered from it.

The hash is only sent when a `go.mod` file is found. If no `go.mod` is present, `go/invocations` is sent without the property.

## Dependency changes

- Updated vendored `github.com/microsoft/go-infra/telemetry` to include support for `NewWithProperties()`, `PropertyFilter`, and counter property configuration in `config.json`.

## Patch changes

- **Patch 0001** (Vendor external dependencies): Updated go-infra/telemetry dependency version with property support.
- **Patch 0009** (Add appinsights telemetry): Added module hash computation and `NewWithProperties` call in `Inc()`.

## Documentation

- Updated `eng/doc/Telemetry.md` to document the new `msgo/module/hash` property and clarify that module names are not collected (only hashes).

### DIFF
--- a/eng/doc/Telemetry.md
+++ b/eng/doc/Telemetry.md
@@ -24,7 +24,7 @@ The counters and properties collected are defined in the [telemetry upload confi
 
 | Property | Description |
 |---|---|
-| `msgo/module/hash` | A SHA-256 hash of the Go module path from the nearest `go.mod` file. This is a one-way hash — the original module path cannot be recovered from it. Only sent when a `go.mod` file is found in the current directory or a parent directory. |
+| `msgo/module/hash` | A SHA-256 hash of the Go module path from the nearest `go.mod` file. Only sent when a `go.mod` file is found in the current directory or a parent directory. |
 
 ### Subcommand
 
@@ -61,7 +61,7 @@ The counters and properties collected are defined in the [telemetry upload confi
 
 - No source code or file contents.
 - No file paths from the local file system.
-- No module names or paths — only a one-way SHA-256 hash of the module path is collected.
+- No textual module names or paths (a one-way SHA-256 hash of the module path is collected).
 - No environment variable values (only the presence of specific CI-related variables is checked).
 - No network or authentication information.
 - No personally identifiable information (PII).
--- a/eng/doc/Telemetry.md
+++ b/eng/doc/Telemetry.md
@@ -8,17 +8,23 @@ For the blog post announcing this feature, see [Microsoft Go Telemetry](https://
 
 ## What is collected
 
-All telemetry data consists of **counters** — simple event names that are incremented.
+Telemetry data consists of **counters** (simple event names that are incremented) and **properties** (key-value custom dimensions attached to a counter event).
 No source code, file paths, module contents, or personally identifiable information is collected.
 Each telemetry session is assigned a random session ID that cannot be linked back to a user.
 
-The counters collected are defined in the [telemetry upload configuration](/go/src/cmd/vendor/github.com/microsoft/go-infra/telemetry/config/config.json) and are listed below.
+The counters and properties collected are defined in the [telemetry upload configuration](https://github.com/microsoft/go-infra/blob/main/telemetry/config/config.json) and are listed below.
 
 ### Invocation counter
 
 | Counter | Description |
 |---|---|
 | `go/invocations` | Incremented once per `go` command invocation. |
+
+#### Properties on `go/invocations`
+
+| Property | Description |
+|---|---|
+| `msgo/module/hash` | A SHA-256 hash of the Go module path from the nearest `go.mod` file. This is a one-way hash — the original module path cannot be recovered from it. Only sent when a `go.mod` file is found in the current directory or a parent directory. |
 
 ### Subcommand
 
@@ -55,6 +61,7 @@ The counters collected are defined in the [telemetry upload configuration](/go/s
 
 - No source code or file contents.
 - No file paths from the local file system.
+- No module names or paths — only a one-way SHA-256 hash of the module path is collected.
 - No environment variable values (only the presence of specific CI-related variables is checked).
 - No network or authentication information.
 - No personally identifiable information (PII).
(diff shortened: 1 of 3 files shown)


patches/0009-Add-appinsights-telemetry.patch
```diff
@@ -40,22 +40,43 @@ index 71c9d1dc299388..8cab5daec4da54 100644
...
  
-@@ -39,6 +41,14 @@ func incrementConfig() {
+ func Increment() {
+@@ -39,6 +45,28 @@ func incrementConfig() {
  		counter.Inc("go/cgo:disabled")
  	}
  
++	if os.Getenv("MS_GO_INTERNAL_CUSTOMER") != "" {
++		if gomod := modload.FindGoMod(base.Cwd()); gomod != "" {
++			if data, err := os.ReadFile(gomod); err == nil {
```
qmuntal: use `fsys` instead of `os`. The build system might be using file overlays.

qmuntal (changes requested):

patches/0009-Add-appinsights-telemetry.patch
gdams: updated PTAL

qmuntal (approved):

dagood (changes requested):
Discussion in https://github.com/microsoft/go-lab/issues/382

karianna:
@gdams - needs a rebase

gdams:
cherry-picking the CI counter out into https://github.com/microsoft/go/pull/2314

patches/0009-Add-appinsights-telemetry.patch
```diff
@@ -71,6 +76,17 @@ index c6e58935bb0ee2..0cd7d3582c9ee6 100644
 +			counter.Inc("msgo/systemcrypto:enabled")
 +		}
 +	}
++
++	// Send a SHA-256 hash of the module path as a property on go/invocations.
++	if gomod := modload.FindGoMod(base.Cwd()); gomod != "" {
++		if data, err := os.ReadFile(gomod); err == nil {
```
qmuntal: Use `cmd/go/internal/fsys` in case a virtual file system is used.

patches/0009-Add-appinsights-telemetry.patch
```diff
@@ -71,6 +76,17 @@ index c6e58935bb0ee2..0cd7d3582c9ee6 100644
...
++			if mod := gover.GoModLookup(data, "module"); mod != "" {
++				h := sha256.Sum256([]byte(mod))
++				hash := new(big.Int).SetBytes(h[:]).String()
```
qmuntal: We don't  need to encode the hash as a integer. Better just use [hex.EncodeToString](https://pkg.go.dev/encoding/hex#EncodeToString).

patches/0009-Add-appinsights-telemetry.patch
```diff
@@ -71,6 +76,17 @@ index c6e58935bb0ee2..0cd7d3582c9ee6 100644
...
++				counter.IncWithProperties("go/invocations", map[string]string{"msgo/module/hash": hash})
```
qmuntal: The `go/invocations` counter is already incremented in `src/cmd/go/main.go` line 121. This would be duplicating the counter. Better patch that line instead.

qmuntal (changes requested):

qmuntal (approved):

eng/doc/Telemetry.md
```diff
@@ -8,18 +8,24 @@ For the blog post announcing this feature, see [Microsoft Go Telemetry](https://
...
 
 | Counter | Description |
 |---|---|
 | `go/invocations` | Incremented once per `go` command invocation. |
 
+#### Properties on `go/invocations`
+
+| Property | Description |
+|---|---|
+| `msgo/module/hash` | A SHA-256 hash of the Go module path from the nearest `go.mod` file. This is a one-way hash — the original module path cannot be recovered from it. Only sent when a `go.mod` file is found in the current directory or a parent directory. |
```
dagood: Publicly indexed module paths (for example) are trivial to recover--it would be strange for this document to ignore that here. For now, we can at least not make the claim:

```suggestion
| `msgo/module/hash` | A SHA-256 hash of the Go module path from the nearest `go.mod` file. Only sent when a `go.mod` file is found in the current directory or a parent directory. |
```

dagood: Point at https://pkg.go.dev/cmd/go/internal/modload#FindGoMod to avoid describing intricacies of how it detects the go.mod file? Seems worthwhile to avoid describing that `/tmp` quirk (and someone who looks at this deeply should be happy to see the `/tmp` quirk).

eng/doc/Telemetry.md
```diff
@@ -55,6 +61,7 @@ The counters collected are defined in the [telemetry upload configuration](/go/s
 
 - No source code or file contents.
 - No file paths from the local file system.
+- No module names or paths — only a one-way SHA-256 hash of the module path is collected.
```
dagood:
```suggestion
- No textual module names or paths (a one-way SHA-256 hash of the module path is collected).
```

Tried to slightly clarify the intent here, so in cases where the module hash might as well be the module path, the wording doesn't seem odd. Then, change style to fit with neighboring points.

dagood (changes requested):
Some initial doc changes I'd like to see, and we can refine later to explain in more detail if necessary.

gdams:
@dagood updated PTAL

dagood (approved):

dagood:
(Merge conflict blocking merge.)

PR #2337, opened 2026-06-02
docs: document systemcrypto experiment removal

qmuntal:
Documents PR 2320 follow-up:
- removal of systemcrypto and nosystemcrypto GOEXPERIMENT user guidance
- MS_GO_NOSYSTEMCRYPTO opt-out guidance
- retained goexperiment build tags for source compatibility
- Go 1.27 Linux CGO_ENABLED=0 no-cgo OpenSSL behavior

Validation:
- git diff --cached --check

### DIFF
--- a/eng/doc/AdditionalFeatures.md
+++ b/eng/doc/AdditionalFeatures.md
@@ -19,10 +19,8 @@ See also the [Data Collection policy for the Microsoft build of Go](/README.md#d
 We have implemented changes to the Go standard library to call a platform-specific system-provided cryptography library rather than running the pure Go implementation.
 Its main purposes are to comply with Microsoft's internal cryptography policies and implement FIPS 140 compliance.
 
-As of Go 1.25 on Linux and Windows, and Go 1.26 on macOS, this feature is enabled by default on supported platforms.
+This feature is enabled by default on supported platforms.
 
-As of Go 1.27, `systemcrypto` is no longer configured with `GOEXPERIMENT`: `GOEXPERIMENT=systemcrypto` and `GOEXPERIMENT=nosystemcrypto` are rejected.
-If you need to disable the system-provided cryptography backend, set `MS_GO_NOSYSTEMCRYPTO=1` before building the program.
 
 See [the Migration Guide](/eng/doc/MigrationGuide.md) for more information about when, and how, to configure this feature.
 
--- a/eng/doc/AdditionalFeatures.md
+++ b/eng/doc/AdditionalFeatures.md
@@ -19,10 +19,10 @@ See also the [Data Collection policy for the Microsoft build of Go](/README.md#d
 We have implemented changes to the Go standard library to call a platform-specific system-provided cryptography library rather than running the pure Go implementation.
 Its main purposes are to comply with Microsoft's internal cryptography policies and implement FIPS 140 compliance.
 
-As of Go 1.25, this feature is enabled by default.
+As of Go 1.25 on Linux and Windows, and Go 1.26 on macOS, this feature is enabled by default on supported platforms.
 
-Prior to Go 1.25, this feature can be enabled by setting the `GOEXPERIMENT` environment variable to `systemcrypto` before building a Go program.
-Other methods of changing goexperiment settings also work.
+As of Go 1.27, `systemcrypto` is no longer configured with `GOEXPERIMENT`: `GOEXPERIMENT=systemcrypto` and `GOEXPERIMENT=nosystemcrypto` are rejected.
+If you need to disable the system-provided cryptography backend, set `MS_GO_NOSYSTEMCRYPTO=1` before building the program.
 
 See [the Migration Guide](/eng/doc/MigrationGuide.md) for more information about when, and how, to configure this feature.
 
--- a/eng/doc/MicrosoftToolsetIdentification.md
+++ b/eng/doc/MicrosoftToolsetIdentification.md
@@ -54,8 +54,10 @@ Otherwise, look for one of these clues:
 
 `systemcrypto` is unique to the Microsoft build of Go, so if either of these strings is present, it confirms that the binary is built by the Microsoft build of Go.
 
-If `GOEXPERIMENT=nosystemcrypto` is present in a binary built with Go 1.26 or earlier (note the `no` prefix), it also confirms that the binary was built with the Microsoft build of Go, but it shows that `systemcrypto` was explicitly disabled.
-In Go 1.27 and later, `systemcrypto` and `nosystemcrypto` are no longer `GOEXPERIMENT` values, so these strings are not expected.
+If `GOEXPERIMENT=nosystemcrypto` is present (note the `no` prefix), it also confirms that the binary was built with the Microsoft build of Go, but it shows that `systemcrypto` was explicitly disabled.
+
+> [!NOTE]
+> In Go 1.27 and later, `systemcrypto` and `nosystemcrypto` are no longer `GOEXPERIMENT` values, so it is expected that `microsoft_systemcrypto=1` will appear without a `GOEXPERIMENT` string.
 
 If none of the above are present, we can't confirm which build of Go was used.
 However, this situation means that the application doesn't meet Microsoft internal crypto policy, which may be enough information in some cases.
--- a/eng/doc/MicrosoftToolsetIdentification.md
+++ b/eng/doc/MicrosoftToolsetIdentification.md
@@ -50,11 +50,12 @@ Then, look for a line that contains `microsoft_toolset_version`.
 Otherwise, look for one of these clues:
 
 * `microsoft_systemcrypto=1`
-* `GOEXPERIMENT=systemcrypto`
+* `GOEXPERIMENT=systemcrypto` in binaries built with Go 1.26 or earlier
 
 `systemcrypto` is unique to the Microsoft build of Go, so if either of these strings is present, it confirms that the binary is built by the Microsoft build of Go.
 
-If `GOEXPERIMENT=nosystemcrypto` is present (note the `no` prefix), it also confirms that the binary was built with the Microsoft build of Go, but it shows that `systemcrypto` was explicitly disabled.
+If `GOEXPERIMENT=nosystemcrypto` is present in a binary built with Go 1.26 or earlier (note the `no` prefix), it also confirms that the binary was built with the Microsoft build of Go, but it shows that `systemcrypto` was explicitly disabled.
+In Go 1.27 and later, `systemcrypto` and `nosystemcrypto` are no longer `GOEXPERIMENT` values, so these strings are not expected.
 
 If none of the above are present, we can't confirm which build of Go was used.
 However, this situation means that the application doesn't meet Microsoft internal crypto policy, which may be enough information in some cases.
--- a/eng/doc/MigrationGuide.md
+++ b/eng/doc/MigrationGuide.md
@@ -119,12 +119,12 @@ That page provides links that redirect to the latest version and also immutable 
 
 ## Enable `systemcrypto`
 
-Starting with Go 1.25 on Linux and Windows, and Go 1.26 on macOS, `systemcrypto` is enabled by default on supported platforms.
-Most projects don't need to set anything to enable it.
+`systemcrypto` is enabled by default on supported platforms in currently supported versions of the Microsoft build of Go.
+You don't need to set `GOEXPERIMENT=systemcrypto`.
 
-
-Starting with Go 1.27, `systemcrypto` is no longer a `GOEXPERIMENT` setting.
-Remove `GOEXPERIMENT=systemcrypto` and `GOEXPERIMENT=nosystemcrypto` from build scripts when moving to Go 1.27 or later; the go command rejects both because system crypto is selected automatically on supported platforms and disabled with `MS_GO_NOSYSTEMCRYPTO=1`.
+Starting with Go 1.27, `systemcrypto` and `nosystemcrypto` are no longer `GOEXPERIMENT` values.
+Remove them from `GOEXPERIMENT` when moving to Go 1.27 or later.
+If you need to opt out of `systemcrypto`, see [Disabling `systemcrypto`](#disabling-systemcrypto).
 
 See [the FIPS documentation sections about build configuration](fips/README.md#usage-common-configurations) for more detailed instructions.
 Even if you don't need FIPS compliance, the `systemcrypto` build configuration instructions are located in that document.
@@ -235,7 +235,7 @@ Alternatively, the `build-essential` package contains all of these packages and 
 
 If this isn't feasible, see [disabling systemcrypto](#disabling-systemcrypto).
 
-#### Unknown or removed GOEXPERIMENT systemcrypto
+#### Unknown GOEXPERIMENT systemcrypto
 
 ```
 go: unknown GOEXPERIMENT systemcrypto
@@ -245,7 +245,19 @@ go: unknown GOEXPERIMENT systemcrypto
 go.exe: unknown GOEXPERIMENT systemcrypto
 ```
 
-In Go 1.27 or later, the Microsoft build of Go reports a different error for the same obsolete build setting:
+This error usually indicates you aren't using the Microsoft build of Go.
+
+If you're trying to migrate to the Microsoft build of Go, check your build environment to ensure that the `go` command is the Microsoft build of Go.
+See [Microsoft Toolset Identification](./MicrosoftToolsetIdentification.md).
+
+If you're trying to make a build command compliant with Microsoft crypto policy but still compatible with **both the Microsoft build of Go and the official Go distribution**, remove `systemcrypto` from `GOEXPERIMENT`.
+`systemcrypto` is enabled by default in the Microsoft build of Go, so it's not necessary to specify it using `GOEXPERIMENT`.
+
+See [Disabling `systemcrypto`](#disabling-systemcrypto) for information about how to disable `systemcrypto` if you need to temporarily avoid migrating to it.
+
+#### Removed GOEXPERIMENT systemcrypto or nosystemcrypto
+
+In Go 1.27 or later, the Microsoft build of Go reports these errors when obsolete `GOEXPERIMENT` values are used:
 
 ```
 GOEXPERIMENT=systemcrypto has been removed; system crypto is enabled automatically on supported platforms and can be disabled with MS_GO_NOSYSTEMCRYPTO=1
@@ -257,15 +269,7 @@ GOEXPERIMENT=nosystemcrypto has been removed; use MS_GO_NOSYSTEMCRYPTO=1 to disa
 
 These errors happen when the `GOEXPERIMENT` environment variable includes `systemcrypto` or `nosystemcrypto`.
 In Go 1.27 and later, remove both values from `GOEXPERIMENT`.
-System crypto is enabled automatically on supported platforms, and it can be disabled with `MS_GO_NOSYSTEMCRYPTO=1` if you have an approved exception.
-
-The `unknown GOEXPERIMENT systemcrypto` form usually indicates you aren't using the Microsoft build of Go.
-
-If you're trying to migrate to the Microsoft build of Go, check your build environment to ensure that the `go` command is the Microsoft build of Go.
-See [Microsoft Toolset Identification](./MicrosoftToolsetIdentification.md).
-
-If you're trying to make a build command compliant with Microsoft crypto policy but still compatible with **both the Microsoft build of Go and the official Go distribution**, remove `systemcrypto` from `GOEXPERIMENT`.
-`systemcrypto` is enabled by default in the Microsoft build of Go, so it's not necessary to specify it using `GOEXPERIMENT`.
+The `systemcrypto` backend is enabled automatically on supported platforms, and it can be disabled with `MS_GO_NOSYSTEMCRYPTO=1` if you have an approved exception.
 
 See [Disabling `systemcrypto`](#disabling-systemcrypto) for information about how to disable `systemcrypto` if you need to temporarily avoid migrating to it.
 
--- a/eng/doc/MigrationGuide.md
+++ b/eng/doc/MigrationGuide.md
@@ -251,7 +251,7 @@ GOEXPERIMENT=systemcrypto has been removed; system crypto is enabled automatical
 ```
 
 ```
-GOEXPERIMENT=nosystemcrypto has been removed; use MS_GO_NOSYSTEMCRYPTO=1 to disable system crypto
+GOEXPERIMENT=nosystemcrypto has been removed; use MS_GO_NOSYSTEMCRYPTO=1 to disable system crypto; note that systemcrypto supports CGO_ENABLED=0 since Go 1.27
 ```
 
 These errors happen when the `GOEXPERIMENT` environment variable includes `systemcrypto` or `nosystemcrypto`.
--- a/eng/doc/MigrationGuide.md
+++ b/eng/doc/MigrationGuide.md
@@ -169,19 +169,20 @@ In Go 1.25, you may see this similar error:
 
 > [!NOTE]
 > In Go 1.26, there is a cgo-less experiment available for Linux: `ms_nocgo_opensslcrypto`.
-> In Go 1.27 and later, this behavior is part of `systemcrypto` and is selected automatically when cgo is disabled on a supported Linux architecture.
+>
+> In Go 1.27 and later, cgo-less behavior is part of `systemcrypto` and is selected automatically when cgo is disabled on a supported Linux architecture.
 >
 > For more details, see [cgo-less OpenSSL Backend](NocgoOpenSSL.md).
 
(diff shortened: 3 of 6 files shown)


eng/doc/fips/README.md
```diff
@@ -72,27 +72,28 @@ The following table summarizes common configurations and how suitable each one i
 
 > [!NOTE]
 > Since Go 1.25, `systemcrypto` is enabled by default on Linux and Windows. There is no need to manually enable using OpenSSL/CNG under the hood anymore. See also [the Go 1.25 changelog](#go-125-aug-2025). Since Go 1.26, `systemcrypto` is also enabled by default on macOS.
+> Since Go 1.27, `systemcrypto` is no longer a `GOEXPERIMENT` setting. Supported platforms use it automatically unless it is disabled with `MS_GO_NOSYSTEMCRYPTO=1`.
```
dagood: It seems to me that it's reasonable to put these all in the same note, but they should have always had their own start-of-line IMO to be easy to spot.

Also, changelog ref.

```suggestion
> Since Go 1.25, `systemcrypto` is enabled by default on Linux and Windows. There is no need to manually enable using OpenSSL/CNG under the hood anymore.
> See also [the Go 1.25 changelog](#go-125-aug-2025).
>
> Since Go 1.26, `systemcrypto` is also enabled by default on macOS.
>
> Since Go 1.27, `systemcrypto` is no longer a `GOEXPERIMENT` setting. Supported platforms use it automatically unless it is disabled with `MS_GO_NOSYSTEMCRYPTO=1`.
> See also [the Go 1.27 changelog](#go-127-aug-2026).
```

eng/doc/fips/README.md
```diff
@@ -72,27 +72,28 @@ The following table summarizes common configurations and how suitable each one i
...
 | `-tags=requirefips` | Default | Compliant | Can be used to create a compliant app. The behavior is the same as `GODEBUG=fips140=on` and `GOFIPS=1`, but no runtime configuration is necessary. See [the `requirefips` section](#build-option-to-require-fips-mode) for more information on when this "locked-in" approach may be useful rather than the flexible approach. |
-| `MS_GO_NOSYSTEMCRYPTO=1` or `GOEXPERIMENT=nosystemcrypto` | Default | Not compliant | Crypto usage is not FIPS compliant. |
-| `GOOS=linux CGO_ENABLED=0 GOEXPERIMENT=ms_nocgo_opensslcrypto`, Go 1.26 | Default | Compliant | [Experimental](https://github.com/microsoft/go/blob/microsoft/main/eng/doc/NocgoOpenSSL.md). Crypto usage is FIPS compliant. |
+| `MS_GO_NOSYSTEMCRYPTO=1` | Default | Not compliant | Crypto usage is not FIPS compliant. |
+| `GOOS=linux CGO_ENABLED=0`, Go 1.27+ | Default | Compliant | Uses the [cgo-less OpenSSL backend](../NocgoOpenSSL.md) on supported architectures. Crypto usage is FIPS compliant. |
```
dagood: "Crypto usage is FIPS compliant" can be misleading vs. the "Can be used to create a compliant app" language, because not all crypto usage would be FIPS compliant.

eng/doc/fips/README.md
```diff
@@ -103,10 +104,10 @@ See the [Migration Guide](/eng/doc/MigrationGuide.md) for more information on in
...
 
 > [!NOTE]
-> "Experiment" doesn't indicate the FIPS features are experimental. The original intent of `GOEXPERIMENT` is to use it to enable experimental features in the Go runtime and toolchain, but we and Google are now using `GOEXPERIMENT` for this FIPS-related feature because the mechanism itself perfectly fits our needs.
+> Prior to Go 1.27, `systemcrypto` could be selected with `GOEXPERIMENT=systemcrypto`. In Go 1.27 and later, `systemcrypto` is no longer a `GOEXPERIMENT` setting. It is enabled automatically on supported platforms, `go env GOEXPERIMENT` doesn't report it, and `GOEXPERIMENT=systemcrypto` is rejected.
 
 > [!NOTE]
-> Prior to Go 1.27, per-platform experiments (`opensslcrypto`, `cngcrypto`, `darwincrypto`) were available. These have been removed in Go 1.27. Use `systemcrypto` instead, which automatically selects the appropriate backend based on the target platform.
+> Prior to Go 1.27, per-platform experiments (`opensslcrypto`, `cngcrypto`, `darwincrypto`) were available. These have been removed in Go 1.27. The build tags associated with those names remain available for source compatibility when `systemcrypto` is enabled.
```
dagood: "Available" seems like an odd wording:


```suggestion
> Prior to Go 1.27, per-platform experiments (`opensslcrypto`, `cngcrypto`, `darwincrypto`) were available. These experiments have been removed in Go 1.27, but the build tag associated with each experiment remains supported, ensuring source compatibility.
```

eng/doc/fips/README.md
```diff
@@ -119,19 +120,20 @@ The `systemcrypto` experiment uses platform-specific code via build constraints.
 In a cross-build scenario, such as using Linux to build an app that will run on Windows, `GOOS=windows` will correctly use CNG-based code for the `systemcrypto` backend.
 
 A cross-build to Windows or macOS will typically work, because these backends use approaches like syscalls to call the crypto library rather than cgo.
-A cross-build to Linux, however, is more complicated (perhaps infeasible), because its backend uses cgo.
+On Linux in Go 1.27 and later, `systemcrypto` can also be used when cgo is disabled by selecting the [cgo-less OpenSSL backend](../NocgoOpenSSL.md) on supported architectures.
```
dagood: Split up for attention, remove clutter to make it flow better, reduce frights (we support way more architectures than I'd expect):


```suggestion

A cross-build to Linux in Go 1.27 and later will work if the [cgo-less OpenSSL backend](../NocgoOpenSSL.md) is used.
The cgo-less backend may be unavailable for some processor architectures that aren't commonly used at Microsoft.
```

eng/doc/fips/README.md
```diff
@@ -305,19 +307,16 @@ It's possible to disable `systemcrypto` and use the Go standard library's implem
...
-- `MS_GO_NOSYSTEMCRYPTO=1` doesn't involve the `GOEXPERIMENT` mechanism. It's simpler to use and to incorporate into any build process.
+- `MS_GO_NOSYSTEMCRYPTO=1` doesn't involve the `GOEXPERIMENT` mechanism. It's simple to use and to incorporate into any build process.
+- Only the exact value `1` disables `systemcrypto`. Other values leave the default behavior in place.
 
 > [!WARNING]
-> `MS_GO_NOSYSTEMCRYPTO=1` has precedence over `GOEXPERIMENT` values.
-> It will disable the backend even if `GOEXPERIMENT=systemcrypto` is set.
->
-> Specifically, `MS_GO_NOSYSTEMCRYPTO=1 GOEXPERIMENT=systemcrypto go build .` builds a program that uses Go standard library cryptography.
+> Go 1.27 and later reject `GOEXPERIMENT=systemcrypto` and `GOEXPERIMENT=nosystemcrypto` with an error. Use `MS_GO_NOSYSTEMCRYPTO=1` to disable the backend.
```
dagood, line 335: Too early to remove this info about 1.25 and 1.26 IMO.

eng/doc/fips/README.md
```diff
@@ -344,18 +343,21 @@ This algorithm can be overridden by setting the environment variable `GO_OPENSSL
...
 For more information about other Go experiments, read the output of the command `go doc goexperiment.Flags` to see the experiments available in your specific build of the Go toolset, or check [the online goexperiment package doc](https://pkg.go.dev/internal/goexperiment) to see the options for other versions.
 
 ### Build tags
 
-Selecting most `GOEXPERIMENT`s can also be done by setting the corresponding `goexperiment.*` build tag. This is supported for all crypto backends.
-
-For example, `go build -tags=goexperiment.systemcrypto` command will enable the same backend as setting `GOEXPERIMENT=systemcrypto` then running the build command.
+When `systemcrypto` is enabled, the Microsoft build of Go emits the `goexperiment.systemcrypto` build tag.
+It also emits one legacy per-platform tag for source compatibility: `goexperiment.opensslcrypto` on Linux, `goexperiment.cngcrypto` on Windows, and `goexperiment.darwincrypto` on macOS.
+These tags reflect the backend selected by the toolchain; they are not the supported way to enable or disable `systemcrypto`.
```
dagood: Wrong section for this info, next section "Conditional behavior if a crypto backend is enabled" is about this.

eng/doc/fips/README.md
```diff
@@ -344,18 +343,21 @@ This algorithm can be overridden by setting the environment variable `GO_OPENSSL
...
+
+You can still use `GOEXPERIMENT` for other toolchain experiments, using a comma separator when enabling multiple values.
 
 For more information about other Go experiments, read the output of the command `go doc goexperiment.Flags` to see the experiments available in your specific build of the Go toolset, or check [the online goexperiment package doc](https://pkg.go.dev/internal/goexperiment) to see the options for other versions.
 
 ### Build tags
 
-Selecting most `GOEXPERIMENT`s can also be done by setting the corresponding `goexperiment.*` build tag. This is supported for all crypto backends.
-
-For example, `go build -tags=goexperiment.systemcrypto` command will enable the same backend as setting `GOEXPERIMENT=systemcrypto` then running the build command.
```
dagood, line 355: Doesn't some of this still apply to 1.26 and shouldn't be deleted yet?

eng/doc/fips/README.md
```diff
@@ -460,9 +462,15 @@ This list of major changes is intended for quick reference and for access to his
...
 - The per-platform GOEXPERIMENTs (`opensslcrypto`, `cngcrypto`, `darwincrypto`) have been removed.
   - Using any of the removed experiments will result in a build error.
-  - The `systemcrypto` GOEXPERIMENT has been the preferred way to select a crypto backend since it was introduced in Go 1.21. It is now the only way.
   - The build tags associated with the removed GOEXPERIMENTs remain supported for legacy source compatibility.
   - The `goexperiment.systemcrypto` build tag remains supported, and its behavior has not changed.
+- `GOEXPERIMENT=systemcrypto` and `GOEXPERIMENT=nosystemcrypto` have been removed.
+  - System crypto is enabled automatically on supported platforms.
+  - To disable system crypto, set `MS_GO_NOSYSTEMCRYPTO=1`.
+  - `systemcrypto` is no longer included in `go env GOEXPERIMENT`, `goexperiment.Flags`, or other GOEXPERIMENT-derived output.
+  - The `goexperiment.systemcrypto` build tag is still emitted when system crypto is enabled.
```
dagood: We don't call it "system crypto":

```suggestion
  - `systemcrypto` is enabled automatically on supported platforms.
  - To disable `systemcrypto`, set `MS_GO_NOSYSTEMCRYPTO=1`.
  - `systemcrypto` is no longer included in `go env GOEXPERIMENT`, `goexperiment.Flags`, or other GOEXPERIMENT-derived output.
  - The `goexperiment.systemcrypto` build tag is still emitted when `systemcrypto` is enabled.
```

"System crypto backend" could also fit the last point well:


```suggestion
  - `systemcrypto` is enabled automatically on supported platforms.
  - To disable `systemcrypto`, set `MS_GO_NOSYSTEMCRYPTO=1`.
  - `systemcrypto` is no longer included in `go env GOEXPERIMENT`, `goexperiment.Flags`, or other GOEXPERIMENT-derived output.
  - The `goexperiment.systemcrypto` build tag is still emitted when a system crypto backend is enabled.
```

eng/doc/fips/README.md
```diff
@@ -491,12 +499,12 @@ This list of major changes is intended for quick reference and for access to his
 
 ### Go 1.25 (Aug 2025)
 
-- The `systemcrypto` goexperiment is now enabled by default on Windows and Linux. To disable it, set `GOEXPERIMENT=nosystemcrypto`.
+- The `systemcrypto` goexperiment is now enabled by default on Windows and Linux. In current releases, disable it with `MS_GO_NOSYSTEMCRYPTO=1`; in early Go 1.25 releases, `GOEXPERIMENT=nosystemcrypto` was the available disable knob.
```
dagood: I think it's ok to revise this, but it'd be better to link rather than keep each note up to date individually:


```suggestion
- The `systemcrypto` goexperiment is now enabled by default on Windows and Linux. In Go 1.25.0, `GOEXPERIMENT=nosystemcrypto` was the available disable knob, but the current supported disable knob is documented by the [Build option to use Go crypto](#build-option-to-use-go-crypto) section.
```

eng/doc/fips/README.md
```diff
@@ -505,7 +513,7 @@ This list of major changes is intended for quick reference and for access to his
...
+- `GOEXPERIMENT=allowcryptofallback` has been removed. Instead, if it's necessary to opt out from using a system crypto backend, use `MS_GO_NOSYSTEMCRYPTO=1` in current releases. This is an internal mechanism that is not intended for use when building a Go application. This document has always recommended against using it, so we anticipate that this change won't affect users of the Microsoft build of Go. Please [contact the maintainers of the Microsoft build of Go](https://github.com/microsoft/go/blob/microsoft/main/SUPPORT.md) if you need to use it so we can understand the scenario and help find a safer alternative.
```
dagood: (Same as above, also clean up bad "this".)

```suggestion
- `GOEXPERIMENT=allowcryptofallback` has been removed. Instead, if it's necessary to opt out from using a system crypto backend, use the disable knob documented by the [Build option to use Go crypto](#build-option-to-use-go-crypto) section. `allowcryptofallback` is an internal mechanism that is not intended for use when building a Go application. This document has always recommended against using it, so we anticipate that this change won't affect users of the Microsoft build of Go. Please [contact the maintainers of the Microsoft build of Go](https://github.com/microsoft/go/blob/microsoft/main/SUPPORT.md) if you need to use it so we can understand the scenario and help find a safer alternative.
```

eng/doc/AdditionalFeatures.md
```diff
@@ -19,10 +19,10 @@ See also the [Data Collection policy for the Microsoft build of Go](/README.md#d
 We have implemented changes to the Go standard library to call a platform-specific system-provided cryptography library rather than running the pure Go implementation.
 Its main purposes are to comply with Microsoft's internal cryptography policies and implement FIPS 140 compliance.
 
-As of Go 1.25, this feature is enabled by default.
+As of Go 1.25 on Linux and Windows, and Go 1.26 on macOS, this feature is enabled by default on supported platforms.
```
dagood: This adds too much to both sides... specifying both the detailed list of platforms and saying "on supported platforms" implies that there are some Linux/Windows/macOS platforms that still aren't supported.

We can make this very clean now that 1.25 is the oldest in support:

```suggestion
This feature is enabled by default on supported platforms.
```

eng/doc/AdditionalFeatures.md
```diff
@@ -19,10 +19,10 @@ See also the [Data Collection policy for the Microsoft build of Go](/README.md#d
...
 
-Prior to Go 1.25, this feature can be enabled by setting the `GOEXPERIMENT` environment variable to `systemcrypto` before building a Go program.
-Other methods of changing goexperiment settings also work.
+As of Go 1.27, `systemcrypto` is no longer configured with `GOEXPERIMENT`: `GOEXPERIMENT=systemcrypto` and `GOEXPERIMENT=nosystemcrypto` are rejected.
+If you need to disable the system-provided cryptography backend, set `MS_GO_NOSYSTEMCRYPTO=1` before building the program.
```
dagood: I think it'd be better to simply remove this. It's clear where to go to figure out how to configure this, and it'd be better to filter through the migration guide which may help them decide not to disable it.

```suggestion
```

eng/doc/MicrosoftToolsetIdentification.md
```diff
@@ -50,11 +50,12 @@ Then, look for a line that contains `microsoft_toolset_version`.
 Otherwise, look for one of these clues:
 
 * `microsoft_systemcrypto=1`
-* `GOEXPERIMENT=systemcrypto`
+* `GOEXPERIMENT=systemcrypto` in binaries built with Go 1.26 or earlier
 
 `systemcrypto` is unique to the Microsoft build of Go, so if either of these strings is present, it confirms that the binary is built by the Microsoft build of Go.
 
-If `GOEXPERIMENT=nosystemcrypto` is present (note the `no` prefix), it also confirms that the binary was built with the Microsoft build of Go, but it shows that `systemcrypto` was explicitly disabled.
+If `GOEXPERIMENT=nosystemcrypto` is present in a binary built with Go 1.26 or earlier (note the `no` prefix), it also confirms that the binary was built with the Microsoft build of Go, but it shows that `systemcrypto` was explicitly disabled.
```
dagood: Unnecessary specificity (does the reader really need to consider `GOEXPERIMENT=nosystemcrypto` appearing in a Go 1.27+ binary?) makes it hard to piece together the logic when reading:

```suggestion
If `GOEXPERIMENT=nosystemcrypto` is present (note the `no` prefix), it also confirms that the binary was built with the Microsoft build of Go, but it shows that `systemcrypto` was explicitly disabled.
```

 If someone's suprised it's not present, the following sentence clears it up nicely.

eng/doc/MicrosoftToolsetIdentification.md
```diff
@@ -50,11 +50,12 @@ Then, look for a line that contains `microsoft_toolset_version`.
...
+In Go 1.27 and later, `systemcrypto` and `nosystemcrypto` are no longer `GOEXPERIMENT` values, so these strings are not expected.
```
dagood: The last time "these strings" was mentioned, it included `microsoft_systemcrypto=1`, so this is misleading. Also, this doesn't seem to have much to do with the last sentence directly, and just clarifies that it's ok when something appears to maybe be missing, so a note:

```suggestion

> [!NOTE]
> In Go 1.27 and later, `systemcrypto` and `nosystemcrypto` are no longer `GOEXPERIMENT` values, so it is expected that `microsoft_systemcrypto=1` will appear without a `GOEXPERIMENT` string.
```

eng/doc/MigrationGuide.md
```diff
@@ -119,14 +119,15 @@ That page provides links that redirect to the latest version and also immutable
 
 ## Enable `systemcrypto`
 
-These instructions are for projects using versions of Go that don't enable `systemcrypto` by default.
-Starting with 1.25 (Linux and Windows) and 1.26 (macOS), `systemcrypto` is enabled by default.
+Starting with Go 1.25 on Linux and Windows, and Go 1.26 on macOS, `systemcrypto` is enabled by default on supported platforms.
```
dagood: Same unclear combination as AdditionalFeatures.md

eng/doc/MigrationGuide.md
```diff
@@ -119,14 +119,15 @@ That page provides links that redirect to the latest version and also immutable
...
+Most projects don't need to set anything to enable it.
```
dagood: Most? How would a project set something to enable it, or see that they need to? Not sure this uncertainty is needed. Or is this referring to build errors?

eng/doc/MigrationGuide.md
```diff
@@ -119,14 +119,15 @@ That page provides links that redirect to the latest version and also immutable
...
 
-To comply with Microsoft internal cryptography policy, enable the `systemcrypto` feature in your build environment before building your project.
-This is done by setting the `GOEXPERIMENT` environment variable to `systemcrypto`.
+
+Starting with Go 1.27, `systemcrypto` is no longer a `GOEXPERIMENT` setting.
+Remove `GOEXPERIMENT=systemcrypto` and `GOEXPERIMENT=nosystemcrypto` from build scripts when moving to Go 1.27 or later; the go command rejects both because system crypto is selected automatically on supported platforms and disabled with `MS_GO_NOSYSTEMCRYPTO=1`.
```
dagood: Wording seems odd and rambles a little, maybe replace with link to FIPS doc and/or the later section in this doc about disabling?

eng/doc/MigrationGuide.md
```diff
@@ -167,25 +169,28 @@ In Go 1.25, you may see this similar error:
 
 > [!NOTE]
 > In Go 1.26, there is a cgo-less experiment available for Linux: `ms_nocgo_opensslcrypto`.
+> In Go 1.27 and later, this behavior is part of `systemcrypto` and is selected automatically when cgo is disabled on a supported Linux architecture.
```
dagood:
```suggestion
>
> In Go 1.27 and later, cgo-less behavior is part of `systemcrypto` and is selected automatically when cgo is disabled on a supported Linux architecture.
```

eng/doc/MigrationGuide.md
```diff
@@ -167,25 +169,28 @@ In Go 1.25, you may see this similar error:
...
 >
 > For more details, see [cgo-less OpenSSL Backend](NocgoOpenSSL.md).
 
 When building a Go program that imports a `crypto` package (or has a dependency that imports a `crypto` package), the build will check that the build environment and target are compatible with the crypto backend being used.
 If it's incompatible, the build will fail with an error like the above.
 
-When targeting Linux, `systemcrypto` requires cgo.
+In Go 1.26 and earlier, targeting Linux with `systemcrypto` usually requires cgo unless the `ms_nocgo_opensslcrypto` experiment is enabled.
```
dagood: Avoid corner case of having systemcrypto enabled without using any crypto code: "using" more strongly implies crypto calls:

```suggestion
In Go 1.26 and earlier, using `systemcrypto` on Linux requires cgo unless the `ms_nocgo_opensslcrypto` experiment is enabled.
```

The earlier paragraph also mentions importing `crypto` packages.

eng/doc/MigrationGuide.md
```diff
@@ -167,25 +169,28 @@ In Go 1.25, you may see this similar error:
...
 Cgo is disabled by default on some platforms or when a C compiler is not detected.
 Sometimes a project's build scripts might explicitly disable cgo.
 
-In this case, you should first try to install a C compiler, like `gcc`.
+In this case, you should first try upgrading to Go 1.27 or later.
+If you need or want the cgo-based OpenSSL backend, install a C compiler, like `gcc`.
```
dagood: Some projects require specific Gos, unfortunately:

```suggestion
If you need or want the cgo-based OpenSSL backend or the pre-1.27 Go version, install a C compiler, like `gcc`.
```

eng/doc/MigrationGuide.md
```diff
@@ -239,14 +244,27 @@ go: unknown GOEXPERIMENT systemcrypto
 go.exe: unknown GOEXPERIMENT systemcrypto
 ```
 
-This error indicates you aren't using the Microsoft build of Go.
-It happens when the `GOEXPERIMENT` environment variable includes `systemcrypto` (or `nosystemcrypto`) and the Go toolset doesn't recognize it.
+In Go 1.27 or later, the Microsoft build of Go reports a different error for the same obsolete build setting:
+
+```
+GOEXPERIMENT=systemcrypto has been removed; system crypto is enabled automatically on supported platforms and can be disabled with MS_GO_NOSYSTEMCRYPTO=1
```
dagood: This seems like a distinct problem from unknown GOEXPERIMENT systemcrypto with a very different cause and solution. A new section makes more sense to me.

eng/doc/NocgoOpenSSL.md
```diff
@@ -1,21 +1,21 @@
...
-In Go 1.26, there is a cgo-less experiment available for Linux: `ms_nocgo_opensslcrypto`.
+In Go 1.27 and later, the cgo-less OpenSSL backend is part of `systemcrypto` on Linux.
+It is selected automatically when cgo is disabled and the target architecture is supported.
 
 > [!NOTE]
-> While `systemcrypto` is a fully supported `GOEXPERIMENT` value (it is not "experimental"), `ms_nocgo_opensslcrypto` **is** experimental in Go 1.26 and may have limitations.
-
-In Go 1.27, the experiment will be removed and the cgo requirement for `systemcrypto` on Linux will be lifted by default.
+> In Go 1.26, this backend was available as the experimental `GOEXPERIMENT=ms_nocgo_opensslcrypto` feature.
+> In Go 1.27, that experiment has been removed because the cgo-less backend is selected automatically when needed.
```
dagood, line 13:
```suggestion
>
> In Go 1.27, that experiment has been removed because the cgo-less backend is selected automatically when needed.
```

eng/doc/Telemetry.md
```diff
@@ -55,7 +55,7 @@ The counters and properties collected are defined in the [telemetry upload confi
 
 | Counter | Description |
 |---|---|
-| `msgo/systemcrypto:{enabled,disabled}` | Whether the system cryptography backend is enabled on platforms that support it. |
+| `msgo/systemcrypto:{enabled,disabled}` | Whether the system cryptography backend is enabled on platforms that support it. This reflects the default platform selection and the `MS_GO_NOSYSTEMCRYPTO=1` opt-out. |
```
dagood: Is it possible to simplify? I don't see what detail I get as a reader that isn't obvious from:


```suggestion
| `msgo/systemcrypto:{enabled,disabled}` | Whether the system cryptography backend is enabled. |
```

dagood (changes requested):

dagood (approved):
Thanks!

Trustinbtc999-hue (commented):
New

PR #2342, opened 2026-06-02
Wire ML-DSA into the crypto backend

### LINKED ISSUE #2293: Patch mldsa

michelle-clayton-work:
Wires ML-DSA into the system crypto backend, mirroring the existing ML-KEM integration. When the backend is enabled and supports the requested parameter set, `crypto/mldsa` dispatches key generation, parsing, hedged signing, and verification to the native backend (OpenSSL / CNG / CryptoKit); otherwise it falls back to the Go FIPS 140-3 module.

- `crypto/internal/backend`: add ML-DSA shims for `backend_linux.go` (OpenSSL), `backend_windows.go` (CNG), `backend_darwin.go` (CryptoKit), and panic stubs in `nobackend.go`.
- `crypto/mldsa/mldsa_fips140v1.26.go`: add a `boring` backend handle to `PrivateKey`/`PublicKey`, a `useBoringMLDSA` gate, and backend dispatch for `GenerateKey`, `NewPrivateKey`, `newPublicKey`, `Bytes`, `PublicKey`, `Equal`, `Parameters`, hedged `Sign`, and `Verify`.

### DIFF
--- a/eng/_util/cmd/updatecryptodocs/docs.go
+++ b/eng/_util/cmd/updatecryptodocs/docs.go
@@ -266,7 +266,7 @@ var doc = Document{
 		},
 		{
 			Title:       "Asymmetric encryption",
-			Packages:    []string{"RSA", "ECDSA", "ECDH", "Ed25519", "DSA"},
+			Packages:    []string{"RSA", "ECDSA", "ECDH", "Ed25519", "ML-DSA", "DSA"},
 			Description: "",
 			Subsections: []Section{
 				{
@@ -482,6 +482,39 @@ var doc = Document{
 								Windows: PlatformStatus{Supported: NotSupported},
 								Linux:   PlatformStatus{Supported: NotSupported},
 								MacOS:   PlatformStatus{Supported: NotSupported},
+							},
+						},
+					},
+				},
+				{
+					Title:        "ML-DSA",
+					ColumnHeader: "Parameters",
+					Packages:     []string{"crypto/mldsa"},
+					MinGoVersion: "1.27",
+					Description:  "Operations that require random numbers (rand io.Reader) only support [rand.Reader](https://pkg.go.dev/crypto/rand#Reader). Deterministic signing is always performed by the Go cryptographic module. External-mu signing falls back to the Go cryptographic module on platforms whose backend does not implement it (currently macOS).",
+					Items: []Item{
+						{
+							Name: "44",
+							Platforms: Platforms{
+								Windows: PlatformStatus{MinVersion: "11 (24H2)"},
+								Linux:   PlatformStatus{MinVersion: "3.5.0"},
+								MacOS:   PlatformStatus{Supported: NotSupported},
+							},
+						},
+						{
+							Name: "65",
+							Platforms: Platforms{
+								Windows: PlatformStatus{MinVersion: "11 (24H2)"},
+								Linux:   PlatformStatus{MinVersion: "3.5.0"},
+								MacOS:   PlatformStatus{MinVersion: "26"},
+							},
+						},
+						{
+							Name: "87",
+							Platforms: Platforms{
+								Windows: PlatformStatus{MinVersion: "11 (24H2)"},
+								Linux:   PlatformStatus{MinVersion: "3.5.0"},
+								MacOS:   PlatformStatus{MinVersion: "26"},
 							},
 						},
 					},
--- a/patches/0002-Add-crypto-backends.patch
+++ b/patches/0002-Add-crypto-backends.patch
@@ -107,8 +107,8 @@ Subject: [PATCH] Add crypto backends
  src/crypto/internal/rand/rand.go              |   2 +-
  src/crypto/md5/md5.go                         |  10 +
  src/crypto/md5/md5_test.go                    |  18 +-
- src/crypto/mldsa/mldsa_fips140v1.26.go        | 173 ++++++-
- src/crypto/mldsa/mldsa_test.go                |  64 ++-
+ src/crypto/mldsa/mldsa_fips140v1.26.go        | 175 ++++++-
+ src/crypto/mldsa/mldsa_test.go                |  65 ++-
  src/crypto/mlkem/mlkem.go                     | 118 ++++-
  src/crypto/mlkem/mlkem_test.go                |   8 +
  src/crypto/pbkdf2/pbkdf2.go                   |   7 +
@@ -167,7 +167,7 @@ Subject: [PATCH] Add crypto backends
  src/net/lookup_test.go                        |   3 +
  src/os/exec/exec_test.go                      |   9 +
  src/runtime/runtime_boring.go                 |   5 +
- 163 files changed, 5288 insertions(+), 320 deletions(-)
+ 163 files changed, 5292 insertions(+), 319 deletions(-)
  create mode 100644 src/cmd/go/internal/modindex/build_test.go
  create mode 100644 src/cmd/go/systemcrypto_test.go
  create mode 100644 src/crypto/dsa/boring.go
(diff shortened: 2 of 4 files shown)


gdams:
You'll need to update https://github.com/microsoft/go/blob/microsoft/main/eng/_util/cmd/updatecryptodocs/docs.go and run the generator to add ML-DSA to the crypto docs.

patches/0002-Add-crypto-backends.patch
```diff
@@ -5644,6 +5801,359 @@ index 403ff2881f4b68..5417556a86700e 100644
...
+ 	"errors"
+ 	"io"
+ )
+@@ -18,13 +21,55 @@ import (
+ //
+ // A PrivateKey is safe for concurrent use.
+ type PrivateKey struct {
+-	k mldsa.PrivateKey
++	k      mldsa.PrivateKey
++	boring *boring.PrivateKeyMLDSA
```
qmuntal: Can we keep `boring` by value instead of by reference? Same a `k`.

patches/0002-Add-crypto-backends.patch
```diff
@@ -5644,6 +5801,359 @@ index 403ff2881f4b68..5417556a86700e 100644
...
+ 	default:
+ 		return nil, errInvalidSignerOpts
+ 	}
+@@ -145,7 +260,8 @@ func (sk *PrivateKey) SignDeterministic(message []byte, opts crypto.SignerOpts)
+ //
+ // A PublicKey is safe for concurrent use.
+ type PublicKey struct {
+-	p mldsa.PublicKey
++	p      mldsa.PublicKey
++	boring *boring.PublicKeyMLDSA
```
qmuntal, line 6091: Can we keep `boring` by value instead of by reference? Same a `k`.

patches/0002-Add-crypto-backends.patch
```diff
@@ -5644,6 +5801,359 @@ index 403ff2881f4b68..5417556a86700e 100644
...
+ 	switch params {
+@@ -75,11 +127,21 @@ func (sk *PrivateKey) Equal(x crypto.PrivateKey) bool {
+ 	if !ok || other == nil {
+ 		return false
+ 	}
++	if sk.boring != nil || other.boring != nil {
++		// Bytes returns only the 32-byte seed, which is the same length for all
++		// parameter sets, so the parameter sets must be compared separately to
++		// match the standard library Equal, which distinguishes them.
++		return sk.paramSet() == other.paramSet() &&
```
qmuntal: Hmm, think should Go into the backends. As a general rule (which has exceptions), we shouldn't implement in this layer operations that upstream implements in `internal/fips140`. For example, OpenSSL support [EVP_PKEY_eq](https://docs.openssl.org/master/man3/EVP_PKEY_eq). I'm not sure ifCNG/CryptoKit provide similar helpers. If they don't, then we can always fallback to comparing the bytes in the backend.

patches/0002-Add-crypto-backends.patch
```diff
@@ -5644,6 +5801,359 @@ index 403ff2881f4b68..5417556a86700e 100644
...
++	}
++	return sk.k.PublicKey().Parameters()
++}
++
++// goKey returns a standard library private key equivalent to sk. If sk is
++// backed by the crypto backend, the key is reconstructed from its seed. It is
++// used for operations the backend does not support across all platforms, such
++// as deterministic and external-mu signing, which then run through the FIPS
++// 140-3 Go Cryptographic Module.
++func (sk *PrivateKey) goKey() (*mldsa.PrivateKey, error) {
```
qmuntal, line 6001: Calling NewPrivate[44,65,87] is a bit slow. We could set `sk.k` the first time goKey is used so that the next call reuses the same upstream key.

patches/0002-Add-crypto-backends.patch
```diff
@@ -5644,6 +5801,359 @@ index 403ff2881f4b68..5417556a86700e 100644
...
+ 	//
+@@ -167,6 +205,12 @@ func TestAllocations(t *testing.T) {
+ 	//
+ 	// Still, check we are not slipping more allocations in.
+ 	var expected float64 = 7
++	if boring.Enabled {
++		// With a crypto backend enabled, key generation, parsing, signing, and
++		// verification may route through the backend, whose allocation behavior
++		// differs by platform and parameter set.
++		t.Skip("skipping allocations test with crypto backend enabled")
```
qmuntal: Don't skip allocation tests. It's better to adjust the allocation expectations when systemcrypto is enabled. This way we catch regressions.

qmuntal (changes requested):

patches/0002-Add-crypto-backends.patch
```diff
@@ -5644,6 +5801,359 @@ index 403ff2881f4b68..5417556a86700e 100644
...
+ 			context = opts.Context
+ 		}
++		if sk.boring != nil {
++			return sk.boring.Sign(message, context)
++		}
+ 		return mldsa.Sign(&sk.k, message, context)
+ 	case crypto.MLDSAMu:
+-		return mldsa.SignExternalMu(&sk.k, message)
++		// External-mu signing is not supported by every crypto backend, so use
++		// the standard library implementation.
```
dagood: Shouldn't this be a capability check, in that case? ("Every" implying that some backend(s) do support it.)

patches/0002-Add-crypto-backends.patch
dagood: In theory this block implements compatibility between boring `sk` and Go `other` and vice versa, it sounds like there should be a new initial "both boring" case that calls into the backend to implement this (if supported) rather than using `subtle.ConstantTimeCompare`?

patches/0002-Add-crypto-backends.patch
michelle-clayton-work: Updated — the test runs in both modes now

patches/0002-Add-crypto-backends.patch
michelle-clayton-work: Yeah that makes sense. I made it a capability check using the backends SignExternalMu where possible

eng/_util/cmd/updatecryptodocs/docs.go
```diff
@@ -486,6 +486,39 @@ var doc = Document{
 						},
 					},
 				},
+				{
+					Title:        "ML-DSA",
+					ColumnHeader: "Parameters",
+					Packages:     []string{"crypto/mldsa"},
+					MinGoVersion: "1.26",
```
dagood: Is it true that we're planning to backport this from 1.27?
(If so, we should hold off on the doc change until we actually do.)

gdams: I don't think it's possible to backport this as it only landed in gotip. Let's update this to be 1.27

gdams:
```suggestion
					MinGoVersion: "1.27",
```

patches/0002-Add-crypto-backends.patch
michelle-clayton-work: - https://github.com/microsoft/go-crypto-darwin/pull/214
- https://github.com/microsoft/go-crypto-winnative/pull/182
- https://github.com/microsoft/go-crypto-openssl/pull/25

michelle-clayton-work: - https://github.com/microsoft/go/pull/2353 should go in first

patches/0002-Add-crypto-backends.patch
```diff
@@ -5681,6 +5863,419 @@ index 403ff2881f4b68..5417556a86700e 100644
...
+ 	}
+ 	cryptotest.SkipTestAllocations(t)
++	// ML-DSA-65 is implemented by every crypto backend. When one is enabled, key
++	// generation, parsing, signing, and verification route through the backend,
++	// whose allocation profile differs by platform and backend implementation.
++	params := MLDSA65()
++	if boring.Enabled && boring.SupportsMLDSA(boring.MLDSA65()) {
++		switch runtime.GOOS {
++		case "darwin", "windows":
++			expected = 16
```
qmuntal: On my Windows machine, which supports ML-DSA, this tests only produces 8 allocations. 16 seems is too conservative.

patches/0002-Add-crypto-backends.patch
```diff
@@ -5681,6 +5863,419 @@ index 403ff2881f4b68..5417556a86700e 100644
...
++		case "linux":
++			// OpenSSL-backed ML-DSA may allocate extra wrapper/import objects.
++			expected = 32
```
qmuntal: 32 allocations seems like too much. Worth investigating before merging.

qmuntal (changes requested):

qmuntal (approved):
LGTM!

PR #2344, opened 2026-06-03
Add copilot instructions for crypto backend implementation

gdams:
Adds a `.github/copilot-instructions.md` that captures the architecture patterns, integration conventions, and review guidance for implementing crypto algorithms in the Microsoft Go backend system.

The instructions were derived from:
- **Patch 0002** architecture (backend shim files, nobackend stubs, integration patterns for ECDSA, ML-KEM, Ed25519, ChaCha20Poly1305, etc.)
- **@qmuntal's PR reviews** scraped via `gh` CLI, including:
  - PR #2342 (ML-DSA): value vs pointer for backend keys, pushing operations into backends, caching reconstructed Go keys, adjusting allocation tests
  - PR #2149 (FIPS 140-only): keeping original variable names, `FIPSApprovedHash` as top-level function
  - PR #2126 (GOEXPERIMENT consolidation): docs requirements, error message guidance
  - PR #2106 (cgo-less promotion): documentation needs
  - PR #1984 (CSHAKE): CrossPlatformCryptography.md update requirement

Covers:
- End-to-end `git go-patch` workflow (apply → edit → commit → rebase/squash → extract → verify)
- Backend shim patterns for all four files (linux/windows/darwin/nobackend)
- Algorithm integration patterns (gate functions, dispatch, fallback)
- Build tag and import conventions
- FIPS 140-only mode enforcement
- Documentation generation via `updatecryptodocs`
- Patch separation rules (vendor in 0001, crypto in 0002)


.github/copilot-instructions.md
```diff
@@ -0,0 +1,323 @@
...
+- Types: empty structs (`struct{}` or `struct{ _ int }`)
+- Methods on stub types: panic
+- `Supports*()`: must panic (not return false) — they should never be called when `Enabled` is `false`
+- Exception: `SupportsChaCha20Poly1305()` returns `false` (for FIPS-only mode checks)
+
+If an operation is not supported on a platform (e.g., Darwin doesn't support DSA), the platform backend should also panic with `"cryptobackend: not available"` and the `Supports*` function should return `false`.
+
+### 2. Integrate into the crypto package
+
+Add a backend key field to the key struct. **Use value types, not pointers**:
```
qmuntal: This is not a good generic advice. Some backend pointers lifetime might be tied to a system-provided object. If we dereference the pointer at this layer, then the system-provided object will be decoupled from the backend object, which is bad.

What I would say here is that the patch should take great care of allocations. If we can avoid one, we should, but that should be a conscious desicion.

qmuntal (changes requested):

gdams:
@qmuntal updated PTAL

qmuntal (approved):

Issue #181, opened 2021-08-18
Rename `.sig` files to `.asc`: match upstream Go


dagood:
Upstream Go provides `.asc` files, serving the same purpose as our `.sig` files:

https://golang.org/dl/go1.17.linux-amd64.tar.gz
https://golang.org/dl/go1.17.linux-amd64.tar.gz.asc

https://dotnetbuildoutput.blob.core.windows.net/golang/microsoft/main/20210818.1/go.20210818.1.linux-amd64.tar.gz.sig

They are both armored PGP signatures.

If I knew Go provided `.asc` files, I would have followed the naming, but they aren't advertised on the download page. I found out about them because the official golang Dockerfiles use the `.asc` files to verify the download.

---

Changing to `.asc` would also let us remove some tweaks I made to the golang Dockerfiles that make them use `.sig` when downloading our builds.

dagood:
Some recent thoughts on this from https://github.com/microsoft/go/pull/2348#discussion_r3357753559, re: this being a breaking change at this point:

I think that publishing both `.sig` and `.asc` makes sense. We can then use `.asc` in our own infra to reduce patching. I don't know if it ever makes sense to stop publishing `.sig`, though.

`.sig` is actually a more intuitive extension to slap on our download URLs to get a signature (depending on exposure to other projects, etc.). It might actually be better to treat `.asc` as a "compatibility" signature, not even the main one.

dagood:
I don't think we can justifiably work on this quite yet, see email thread "Publish internally hit "The Publishing Request SourceFileData List Contains 2 Files That Are Identical"". Currently, we would have to split up the publish into multiple requests. We would have to run them in parallel in order to avoid a significant time hit. This probably involves moving publishing to go-infra to do this reasonably.

The publishing team has an issue on their side tracking a way to do this without a workaround on our side. https://dev.azure.com/devdiv/DevDiv/_workitems/edit/3030311

PR #2348, opened 2026-06-04
Ship .asc signature alongside .sig

gdams:
Ship an `.asc` PGP signature file alongside the existing `.sig` for every archive. The `.sig` remains the default/primary extension; `.asc` is produced in parallel for compatibility with upstream Go tooling (e.g. official Dockerfiles).

For https://github.com/microsoft/go/issues/181

### Changes
- **sign tool (archive.go)**: Produce `.sig` via signing, then copy both `.sig` and `.asc` to the destination directory
- **sign tool (sign.go)**: Test mode handles `.sig` signature files
- **updatelinktable**: No change (`.sig` remains the advertised suffix)

### Companion PR
- https://github.com/microsoft/go-infra/pull/502

### DIFF
--- a/eng/_util/cmd/sign/archive.go
+++ b/eng/_util/cmd/sign/archive.go
@@ -428,7 +428,7 @@ func (a *archive) prepareArchiveSignatures(ctx context.Context) ([]*fileToSign, 
 	if err := ctx.Err(); err != nil {
 		return nil, err
 	}
-	// Copy the archive file with a ".sig" suffix. The signing process sends this file to get a
+	// Copy the archive file and add a ".sig" suffix. The signing process sends this new file to get a
 	// signature, then replaces its content in-place with the result. We need to preemptively make
 	// a renamed copy of the file so we end up with both the original file and signature on the
 	// machine.
--- a/eng/_util/cmd/sign/archive.go
+++ b/eng/_util/cmd/sign/archive.go
@@ -428,10 +428,10 @@ func (a *archive) prepareArchiveSignatures(ctx context.Context) ([]*fileToSign, 
 	if err := ctx.Err(); err != nil {
 		return nil, err
 	}
-	// Copy the archive file to have .sig suffix, e.g. "tar.gz" to "tar.gz.sig". The signing
-	// process sends the "tar.gz.sig" file to get a signature, then replaces the "tar.gz.sig"
-	// file's content in-place with the result. We need to preemptively make a renamed copy of the
-	// file so we end up with both the original file and sig on the machine.
+	// Copy the archive file with a ".sig" suffix. The signing process sends this file to get a
+	// signature, then replaces its content in-place with the result. We need to preemptively make
+	// a renamed copy of the file so we end up with both the original file and signature on the
+	// machine.
 	log.Printf("Copying file for signature generation: %q -> %q", a.latestPath(), a.sigPath())
 	if err := copyFile(a.sigPath(), a.latestPath()); err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


eng/_util/cmd/sign/archive.go
```diff
@@ -94,8 +95,22 @@ func (a *archive) latestPath() string {
...
 
-func (a *archive) sigPath() string {
-	return filepath.Join(a.workDir, a.name+".sig")
+// legacySigNeeded reports whether this archive needs a legacy .sig file
+// in addition to the .asc file, for backward compatibility with go1.26 and earlier.
+func (a *archive) legacySigNeeded() bool {
+	// Archive names look like "go1.27.linux-amd64.tar.gz" or "go1.26.3.src.tar.gz".
+	if after, ok := strings.CutPrefix(a.name, "go1."); ok {
+		if dot := strings.IndexByte(after, '.'); dot > 0 {
+			if minor, err := strconv.Atoi(after[:dot]); err == nil && minor <= 26 {
```
dagood: We're in microsoft/go, so we know what version we're building based on what branch we're in. It isn't necessary to parse the version, and simplifying it leaves the intent clearer.

eng/_util/cmd/sign/archive.go
```diff
@@ -94,8 +95,22 @@ func (a *archive) latestPath() string {
 	return a.path
 }
 
-func (a *archive) sigPath() string {
-	return filepath.Join(a.workDir, a.name+".sig")
+// legacySigNeeded reports whether this archive needs a legacy .sig file
+// in addition to the .asc file, for backward compatibility with go1.26 and earlier.
```
dagood: So the plan here *is* to break it in 1.27? Note that this breaks download tools, not only individual users' workflows. I don't think we should do this break, especially without already establishing simultaneous publishing for a while to allow for time to move.

dagood (changes requested):

eng/_util/cmd/sign/archive.go
dagood: To be clear: I think that publishing both `.sig` and `.asc` makes sense. We can then use `.asc` in our own infra to reduce patching. I don't know if it ever makes sense to stop publishing `.sig`, though.

`.sig` is actually a more intuitive extension to slap on our download URLs to get a signature. It might actually be better to treat `.asc` as a "compatibility" signature, not even the main one.

Users of Go don't even necessarily know about `.asc` from upstream (my old note on https://github.com/microsoft/go/issues/181):

> If I knew Go provided `.asc` files, I would have followed the naming, but they aren't advertised on the download page. I found out about them because the official golang Dockerfiles use the `.asc` files to verify the download.

gdams: Agreed — I've switched this around. `.sig` stays the default/primary extension and we now ship `.asc` alongside it for all versions. No version-gating, no breaking change. The companion go-infra PR is updated to match.

eng/_util/cmd/sign/archive.go
```diff
@@ -428,10 +428,10 @@ func (a *archive) prepareArchiveSignatures(ctx context.Context) ([]*fileToSign,
 	if err := ctx.Err(); err != nil {
 		return nil, err
 	}
-	// Copy the archive file to have .sig suffix, e.g. "tar.gz" to "tar.gz.sig". The signing
-	// process sends the "tar.gz.sig" file to get a signature, then replaces the "tar.gz.sig"
-	// file's content in-place with the result. We need to preemptively make a renamed copy of the
-	// file so we end up with both the original file and sig on the machine.
+	// Copy the archive file with a ".sig" suffix. The signing process sends this file to get a
```
dagood: This change makes it ambiguous (there is no archive file with a `.sig` suffix--yet).


```suggestion
	// Copy the archive file and add a ".sig" suffix. The signing process sends this new file to get a
```

dagood (changes requested):
Other than the comment, looks good.

gdams:
@dagood updated PTAL

dagood (approved):

Issue #2356, opened 2026-06-09
HKDF with SHA3 fails under MS Go builds using a system crypto backend


rainenieminenssh:
### Microsoft build of Go version

go version go1.26.4 linux/amd64

### What is your operating system and platform?

GNU/Linux x86_64
OpenSSL 3.5.5

### Output of `go env` in your module/workspace:

```shell
AR='ar'
CC='gcc'
CGO_CFLAGS='-O2 -g'
CGO_CPPFLAGS=''
CGO_CXXFLAGS='-O2 -g'
CGO_ENABLED='1'
CGO_FFLAGS='-O2 -g'
CGO_LDFLAGS='-O2 -g'
CXX='g++'
GCCGO='gccgo'
GO111MODULE=''
GOAMD64='v1'
GOARCH='amd64'
GOAUTH='netrc'
GOBIN=''
GOCACHEPROG=''
GODEBUG=''
GOEXE=''
GOEXPERIMENT=''
GOFIPS140='off'
GOFLAGS=''
GOGCCFLAGS='-fPIC -m64 -pthread -Wl,--no-gc-sections
-fmessage-length=0
-ffile-prefix-map=/tmp/go-build4082029656=/tmp/go-build
-gno-record-gcc-switches'
GOHOSTARCH='amd64'
GOHOSTOS='linux'
GOINSECURE=''
GONOPROXY=''
GONOSUMDB=''
GOOS='linux'
GOPRIVATE=''
GOPROXY='https://proxy.golang.org,direct'
GOSUMDB='sum.golang.org'
GOTELEMETRY='local'
GOTMPDIR=''
GOTOOLCHAIN='local'
GOVCS=''
GOVERSION='go1.26.4'
GOWORK=''
PKG_CONFIG='pkg-config'
```

### What did you do?

```go
package main

import (
	"crypto/hkdf"
	"crypto/sha256"
	"crypto/sha3"
	"fmt"
)

func main() {
	secret := []byte("secret")

	// HKDF with SHA-256, expected to succeed
	_, err := hkdf.Key(sha256.New, secret, nil, "", 32)
	if err != nil {
		fmt.Println("[SHA-256] FAIL:", err)
	} else {
		fmt.Println("[SHA-256] OK")
	}

	// HKDF with SHA3-256, fails when using system crypto backends
	_, err = hkdf.Key(sha3.New256, secret, nil, "", 32)
	if err != nil {
		fmt.Println("[SHA3-256] FAIL:", err)
	} else {
[...]

PR #2360, opened 2026-06-10
Fix HKDF and HMAC with SHA-3 under system crypto backends

michelle-clayton-work:
## Problem

In 1.26, SHA-3 is routed through the system crypto backend (e.g. OpenSSL on
Linux). When a backend is active and supports SHA-3, `sha3.New256` (and the
other constructors) return a `SHA3` whose backend hash is stored in `boringS`,
while the embedded pure-Go digest `s` is left as a zero value.

`fips140hash_sha3Unwrap` returned `&s.s` — that empty Go digest — whose
`Size()` and `BlockSize()` are both `0`. Unwrapping is used by HKDF and HMAC,
so:

- **HKDF** computed `limit := Size() * 255 == 0` and rejected any positive key
  length with the misleading error `hkdf: requested key length too large`.
- **HMAC** received a `BlockSize()` of `0`, writing the key into a zero-rate
  sponge whose absorb loop never makes progress — an infinite hang.

Both symptoms are specific to system-crypto builds; the pure-Go backend
(`GOEXPERIMENT=nosystemcrypto`) and upstream Go are unaffected.

## Fix

- Return `boringS` from `fips140hash_sha3Unwrap` when the `SHA3` is
  backend-backed, and widen the `sha3Unwrap` linkname signature to `hash.Hash`.
- Reorder `hmac.New` so `boring.NewHMAC` runs *after* the hash is unwrapped,
  so the backend receives its own hash type instead of the wrapped `*SHA3`.

## Testing

- `hkdf.Key(sha3.New256, …)` now succeeds.
- HMAC-SHA3 completes and matches both the pure-Go backend and OpenSSL's
  `openssl mac -digest SHA3-256`.

Related:
- https://github.com/microsoft/go/issues/2356

### DIFF
--- a/patches/0004-Use-crypto-backends.patch
+++ b/patches/0004-Use-crypto-backends.patch
@@ -38,7 +38,7 @@ Subject: [PATCH] Use crypto backends
  src/crypto/fips140/enforcement_test.go        |   4 +
  src/crypto/fips140/fips140.go                 |   3 +-
  src/crypto/hkdf/hkdf.go                       |  14 ++
- src/crypto/hkdf/hkdf_test.go                  |  25 ++-
+ src/crypto/hkdf/hkdf_test.go                  |  32 ++-
  src/crypto/hmac/hmac.go                       |  16 +-
  src/crypto/hmac/hmac_test.go                  |  25 ++-
  src/crypto/hpke/aead.go                       |  14 +-
@@ -101,7 +101,7 @@ Subject: [PATCH] Use crypto backends
  src/hash/notboring_test.go                    |   9 +
  src/net/lookup_test.go                        |   3 +
  src/os/exec/exec_test.go                      |   9 +
- 97 files changed, 1562 insertions(+), 190 deletions(-)
+ 97 files changed, 1569 insertions(+), 190 deletions(-)
  create mode 100644 src/crypto/dsa/boring.go
  create mode 100644 src/crypto/dsa/notboring.go
  create mode 100644 src/crypto/ecdsa/badlinkname.go
@@ -1480,14 +1480,15 @@ index 88439922a5032e..5a3ae17e80efbc 100644
  }
  
 diff --git a/src/crypto/hkdf/hkdf_test.go b/src/crypto/hkdf/hkdf_test.go
-index 57d90f88e93e75..8306dbbf5ad515 100644
+index 57d90f88e93e75..772530c071c45a 100644
 --- a/src/crypto/hkdf/hkdf_test.go
 +++ b/src/crypto/hkdf/hkdf_test.go
-@@ -6,13 +6,15 @@ package hkdf
+@@ -6,13 +6,17 @@ package hkdf
  
  import (
  	"bytes"
 -	"crypto/internal/boring"
++	"crypto"
 +	boring "crypto/internal/backend"
  	"crypto/internal/fips140"
  	"crypto/md5"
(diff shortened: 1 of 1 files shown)


patches/0004-Use-crypto-backends.patch
```diff
@@ -1504,6 +1505,34 @@ index e7976e25193dfe..9578e1e985c02d 100644
...
+-		// BoringCrypto did not recognize h, so fall through to standard Go code.
+-	}
+ 	h = fips140hash.UnwrapNew(h)
+ 	if fips140only.Enforced() {
+ 		if len(key) < 112/8 {
+@@ -53,6 +46,13 @@ func New(h func() hash.Hash, key []byte) hash.Hash {
+ 			panic("crypto/hmac: use of hash functions other than SHA-2 or SHA-3 is not allowed in FIPS 140-only mode")
+ 		}
+ 	}
++	if boring.Enabled {
```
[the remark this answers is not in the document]
michelle-clayton-work: will add

patches/0004-Use-crypto-backends.patch
```diff
@@ -3017,7 +3066,13 @@ index 48c67e9fe11005..3fe46309e0fda8 100644
  }
  
  //go:linkname fips140hash_sha3Unwrap crypto/internal/fips140hash.sha3Unwrap
-@@ -109,26 +130,38 @@ func fips140hash_sha3Unwrap(sha3 *SHA3) *sha3.Digest {
+-func fips140hash_sha3Unwrap(sha3 *SHA3) *sha3.Digest {
++func fips140hash_sha3Unwrap(sha3 *SHA3) hash.Hash {
++	if sha3.boringS != nil {
++		return sha3.boringS
++	}
+ 	return &sha3.s
```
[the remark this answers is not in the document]
michelle-clayton-work: will add

qmuntal (approved):

patches/0004-Use-crypto-backends.patch
```diff
@@ -1517,6 +1590,40 @@ index 4046a9555a8e35..4721a2a7ac04f7 100644
...
+ 	}
+ }
++
++// TestHMACSHA3 is a regression test for
++// https://github.com/microsoft/go/issues/2356. [New] used to spin forever when
++// given a SHA-3 hash under a system crypto backend, because the backend HMAC
++// constructor received the still-wrapped *sha3.SHA3 instead of the unwrapped
++// hash. The hash must now be unwrapped before the backend sees it.
++func TestHMACSHA3(t *testing.T) {
++	if boring.Enabled && runtime.GOOS == "darwin" {
```
qmuntal: `hmac.New` returns nil if the backend doesn't support the given hash. You can skip when that happens instead of hardcoding `darwin` here.

patches/0004-Use-crypto-backends.patch
```diff
@@ -1491,8 +1492,52 @@ index 57d90f88e93e75..4069ab057a2525 100644
...
++		if strings.Contains(err.Error(), "too large") {
++			t.Fatalf("issue #2356 regression: %v", err)
++		}
++		// The macOS CryptoKit backend does not implement HKDF with SHA-3 (it
++		// supports only SHA-1 and the SHA-2 family), so it returns an error here
++		// regardless of macOS version. That is a separate backend limitation,
++		// not issue #2356. OpenSSL, CNG, and the pure Go path all support it, so
++		// only skip on the CryptoKit backend and fail otherwise rather than
++		// hiding an unexpected error.
++		if boring.Enabled && runtime.GOOS == "darwin" {
```
qmuntal: If this is a real issue, then create an issue to track it, thanks!

qmuntal (changes requested):

patches/0004-Use-crypto-backends.patch
michelle-clayton-work: - https://github.com/microsoft/go/issues/2365

qmuntal (approved):

PR #2362, opened 2026-06-11
Implement crypto/internal/backend outside of the standard library

### LINKED ISSUE #2373: Clean up crypto backend patches
The crypto backend patch file is growing big. We could extract some of the logic outside of the submodule and make it more maintainable by using a package per algorithm following the upstream FIPS module.

qmuntal:
The crypto backend layer can be moved outside of the standard library. This way the patch files are smaller and changes in the backend will be easier to review. That is, the cryptobackend package is now regular source code.

Note that this PR adds 2500 lines of code, but all of the additions are in `patches/0001-Vendor-external-dependencies.patch`, which is automatically generated by copying the external dependencies, like now the cryptobackend package. On the other hand, `patches/0002-Add-crypto-backends.patch` loses 2500 lines of code, and that's the real win, given that all these lines were real code.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -8,7 +8,7 @@ See [eng/doc/DeveloperGuide.md](/eng/doc/DeveloperGuide.md) for the full develop
 
 ## Architecture overview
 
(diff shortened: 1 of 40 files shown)


gdams (approved):
Love this!

PR #2443, opened 2026-08-11
go: report removal of ms_nocgo_opensslcrypto

gdams:
The `ms_nocgo_opensslcrypto` experiment was removed in Go 1.27 because CGO-free system crypto is now selected automatically on supported platforms. Existing environments that still set it currently receive a generic `unknown GOEXPERIMENT` error.

Return a specific migration error explaining the new default behavior, consistent with the diagnostics for the retired system crypto experiments. Add focused coverage in `internal/buildcfg`.

Tested with:

```
cd go && ./bin/go test internal/buildcfg
```

### DIFF
--- a/patches/0002-Add-crypto-backends.patch
+++ b/patches/0002-Add-crypto-backends.patch
@@ -127,8 +127,8 @@ Subject: [PATCH] Add crypto backends
  src/hash/example_test.go                      |   2 +
  src/hash/marshal_test.go                      |   4 +
  src/internal/buildcfg/cfg.go                  |   2 +
- src/internal/buildcfg/cfg_test.go             |  44 +++
- src/internal/buildcfg/exp.go                  |  20 ++
+ src/internal/buildcfg/cfg_test.go             |  49 +++
+ src/internal/buildcfg/exp.go                  |  22 ++
  src/internal/cfg/cfg.go                       |   1 +
  src/internal/platform/supported.go            |  12 +
  src/internal/systemcrypto/systemcrypto.go     |  20 ++
@@ -137,7 +137,7 @@ Subject: [PATCH] Add crypto backends
  src/os/exec/exec_test.go                      |   9 +
  src/runtime/runtime_boring.go                 |   5 +
  src/syscall/syscall_windows.go                |   3 +
- 133 files changed, 2524 insertions(+), 378 deletions(-)
+ 133 files changed, 2531 insertions(+), 378 deletions(-)
  create mode 100644 src/cmd/go/systemcrypto_test.go
  create mode 100644 src/crypto/dsa/boring.go
  create mode 100644 src/crypto/dsa/notboring.go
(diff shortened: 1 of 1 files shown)


patches/0002-Add-crypto-backends.patch
```diff
@@ -6417,6 +6417,11 @@ index 2bbd478280241e..77eaacdc38756d 100644
 +	} else if !strings.Contains(err.Error(), "MS_GO_NOSYSTEMCRYPTO=1") || !strings.Contains(err.Error(), "systemcrypto supports CGO_ENABLED=0 since Go 1.27") {
 +		t.Fatalf("ParseGOEXPERIMENT nosystemcrypto error = %q", err)
 +	}
++	if _, err := ParseGOEXPERIMENT("linux", "amd64", "ms_nocgo_opensslcrypto"); err == nil {
++		t.Fatal("ParseGOEXPERIMENT accepted ms_nocgo_opensslcrypto")
++	} else if !strings.Contains(err.Error(), "has been removed") || !strings.Contains(err.Error(), "CGO_ENABLED=0 is enabled automatically") || !strings.Contains(err.Error(), "Go 1.27") {
```
dagood:
```suggestion
+	} else if !strings.Contains(err.Error(), "has been removed") || !strings.Contains(err.Error(), "supports CGO_ENABLED=0 automatically on platforms with a cgo-less OpenSSL implementation since Go 1.27") {
```

patches/0002-Add-crypto-backends.patch
```diff
@@ -6448,12 +6453,14 @@ index bf411a8dbbfcde..9b6ca9feb0b23b 100644
...
  
-@@ -130,6 +132,21 @@ func ParseGOEXPERIMENT(goos, goarch, goexp string) (*ExperimentFlags, error) {
+@@ -130,6 +132,23 @@ func ParseGOEXPERIMENT(goos, goarch, goexp string) (*ExperimentFlags, error) {
  			if strings.HasPrefix(f, "no") {
  				f, val = f[2:], false
  			}
 +			// Check for removed crypto experiments.
 +			switch f {
++			case "ms_nocgo_opensslcrypto":
++				return nil, fmt.Errorf("GOEXPERIMENT=ms_nocgo_opensslcrypto has been removed; system crypto with CGO_ENABLED=0 is enabled automatically on supported platforms since Go 1.27")
```
dagood:
```suggestion
+				return nil, fmt.Errorf("GOEXPERIMENT=ms_nocgo_opensslcrypto has been removed; system crypto supports CGO_ENABLED=0 automatically on platforms with a cgo-less OpenSSL implementation since Go 1.27")
```

dagood (commented):
"system crypto with CGO_ENABLED=0 is enabled automatically" isn't a good description of the feature IMO. I think we should borrow the language from the release notes.

dagood (approved):

PR #2444, opened 2026-08-11
[release-branch.go1.27] go: report removal of ms_nocgo_opensslcrypto

gdams:
Backport of #2443 to Go 1.27.

Report an actionable error when `GOEXPERIMENT=ms_nocgo_opensslcrypto` is set, explaining that CGO-free system crypto is enabled automatically on supported platforms since Go 1.27. This avoids the generic `unknown GOEXPERIMENT` error during upgrades.

Testing: Not run, per request. Patch extraction and equivalence checks completed successfully; `git diff --check` passed.

### DIFF
--- a/patches/0002-Add-crypto-backends.patch
+++ b/patches/0002-Add-crypto-backends.patch
@@ -6418,7 +6418,7 @@ index 2bbd478280241e..dd58604b80b77e 100644
 +	}
 +	if _, err := ParseGOEXPERIMENT("linux", "amd64", "ms_nocgo_opensslcrypto"); err == nil {
 +		t.Fatal("ParseGOEXPERIMENT accepted ms_nocgo_opensslcrypto")
-+	} else if !strings.Contains(err.Error(), "has been removed") || !strings.Contains(err.Error(), "CGO_ENABLED=0 is enabled automatically") || !strings.Contains(err.Error(), "Go 1.27") {
++	} else if !strings.Contains(err.Error(), "has been removed") || !strings.Contains(err.Error(), "supports CGO_ENABLED=0 automatically on platforms with a cgo-less OpenSSL implementation since Go 1.27") {
 +		t.Fatalf("ParseGOEXPERIMENT ms_nocgo_opensslcrypto error = %q", err)
 +	}
 +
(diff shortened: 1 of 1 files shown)


patches/0002-Add-crypto-backends.patch
```diff
@@ -6447,12 +6452,14 @@ index bf411a8dbbfcde..9b6ca9feb0b23b 100644
...
  
-@@ -130,6 +132,21 @@ func ParseGOEXPERIMENT(goos, goarch, goexp string) (*ExperimentFlags, error) {
+@@ -130,6 +132,23 @@ func ParseGOEXPERIMENT(goos, goarch, goexp string) (*ExperimentFlags, error) {
  			if strings.HasPrefix(f, "no") {
  				f, val = f[2:], false
  			}
 +			// Check for removed crypto experiments.
 +			switch f {
++			case "ms_nocgo_opensslcrypto":
++				return nil, fmt.Errorf("GOEXPERIMENT=ms_nocgo_opensslcrypto has been removed; system crypto with CGO_ENABLED=0 is enabled automatically on supported platforms since Go 1.27")
```
dagood:
```suggestion
+				return nil, fmt.Errorf("GOEXPERIMENT=ms_nocgo_opensslcrypto has been removed; system crypto supports CGO_ENABLED=0 automatically on platforms with a cgo-less OpenSSL implementation since Go 1.27")
```

patches/0002-Add-crypto-backends.patch
```diff
@@ -6416,6 +6416,11 @@ index 2bbd478280241e..77eaacdc38756d 100644
 +	} else if !strings.Contains(err.Error(), "MS_GO_NOSYSTEMCRYPTO=1") || !strings.Contains(err.Error(), "systemcrypto supports CGO_ENABLED=0 since Go 1.27") {
 +		t.Fatalf("ParseGOEXPERIMENT nosystemcrypto error = %q", err)
 +	}
++	if _, err := ParseGOEXPERIMENT("linux", "amd64", "ms_nocgo_opensslcrypto"); err == nil {
++		t.Fatal("ParseGOEXPERIMENT accepted ms_nocgo_opensslcrypto")
++	} else if !strings.Contains(err.Error(), "has been removed") || !strings.Contains(err.Error(), "CGO_ENABLED=0 is enabled automatically") || !strings.Contains(err.Error(), "Go 1.27") {
```
dagood:
```suggestion
+	} else if !strings.Contains(err.Error(), "has been removed") || !strings.Contains(err.Error(), "supports CGO_ENABLED=0 automatically on platforms with a cgo-less OpenSSL implementation since Go 1.27") {
```

dagood (commented):
"system crypto with CGO_ENABLED=0 is enabled automatically" isn't a good description of the feature IMO. I think we should borrow the language from the release notes.

dagood (changes requested):
This PR will need to be redone after https://github.com/microsoft/go/pull/2443 gets the change.

gdams:
@dagood updated PTAL

dagood (approved):

PR #2470, opened 2026-08-21
Parameterize getmingw call rather than duplicate long line

dagood:
* Patch recent change https://github.com/microsoft/go/pull/2465
  * Avoid duplicating for maintainability and readability.
  * Chop long line for readability.

### DIFF
--- a/eng/pipeline/stages/run-stage.yml
+++ b/eng/pipeline/stages/run-stage.yml
@@ -230,10 +230,16 @@ stages:
               - pwsh: |
                   eng/run.ps1 go run github.com/microsoft/go-infra/cmd/getmingw diagnose
                   if ('${{ parameters.builder.arch }}' -eq '386') {
-                    eng/run.ps1 go run github.com/microsoft/go-infra/cmd/getmingw run -ci azdo -source nixman -version 13.2.0-rt_v11-rev0 -arch i686 -threading posix -exception dwarf -runtime msvcrt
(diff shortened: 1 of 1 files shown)


eng/pipeline/stages/run-stage.yml
```diff
@@ -230,10 +230,16 @@ stages:
...
+                    $arch = 'i686'
+                    $exception = 'dwarf'
                   } else {
-                    eng/run.ps1 go run github.com/microsoft/go-infra/cmd/getmingw run -ci azdo -source nixman -version 13.2.0-rt_v11-rev0 -arch x86_64 -threading posix -exception seh -runtime msvcrt
+                    $arch = 'x86_64'
+                    $exception = 'seh'
                   }
+                  eng/run.ps1 go run github.com/microsoft/go-infra/cmd/getmingw run `
+                    -ci azdo `
+                    -source nixman -version 13.2.0-rt_v11-rev0 `
```
karianna, line 241: Do we externalize version #?

gdams (approved):

eng/pipeline/stages/run-stage.yml
dagood: As in make this update via dependabot/etc.? We don't currently, and `getmingw` has a catlogue of all expected versions (and checksums), so the path isn't super clear. Opened https://github.com/microsoft/go-lab/issues/496 to track.

